PR #1, opened 2018-06-14
Implement UART for sercom

zklapow:
This adds structs and corresponding `serial::{Read, Write}` impls for a UART on each sercom.

This works as is but needs some improvement to fully support all the features available on the samd21. Currently this is:

- Only asynchronous
- Only Arithmetic baud mode is supported
- Only 8N1
- Both TX/RX are always on

@wez

### DIFF
--- a/hal/Cargo.toml
+++ b/hal/Cargo.toml
@@ -30,6 +30,13 @@ path = "../atsamd21e18a"
 version = "~0.2"
 optional = true
 
+[dev-dependencies]
+panic-abort = "~0.1"
+panic-semihosting = "~0.1"
+cortex-m-semihosting = "~0.2"
+cortex-m-rtfm = "~0.3"
+sx1509 = "~0.2"
+
 [features]
 # We default to picking one of the target mcus so that a `cargo build` or
 # `cargo package` run for this crate will succeed.  Without this we won't
--- a/hal/src/sercom/mod.rs
+++ b/hal/src/sercom/mod.rs
@@ -11,7 +11,9 @@
 mod i2c;
 mod pads;
 mod spi;
+mod uart;
 
 pub use self::i2c::*;
 pub use self::pads::*;
 pub use self::spi::*;
+pub use self::uart::*;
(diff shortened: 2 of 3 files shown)


wez:
Many thanks for this!  I'll have some time to properly review on saturday.
Hmm, looks like the travis CI bits might need updating for more recent changes in rust.

zklapow:
👌 thanks @wez I'll probably update this a bit this weekend as well so no rush, at the very least I want to add a UART example but I need to clean up some of my code.

zklapow:
I also pinned the nightly version to `2018-05-16` (which was the version I was developing on) to fix some the upstream build issues. I was able to fix the `panic_implementation` errors by upgrading `panic-abort` but then got a bunch of proc-macro issues which seem to need upstream changes to get merged in `cortex-m-rtfm` so I just went with rolling back nightly.

zklapow:
@wez any chance you've had some time to take a look at this?

wez:
@zklapow sorry for not getting back to you already; I'll look at this tomorrow!

djmcgill:
I've been using this branch, seems to be okay on a samd21g18a. Is there a reason that `core::fmt::Write` isn't implemented for the blocking variant?

The version which I'm using is (the struct is only there to get around the orphan issue):
```
struct BlockingWrite<W>(W) where W: hal::hal::blocking::serial::Write<u8>;
impl<W: hal::hal::blocking::serial::Write<u8>> core::fmt::Write for BlockingWrite<W> {
    fn write_str(&mut self, s: &str) -> core::fmt::Result {
        match self.0.bwrite_all(s.as_bytes()) {
            Ok(()) => Ok(()),
            Err(_) => Err(core::fmt::Error),
        }
    }
}
```

zklapow:
@djmcgill glad its working for you! I actually had a commit with implemented the `core::fmt` side of things that I just forgot to push, should be there now. I think I managed to avoid the intermediate struct by moving some imports around.

PR #5, opened 2018-08-26
Add support for arduino mkr_zero pins

djmcgill:
Adds support for the arduino mkrzero's pins.

Additionally:
- Changes the .cargo/config file to support the recent nightly change to linking.
- Defines and uses the macro input_pins! which could potentially be used in the other sections.
- When the "rt" feature is enabled (which it is by default) it also exports the "entry" macro (possibly the others would be useful too?)
- When the "panic_abort" feature is enabled (also by default) it exports the panic language item for you.

I would be grateful if you could double-check that the pin information in this commit matches the linked [variant.cpp](https://github.com/arduino/ArduinoCore-samd/blob/master/variants/mkrzero/variant.cpp)


djmcgill:
Note that the build is failing because the metro_m0's rtfm dependency is still on `cortex-m-0.4.x`, and if you have `cortex-m-0.5.x` in the build _at all_ it seems like it the no_mangle symbols like `CORE_PERIPHERALS` get redefined and conflict.

I have pushed a change to jenkins removing the rtfm examples from the build just to get my build passing, but I can see why you wouldn't want that in master.

wez:
Oh, I didn't see this until now for some reason (github notifications failed me).  I'll review this this evening!

arduino_mkrzero/.cargo/config
```diff
@@ -0,0 +1,10 @@
+# samd21 is a Cortex-M0 and thus thumbv6m
```
wez: can we survive with the top level `.cargo/config`, or alternatively, should each of the board crates have one of these?

arduino_mkrzero/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+    #[doc = "Analog 4: PWM, TCC"]
+    pin a4 = pa5: Pa5,
+
+    #[doc = "Analog 5"]
+    pin a5 = pa6: Pa6,
+
+    #[doc = "Analog 6"]
+    pin a6 = pa7: Pa7,
+
+    pin usb_n = pa24: Pa24,
```
wez: naming wise, these pins correlate with `DmPad` (D-), `DpPad` (D+) and the `SofPad`.   I picked those names in `hal/src/usb/mod.rs` so it certainly isn't an established convention there.  It's not clear from the variant.cpp you linked to whether these pins are well known via some other symbolic name, so I wonder if the names here should match those from `usb/mod.rs`?

I noticed that the `PaXX` numbers in `usb/mod.rs` don't match those here or in variant.cpp, but the numbers in `usb/mod.rs` do match the arduino pin numbers, so I suspect that `usb/mod.rs` may have a transcription error.

arduino_mkrzero/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+use hal::prelude::*;
+pub use hal::*;
+
+use gpio::{Floating, Input, Port};
+
+// This could possibly be simplified with concat_ident p for the name and P for the type
+// but I couldn't get it to work. Meta attribues (i.e. doc comments) are optional.
+macro_rules! input_pins {
+    // Insert trailing comma if it's missing
+    ($( $(#[$attr:meta])* pin $name:ident = $pin_name:ident: $pin_type:ident),+ ) =>
```
wez: nice macro!   FWIW, I would have (and have actually done that in other macros in this repo) just require the trailing `,`.  I think the audience for this macro is small enough that this isn't a terrible burden.

I'd be in favor of moving this into the `hal` crate and exporting it to the board support crates.

arduino_mkrzero/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+    ($( $(#[$attr:meta])* pin $name:ident = $pin_name:ident: $pin_type:ident),+ , ) => {
+        /// Maps the pins to their arduino names and the numbers printed on the board.
+        /// Information from: https://github.com/arduino/ArduinoCore-samd/blob/master/variants/mkrzero/variant.cpp
+        pub struct Pins {
+            /// Opaque port reference
+            pub port: Port,
+            $($(#[$attr])* pub $name: gpio::$pin_type<Input<Floating>>),+
+        }
+
+        pub fn pins(port: atsamd21g18a::PORT) -> Pins {
```
wez: if/when this macro moves to the `hal` crate, this should probably become something like `port: hal::target_device::PORT`.

arduino_mkrzero/src/lib.rs
wez: I also wonder whether we can use something similar to this in `hal/src/gpio.rs` itself; I'm thinking of the lame/missing doc comments there.   I do wish that rust had a better story around constructing identifiers and strings in macros.

arduino_mkrzero/.cargo/config
wez: I pushed 39874f8b5b9f1e9f6862f1910034a62e112f334f with the net effect of this at the top level, so I think you can remove this file from your PR.  (Also, I only just noticed that you changed the top level file in this PR too)

Thanks!

arduino_mkrzero/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+
+extern crate atsamd21_hal as hal;
+
+#[cfg(feature = "rt")]
+extern crate cortex_m_rt;
+#[cfg(feature = "rt")]
+pub use cortex_m_rt::entry;
+
+#[cfg(feature = "panic_abort")]
+pub extern crate panic_abort;
```
wez: I have mixed feelings about this, but it does feel more ergnomic overall.  We should probably do the same for all of the board support crates, or potentially even in the `hal` crate in the spirit of DRY.

arduino_mkrzero/src/lib.rs
```diff
@@ -0,0 +1,130 @@
+#![no_std]
+
+extern crate atsamd21_hal as hal;
+
+#[cfg(feature = "rt")]
+extern crate cortex_m_rt;
+#[cfg(feature = "rt")]
+pub use cortex_m_rt::entry;
```
wez: +1 on this.  I was a little disappointed by the amount of boiler plate that the executables needed with the changes around `entry`.

arduino_mkrzero/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+                port: pins.port,
+                $($name: pins.$pin_name),+
+            }
+        }
+    }
+}
+
+// The docs could be further improved with details of the specific channels etc
+input_pins!(
+    #[doc = "Digital 0: PWM, TC"]
```
wez: I've confirmed that these pin assignments match up to the variant.cpp you linked.

arduino_mkrzero/src/lib.rs
wez: I checked my samd21 datasheet and the assignments in `usb/mod.rs` match up to it.  
https://www.arduino.cc/en/uploads/Main/ArduinoMKRZero-schematic.pdf shows that the `usb_n` and `usb_p` pins are attached to `Pa24` and `Pa25` and correlate with `usb/mod.rs`.   Ah, and `USB_ID` is connected to USB OTG ID pin in the micro connector.

djmcgill:
Okay @wez deleted the local file.
I only had it so I didn't have to have to type manifest-path each time or Cs into the directory and lose the effect of it.

djmcgill:
Build failed again with https://github.com/rust-lang/rust/issues/53849

wez:
Thanks!

PR #10, opened 2018-09-05
fix macro type rename bug

djmcgill:
I tested this by temporarily renaming the Pins struct in one of the device crates - all worked fine


wez:
Good catch!

PR #12, opened 2018-09-16
Circuit Playground Express support

sajattack:
Hey, just throwing this up now because I don't want someone else duplicating work. It's not ready yet.


wez:
Thanks!  A couple of things:
1. I moved the boards into the boards subdir this weekend, and git probably isn't smart enough to know to move new files when rebasing, so you'll want to rebase and move your crate into the `boards` subdir
2. Please add a line to `build-all.sh` to build your blinky example; that way the CI will help keep this board building and healthy
3. Please add your crate to this list: https://github.com/wez/atsamd21-rs/blob/master/build-docs.py#L14 so that we can update the docs when this gets merged
4. You'll want to revise the doc URL in the top level readme; you'll see how the others look when you rebase

BenBergman:
I'm wondering if the analog pin names should be used instead of the digital since the board labels only show the analog and serial names.

BenBergman:
Also, the blink example works on my board but the first Neopixel gets turned on. I'm guessing that is just junk data getting to the first LED in the chain, and I'm sure nothing should change in the example, but I'm wondering if something can be added to the core crate to prevent this from happening. (Though maybe that is a hard problem to solve.)

sajattack:
I'm eventually planning on doing a neopixel driver.

sajattack:
That should do it, some of the digital pins are actually labelled, the ones attached to the buttons and the red LED for example.

sajattack:
As an aside, I think it would be great if pins could have multiple names, but I guess that's up to the embedded-hal folks to implement.

wez:
Squashed and pushed; thanks!

BenBergman:
> On the Metro M0 board, the pin used for the neopixel does double duty as SWD IO or clock signal when a SWD debugger is attached, resulting in some bright and irritating flashes during debugging. Is that also the case for this board? If so, there's not a lot you can do about it except for not using SWD :-/

The NeoPixel pin is not shared with anything else from what I can tell (and I'm loading the board with the bossac USB bootloader anyway). The bootloader does drive the LEDs though, so maybe it is just being interrupted by the new code and that last bit of data is lingering on the LED. Using Arduino with the board does not have this issue, but they might be initializing a NeoPixel library and blanking it out on start or something.

BenBergman:
> Also, thank you for testing.

Thank you for writing! 😄 

> I'm eventually planning on doing a neopixel driver. I'll have a look at what the pins are labelled on the silkscreen.

I see that the NeoPixels are attached to pin 3 of SERCOM 5. I imagine you could make a generic SPI-based NeoPixel driver like has been done for Arduino using this board.

sajattack:
> Also, the blink example works on my board but the first Neopixel gets turned on. I'm guessing that is just junk data getting to the first LED in the chain, and I'm sure nothing should change in the example, but I'm wondering if something can be added to the core crate to prevent this from happening. (Though maybe that is a hard problem to solve.)

I just got my board and I can't reproduce this issue.

BenBergman:
Huh. How are you flashing your board? I'm using the bossac binary supplied
with the Arduino package.

BenBergman:
I think I experience the same thing with the LEDs staying on after reset, which makes sense since the bootloader turns on the LEDs and reset immediately gets the chip to run our Rust code which does not drive the LEDs. I think the only fix for that is to by default turn off all LEDs at the start of any Rust program.

I wonder if my issue is related to having an older board (I've had mine for about a year). Perhaps there was a bug in the bootloader that has been fixed in your copy.

BenBergman:
Yeah, my money is on this: https://github.com/adafruit/uf2-samdx1/commit/72e311b7908ec32ffd62e23cb2a23935b1f2849e

I'll have to flash a new bootloader when I get my Black Magic Probe setup in a few weeks.

BenBergman:
Turns out they have a UF2 bootloader upgrade path. Updating my bootloader fixed my issue so I can now look at my board without going blind. :smile:

sajattack:
UF2 FTW

PR #14, opened 2018-10-10
Implement improved Delay, covering full u32 range

jamesmunns:
Based on @gbip's impl in https://github.com/therealprof/stm32f4xx-hal

### DIFF
--- a/hal/src/delay.rs
+++ b/hal/src/delay.rs
@@ -50,17 +50,29 @@ impl DelayMs<u8> for Delay {
 
 impl DelayUs<u32> for Delay {
     fn delay_us(&mut self, us: u32) {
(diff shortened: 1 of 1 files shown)


TeXitoi:
All these copy and pasted delay implementations seems wrong to me. I think this should be addressed in embedded_hal, providing a way to wrap any timer in such a delay function.

jamesmunns:
I generally agree with you, the issue up to now was the different ways that the main sysclock is exposed in the different HALs.

That being said, I think https://github.com/rust-embedded/embedded-hal/issues/103 is a move in the right direction.

wez:
Thanks; tested this with the blinky example for metro m0 using a 1s interval and it no longer faults like the old impl used to :-p

PR #17, opened 2018-10-16
Add basic support for Trinket M0 board

BenBergman:
I need to test this yet (my spare Trinket M0 is currently burried in a project that I don't currently have access too) but thought I'd at least put up the branch for comment until I can test it out it a few days.


BenBergman:
I tested the blink example and confirmed it works. I tried modifying it to read buttons from other pins to set the led value but it seems that the `has::gpio:Pa*` types only implement OutputPin, but not InputPin.

BenBergman:
Scratch that. I see now that InputPin is behind the `unproven` feature gate. Enabling that got my input test working (not included in this PR).

I think this is ready for final review.

wez:
Looks great, thanks!

PR #30, opened 2018-12-12
Initial support for the SODAQ ONE board

freylax:


boards/sodaq_one/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+path = "../../hal"
+version = "~0.2"
+default-features = false
+
+[dev-dependencies]
+panic-abort = "~0.3"
+panic-semihosting = "~0.5"
+cortex-m-semihosting = "~0.3"
+cortex-m-rtfm = "~0.3"
+sx1509 = "~0.2"
```
wez: you can remove this dep because none of your examples use this; please double check to make sure that you're only listing what you need!

boards/sodaq_one/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// The I2C clock line
+    pin i2c_scl = a23,
+    
+    /// LoRa RESET
+    pin lora_reset = a4,
+    /// MAG_INT
+    pin mag_int = a19,
+);
+
+/// Convenience for setting up the labelled SPI peripheral.
```
wez: Could you tweak this doc comment?  From the sodaq web page it looks like the SPI pins aren't labelled as SPI but just D8 through D11.  I think the comment should say something like:

> Convenience for setting up pins D8-D11 for SPI.

wez:
I fixed up the deps and pushed this by hand; thanks!

PR #40, opened 2019-03-03
Add basic support for Adafruit's NeoTrellis M4 Express board

tarcieri:
Adds ATSAMD51G19A PAC and HAL support and a basic board definition, along with a `blinky_rainbow` example which lights up the whole matrix of LED buttons.

Proof-of-blinky:

ezgif-5-66826ee0bddf


sajattack:
I should've mentioned, travis won't actually test your board until you add it to build-all.sh and build-docs.py

sajattack:
Looks good! Thank you for all your hard work. It should be noted that the neopixel strangeness is a result of a smart-leds-rs problem, not a problem with this PR. The fact that you get anything on the neopixels is sufficient to prove this board support crate works.

PR #42, opened 2019-03-10
fix timers

sajattack:
cc @david-sawatzke
`metro_m4/examples/timer.rs` using a frequency of 3MHz (toggling on separate ticks so we get half speed measured)
Before:
before
After: 
after

### DIFF
--- a/hal/src/timer.rs
+++ b/hal/src/timer.rs
@@ -74,7 +74,7 @@ where
 
         let count = self.tc.count_16();
 
-         // Disable the timer while we reconfigure it
+        // Disable the timer while we reconfigure it
         count.ctrla.modify(|_, w| w.enable().clear_bit());
         while count.status.read().syncbusy().bit_is_set() {}
 
--- a/hal/src/timer.rs
+++ b/hal/src/timer.rs
@@ -46,31 +46,9 @@ where
     where
         T: Into<Hertz>,
     {
-        let timeout = timeout.into();
-        let ticks: u32 = self.freq.0/timeout.0.max(1);
-        let divider = ((ticks >> 16) + 1).next_power_of_two();
-        let divider = match divider {
-            1 | 2 | 4 | 8 | 16 | 64 | 256 | 1024 => divider,
-            // There are a couple of gaps, so we round up to the next largest
-            // divider; we'll need to count twice as many but it will work.
-            32 => 64,
-            128 => 256,
-            512 => 1024,
-            // Catch all case; this is lame.  Would be great to detect this
-            // and fail at compile time.
-            _ => 1024,
-        };
-
-        let cycles: u32 = ticks / divider as u32;
-
-        // How many cycles of the clock need to happen to reach our
-        // effective value.
-        if cycles > u16::max_value() as u32 {
-            panic!(
-                "cycles {} is out of range for a 16 bit counter (timeout={})",
-                cycles, timeout.0
-            );
-        }
+        let params = TimerParams::new(timeout, self.freq.0);
+        let divider = params.divider;
+        let cycles = params.cycles;
 
         let count = self.tc.count_16();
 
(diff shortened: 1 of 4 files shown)


sajattack:
Oh maybe I shouldn't have copied samd51 into samd21 lol.

hal/src/timer.rs
```diff
@@ -47,11 +47,26 @@ where
...
+            // and fail at compile time.
+            _ => 1024,
+        };
+
+        let cycles: u32 = ticks / divider as u32;
+
         let count = self.tc.count_16();
 
-        // Disable the timer while we reconfigure it
+         // Disable the timer while we reconfigure it
```
david-sawatzke: **WHITESPACE** Changes

sajattack:
You can't run < 2Hz on SAMD51 locked to DPLL0 (120MHz), but you probably couldn't before either. You can still run < 2Hz if you run off a reasonable clock.

david-sawatzke:
Currently timers are **very** broken and only accidentally work.

The first thing the current timer start code does is call [ClockParams::new](https://github.com/atsamd-rs/atsamd/blob/2abe66eab73451660c8c538abe775c3164b82047/hal/src/clock.rs#L333), which tries to find a prescaler, so that the resulting frequency is equal or lower given the target one. That means the timer frequency from the prescaler is usually lower than the target frequency. This is bad. 

It accidentally worked for lower frequencies, since the largest prescaler is 1024 and that works out fine for frequencies (120MHz/1024 = 117kHz) lower than 20kHz. Everything above 117 kHz is almost always way off.

This PR changes the behaviour, so the smallest prescaler that's still sufficient is selected. 
It first determines the clock ticks that make up the duration.
It gets the bits that are too large for the 16 bit counter register
```
ticks >> 16
```
adds 1 to round up instead of down, like with 0x1FFFF ticks. This is the smallest theoretical whole number prescaler.

It then determines the next largest existing prescaler and gets the resulting cycles by dividing the total tick amount by the prescaler. This *should* work, but I can't test it as i don't have the hardware, but @sajattack tested it for a few frequencies

david-sawatzke:
The minimum possible frequency is 120MHz/1024/2^16 = 1.78 kHz, so unfortunately we're missing 1 Hz, but that seems to be a hardware limitation. (I haven't looked at further options)

hal/src/timer.rs
```diff
@@ -47,11 +47,26 @@ where
         T: Into<Hertz>,
     {
         let timeout = timeout.into();
-        let params = clock::ClockParams::new(self.freq, timeout);
-        let divider = params.divider;
+        let ticks: u32 = self.freq.0/timeout.0.saturating_sub(1).max(1);
+        let overflow: u16 = ((ticks >> 16) + 1) as u16;
```
wez: Can we fix this in `clock::ClockParams::new` instead of duplicating most of its code out into two other places?

david-sawatzke: ClockParams::new is supposed to find the nearest matching frequency, so it's not broken. The usage of it here is and I can't think of a nice way of nudging it into what we want it to do. (Find the highest possible frequency, that's less than 2^16 * desired frequency)

sajattack: I wasn't sure of all the places it was used and didn't want to cause any side effects elsewhere, but if you think it's fine I'll go ahead with it.

hal/src/timer.rs
```diff
@@ -47,11 +47,26 @@ where
         T: Into<Hertz>,
     {
         let timeout = timeout.into();
-        let params = clock::ClockParams::new(self.freq, timeout);
-        let divider = params.divider;
+        let ticks: u32 = self.freq.0/timeout.0.saturating_sub(1).max(1);
```
david-sawatzke: @sajattack I don't think we need the `saturating_sub` here anymore. It breaks 2Hz

hal/src/timer.rs
david-sawatzke: @sajattack This also doesn't **need** to be u16, so just remove it & it's probably nicer to just combine it with the next line

hal/src/timer.rs
```diff
@@ -73,7 +88,6 @@ where
         // How many cycles of the clock need to happen to reach our
```
david-sawatzke: Either move the overflow check & the comment up to the division or the division down here

david-sawatzke: (same for the other file)

hal/src/timer.rs
david-sawatzke: I think this stuff is **really** timer specific & it's duplicated either way, since we also have 2 clock.rs files. Since the `ClockParams::new` isn't used anywhere else, we can probably just delete it.

May be nice to do a separate timer method for determining the timing, since the `new` method is getting pretty hairy

sajattack:
@wez Does this look good now?

wez:
Looks good to me; thanks!

PR #43, opened 2019-03-11
trellis_m4 keypad-unproven feature

tarcieri:
Adds keypad support using the `keypad` crate.

Requires use of the `unproven` feature, so it's disabled-by-default and gated under the `keypad-unproven` feature.

Also adds a `pins` module containing sets of pins grouped by functions.

This is necessary to split up ownership of the pins so we can use the keypad-related ones with the `keypad` crate.

ezgif-5-ec13d1fc91c7

### DIFF
--- a/boards/trellis_m4/Cargo.toml
+++ b/boards/trellis_m4/Cargo.toml
@@ -16,6 +16,7 @@ documentation = "https://atsamd-rs.github.io/atsamd/atsamd51g19a/trellis_m4/"
 [dependencies]
 cortex-m = "~0.5"
 embedded-hal = "~0.2"
+keypad = { version = "~0.1", optional = true }
 nb = "~0.1"
 
 [dependencies.cortex-m-rt]
@@ -41,6 +42,7 @@ ws2812-nop-samd51 = { git = "https://github.com/smart-leds-rs/ws2812-nop-samd51.
 default = ["rt", "atsamd-hal/samd51g19a", "atsamd-hal/samd51"]
 rt = ["cortex-m-rt", "atsamd-hal/samd51g19a-rt"]
 unproven = ["atsamd-hal/unproven"]
+keypad-unproven = ["keypad", "unproven"]
 
 [profile.dev]
 incremental = false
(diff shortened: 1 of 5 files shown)


tarcieri:
Since the `keypad-unproven` feature is off-by-default, it should probably be explicitly enabled as part of the build.

tarcieri:
This could also probably use an example that confirms the keypad is working

boards/trellis_m4/src/lib.rs
```diff
@@ -78,6 +87,132 @@ define_pins!(
     pin dotstar_di = b3,
 );
 
+impl Pins {
```
sajattack: Maybe move this section into the pins module.

boards/trellis_m4/Cargo.toml
```diff
@@ -41,6 +42,7 @@ ws2812-nop-samd51 = { git = "https://github.com/smart-leds-rs/ws2812-nop-samd51.
 default = ["rt", "atsamd-hal/samd51g19a", "atsamd-hal/samd51"]
 rt = ["cortex-m-rt", "atsamd-hal/samd51g19a-rt"]
 unproven = ["atsamd-hal/unproven"]
+keypad-unproven = ["keypad", "unproven"]
```
sajattack: I would just call this keypad

tarcieri: The problem is crate dependencies and cargo features exist in the same namespace, so I can't have `keypad` activate the `unproven` feature. That said, the existing gating can be changed from:

```
#[cfg(feature = "keypad-unproven"))]
```

to

```
#[cfg(all(feature = "keypad", feature = "unproven"))]
```

boards/trellis_m4/src/lib.rs
tarcieri: I tried to move the `Pins` struct into the `pins` module, but for whatever reason the `define_pins!` macro isn't accessible there (and you can't `impl` a type that lives in a different module).

tarcieri: Specifically:

```
error: cannot find macro `define_pins!` in this scope
 --> src/pins.rs:6:1
  |
6 | define_pins!(
  | ^^^^^^^^^^^
```

PR #47, opened 2019-03-21
trellis_m4: ADXL343 accelerometer support (and examples)

tarcieri:
Uses my recently released `adxl343` crate:

https://github.com/NeoBirth/ADXL343.rs

Includes the following example to visualize the accelerometer output using the NeoPixels:

ezgif-1-9dd510d18d2d

### DIFF
--- a/boards/trellis_m4/src/lib.rs
+++ b/boards/trellis_m4/src/lib.rs
@@ -3,6 +3,9 @@
 
 extern crate atsamd_hal as hal;
 
+#[cfg(feature = "adxl343")]
+extern crate adxl343;
+
 #[cfg(feature = "rt")]
 extern crate cortex_m_rt;
 
@@ -17,14 +20,18 @@ pub use cortex_m_rt::entry;
 pub use hal::{*, atsamd51g19a::*};
 
 use gpio::{Floating, Input, Port};
+use hal::clock::GenericClockController;
+use hal::sercom::I2CMaster4;
+use hal::time::Hertz;
 
 use hal::prelude::*;
-use hal::clock::GenericClockController;
-use hal::sercom::{I2CMaster4, PadPin};
-use hal::time::Hertz;
 #[cfg(feature = "keypad-unproven")]
 use hal::gpio::{OpenDrain, Output, PullUp};
 
+/// Number of Neopixels on the device
+pub const NEOPIXEL_COUNT: usize = 32;
+
+// TODO(tarcieri): move this to the `pins` module (the macro doesn't work?
 define_pins!(
     /// Maps the pins to their arduino names and
     /// the numbers printed on the board.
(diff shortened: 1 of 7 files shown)


boards/trellis_m4/src/lib.rs
```diff
@@ -12,19 +15,24 @@ pub extern crate keypad;
 
 pub mod pins;
 
+/// Adxl343 (preconfigured to use SERCOM2)
+#[cfg(feature = "adxl343")]
+type Adxl343 = adxl343::Adxl343<hal::sercom::I2CMaster2>;
```
tarcieri: It might be a bit draconian to require people use SERCOM2...

boards/trellis_m4/src/lib.rs
```diff
@@ -12,19 +15,24 @@ pub extern crate keypad;
...
 
 use hal::prelude::*;
-use hal::clock::GenericClockController;
-use hal::sercom::{I2CMaster4, PadPin};
-use hal::time::Hertz;
 #[cfg(feature = "keypad-unproven")]
 use hal::gpio::{OpenDrain, Output, PullUp};
 
+/// Number of Neopixels on the device
+pub const NEOPIXEL_COUNT: usize = 32;
```
tarcieri: Figured this could use its own constant

boards/trellis_m4/src/lib.rs
```diff
@@ -212,25 +220,3 @@ impl Keypad {
...
-    let gclk0 = clocks.gclk0();
-    I2CMaster4::new(
-        &clocks.sercom4_core(&gclk0).unwrap(),
-        bus_speed.into(),
-        sercom4,
-        mclk,
-        sda.into_pad(port),
-        scl.into_pad(port),
-    )
-}
```
tarcieri: With multiple I2C devices I thought it'd make sense to move this into the corresponding pins type. It certainly helps shave off a few arguments!

boards/trellis_m4/src/pins.rs
```diff
@@ -36,6 +44,41 @@ pub struct Accelerometer {
     pub scl: Pa13<Input<Floating>>,
 }
 
+impl Accelerometer {
```
tarcieri: Potentially we could add some generics with default values to this type to allow for use with SERCOMs other than 2...

boards/trellis_m4/src/lib.rs
sajattack: Could you please re-export the new i2c_master function in it's place?

tarcieri: Done

boards/trellis_m4/src/pins.rs
tarcieri: Tried that and wasn't happy with the result. SERCOM4 is the only other one compatible with the accelerometer pins, and the original `i2c_master` function is already hardcoded to that...

PR #48, opened 2019-03-22
trellis_m4: Orientation detector

tarcieri:
~~NOTE: This P4 includes a commit from #47 (ADXL343 accelerometer support) which should get merged first.~~

This PR adds a simple orientation detector based on the ADXL343 accelerometer.

When the device is upright in a portrait orientation, or slightly tilted (either handheld, or in the stand included in the kit), attempts to detect the direction the plugs are facing (either up, down, or unknown if the data is ambiguous).

Uses a moving average to smooth out the noisy accelerometer data. The formula approximating for the device's orientation is a simple function of the accelerometer data and was calibrated empirically by testing the NeoTrellis M4 device in several positions and recording the accelerometer state in each position.

Could probably use some fine-tuning, but otherwise seems to work relatively (and perhaps surprisingly?) well. To the extent it's glitchy, it seems about as glitchy as a smartphone.

ezgif-1-6df0dbd345cb

### DIFF
--- a/boards/trellis_m4/src/orientation.rs
+++ b/boards/trellis_m4/src/orientation.rs
@@ -1,0 +1,180 @@
+//! Orientation tracking support for the NeoTrellis M4
+//!
+//! This module provides tablet-like orientation tracking. It's designed to
+//! work when the device is being handheld, in a "portrait" orientation,
+//! and attempts to detect whether the device's plugs (USB, audio, etc)
+//! are pointed up or down.
+
+pub use adxl343::Ax3;
+
+use crate::Adxl343;
+use core::ops::{Index, IndexMut};
+use hal::sercom::I2CError;
+
+/// Number of samples used to guess the device's orientation
+const NUM_SAMPLES: usize = 32;
+
+/// Denominator to normalize acceleration rates between 0.0 and 1.0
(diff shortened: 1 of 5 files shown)


boards/trellis_m4/src/orientation.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Device's plugs (USB, audio, etc) are pointed down
+    Down,
+
+    /// Unable to determine device's orientation from the current data
+    Unknown
+}
+
+impl Orientation {
+    /// Attempt to guess the orientation from the accelerometer data
```
tarcieri: It really feels like there's a fairly trivial formula the implementation below is an extremely coarse grained approximation of. I will see if I can figure it out, but it seems like it should be possible to compute a single number which represents the devices up/down orientation  using the accelerometer data as inputs, and then it can be sliced into the appropriate `Orientation` based on its value.

tarcieri: I will say I briefly looked at papers regarding this particular problem. There are a lot of them! Didn't find any specific magic formula, but it looks like I can probably convert the logic table below into corresponding arithmetic operations.

tarcieri:
Closing in favor of #51, with a more powerful implementation which is contained in the `accelerometer` crate

PR #50, opened 2019-04-03
PWM Implementation

sajattack:


sajattack:
I'm not sure what to do about the pinout unused warning, we want to own the pin for the duration of the pwm, but we're not actually using it in code.

sajattack:
Should we only implement PwmPin and scrap Pwm? We only have one Channel.

sajattack:
I've decided to move everything into PwmPin.

PR #52, opened 2019-04-04
 trellis_m4: Rust 2018 edition updates

tarcieri:
Updated using `cargo fix --edition` and `cargo fix --edition-idioms`, then removing some unnecessary `crate::` prefixes on imports coming from other crates.

This also enables moving the `define_pins!` call into the `pins` module (while still re-exporting `pins::Pins` from the crate root).

Additionally runs `rustfmt` over the code.

### DIFF
--- a/boards/trellis_m4/examples/neopixel_accel.rs
+++ b/boards/trellis_m4/examples/neopixel_accel.rs
@@ -3,8 +3,8 @@
 #![no_std]
 #![no_main]
 
-extern crate panic_halt;
-
+#[allow(unused_imports)]
+use panic_halt;
 use trellis_m4 as hal;
 use ws2812_nop_samd51 as ws2812;
 
--- a/boards/trellis_m4/examples/neopixel_accel.rs
+++ b/boards/trellis_m4/examples/neopixel_accel.rs
@@ -3,11 +3,10 @@
 #![no_std]
 #![no_main]
 
-extern crate cortex_m;
 extern crate panic_halt;
-extern crate smart_leds;
-extern crate trellis_m4 as hal;
-extern crate ws2812_nop_samd51 as ws2812;
+
+use trellis_m4 as hal;
+use ws2812_nop_samd51 as ws2812;
 
 use hal::adxl343::accelerometer::Accelerometer;
 use hal::prelude::*;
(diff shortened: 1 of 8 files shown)


boards/trellis_m4/examples/neopixel_accel.rs
```diff
@@ -9,10 +9,10 @@ extern crate smart_leds;
 extern crate trellis_m4 as hal;
 extern crate ws2812_nop_samd51 as ws2812;
```
sajattack: I think you should get rid of these extern crates as part of 2018 edition upgrade.

tarcieri: Absolutely... not sure why `cargo fix --edition-idioms` missed them

tarcieri: I got rid of the trivially unnecessary ones in 5561a3e7c.

We still need `extern crate panic_halt` to define a `#[panic_handler]`:

```
   Compiling trellis_m4 v0.1.2 (/Users/bascule/src/atsamd/boards/trellis_m4)
error: `#[panic_handler]` function required, but not found
```

I can go ahead and get rid of the use of aliased crate names in the examples at least...

tarcieri: Took another stab in 9ce402212 and force pushed. That should be all of them except for the `extern crate panic_halt`, which if I try to replace with just `use panic_halt` I get the following:

```
warning: unused import: `panic_halt`
 --> examples/neopixel_rainbow.rs:4:5
  |
4 | use panic_halt;
  |     ^^^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default
```

sajattack: I think you can #[allow(unused)] on there.

tarcieri: Done in 1698bd484

djmcgill:
How is this working out? If you're happy with it I'll do the same for some other the other boards

sajattack:
@djmcgill Go for it. Wez and I have the hal and the metro's done in #37

PR #54, opened 2019-04-16
Feather M4 Express board support

tedsta:
This PR adds support for Adafruit's Feather M4 Express board. It also contains my quick attempt at getting the USB cdc_acm example working.

Paging @wez! I'll read the USB chapters of both the samd21 and samd51 datasheets to try to spot differences, but in the meantime maybe you'll notice something obvious that I missed?

### DIFF
--- a/boards/feather_m4/.cargo/config
+++ b/boards/feather_m4/.cargo/config
@@ -1,0 +1,9 @@
+# vim:ft=toml:
+[target.thumbv7em-none-eabihf]
+runner = 'arm-none-eabi-gdb'
+rustflags = [
+  "-C", "link-arg=-Tlink.x",
+]
+
+[build]
+target = "thumbv7em-none-eabihf"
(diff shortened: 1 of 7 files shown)


sajattack:
Your upload script contains a path that'll only work for your system. Did you add this by mistake?

tedsta:
Oops yes that was a mistake

sajattack:
Also, usb-device was previously commented out because it was preventing us from publishing to crates.io.

tedsta:
Sure that sounds good.

sajattack:
It looks like the neopixel pin is missing?

wez:
re: usb, yeah, leave it commented out for now.  There's been some activity upstream that we need to sync up with that may unblock enabling that on crates.io, but I haven't had bandwidth to look at it for a while.

sajattack:
@tedsta, on most of the other boards, the pin is marked as "neopixel", not dN

sajattack:
Since there's been no response on this for a week, I'll just merge it and fix it myself.

PR #55, opened 2019-04-28
Add basic support for Adafruit's PyPortal board

Shella:
Neopixel rainbow example: ezgif-4-78dcc00b243d


Shella:
Can you help us figure out the pins for the display touchscreen and WiFi/ESP32?

sajattack:
The pins are labelled pretty well in the circuitpython board definition https://github.com/adafruit/circuitpython/blob/master/ports/atmel-samd/boards/pyportal/pins.c

boards/pyportal/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+use atsamd_hal as hal;
+
+#[cfg(feature = "rt")]
+pub use cortex_m_rt::entry;
+
+pub use crate::pins::Pins;
+pub use hal::target_device::*;
+pub use hal::*;
+
+use gpio::{Floating, Input, Port};
```
sajattack: Travis is complaining that Input is unused here.

boards/pyportal/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+use hal::clock::GenericClockController;
+use hal::sercom::{I2CMaster5, PadPin, SPIMaster2};
+use hal::time::Hertz;
+
+#[cfg(feature = "usb")]
+pub use hal::usb::UsbBus;
+#[cfg(feature = "usb")]
+use usb_device::bus::UsbBusWrapper;
+
+/// Convenience for setting up the 2x3 header block for SPI.
```
sajattack: This line of comment is referencing the 2x3 pin header on the Metro, and should be removed.

boards/pyportal/src/lib.rs
sajattack: Oh, I know why, it's because it's unproven.

sajattack: lol nvm it was complaining about trellis anyway @tarcieri

PR #56, opened 2019-04-30
Implement ADC HAL

### LINKED ISSUE #35: Implement ADC abstraction
Embedded-hal has an ADC trait, we have an ADC, let's figure out how to use it.

sajattack:
SAMD51 only at the moment, though I'll start on SAMD21 soon. The main motivator for this was pyportal touchscreen. Also makes minor changes to clocks, because I was in there adding the adc clock and found some stuff that needed changing.

Fixes #35

### DIFF
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -74,3 +74,6 @@ name = "blinky_basic"
 
 [[example]]
 name = "pwm"
+
+[[example]]
+name = "adc"
--- a/boards/feather_m0/examples/adc.rs
+++ b/boards/feather_m0/examples/adc.rs
@@ -12,7 +12,8 @@ extern crate embedded_hal;
 
 use hal::clock::GenericClockController;
 use hal::prelude::*;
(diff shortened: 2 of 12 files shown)


sajattack:
So one thing that's bugging me here is the samd21s only have one adc, and it's just called "adc" in the pac. In the current structure that means we can't name our hal for the adc "adc". So I named it "adc_hal" in samd21 and "adc" in samd51. I think that's pretty jank and bad. So we could either name them both "adc_hal" or it might be best to do a big restructure and put the pac in it's own module separate from the hal. Like I think nrf-rs has where the hal contains the pac under a "pac" module rather than having it all jumbled together. Let me know what you guys think.

tarcieri:
@sajattack there's a big refactor I had in mind which I can attempt. I was hoping to land #60 first, but perhaps I should just try the refactor first instead

sajattack:
@tarcieri Yeah I have to rewrite like half the HAL anyway if I want upstream svd2rust to work so it's a good time for refactors.

PR #57, opened 2019-05-04
Pyportal fixups

Shella:
Published the [`atsamd51j20a`](https://crates.io/crates/atsamd51j20a) crate. Also attempted to publish `pyportal` but ran into the following error. Think `atsamd-hal` needs a release?

```
error: failed to verify package tarball

Caused by:
  failed to select a version for `atsamd-hal`.
    ... required by package `pyportal v0.1.2 (/Users/shella/codez/atsamd/boards/pyportal/target/package/pyportal-0.1.2)`
versions that meet the requirements `~0.4` are: 0.4.1, 0.4.0

the package `pyportal` depends on `atsamd-hal`, with features: `samd51j20a-rt, samd51j20a` but `atsamd-hal` does not have these features.


failed to select a version for `atsamd-hal` which could resolve this conflict
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,9 @@ There are a couple of crates provided by this repo:
   auto-generated crate providing access to the peripherals
   specified for this device by its SVD file.  This is the MCU used in the Trinket M0
   and Gemma M0 boards from Adafruit.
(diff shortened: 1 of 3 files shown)


sajattack:
Yeah I'll have to update the HAL.

PR #59, opened 2019-05-08
[WIP] hal: dma::controller::Registers trait

tarcieri:
Defines a `Registers` trait which abstracts over the DMAC registers, allowing DMA drivers to be written targeting this trait, and then used with any relevant PAC which already maps them with svd2rust.

The trait is defined differently for SAMD21 vs SAMD51 as they have different DMAC registers, but usable with both. An additional higher-level abstraction trait will have to be written to abstract across the differences between the two.

This approach is relatively macro-heavy, but the result should enable the rest of the DMA subsystem to be written sans macros by leveraging the newly defined trait.

### DIFF
--- a/boards/trellis_m4/src/lib.rs
+++ b/boards/trellis_m4/src/lib.rs
@@ -16,13 +16,13 @@ pub use cortex_m_rt::entry;
 pub use hal::{target_device::*, *};
 pub use pins::Pins;
 
-use gpio::{Input, Port};
+use gpio::Port;
 use hal::clock::GenericClockController;
 use hal::sercom::I2CMaster4;
 use hal::time::Hertz;
 
 #[cfg(feature = "keypad-unproven")]
-use hal::gpio::{OpenDrain, Output, PullUp};
+use hal::gpio::{Input, OpenDrain, Output, PullUp};
 #[cfg(feature = "keypad-unproven")]
 use keypad::{keypad_new, keypad_struct};
 
--- a/hal/src/dma.rs
+++ b/hal/src/dma.rs
@@ -1,0 +1,4 @@
+//! Direct Memory Access
+
+// TODO: remove `pub` when we have a DMA driver to leverage this
+pub mod controller;
--- a/hal/src/dma/controller.rs
+++ b/hal/src/dma/controller.rs
@@ -1,0 +1,156 @@
+//! Direct Memory Access Controller (DMAC)
+//!
+//! This module provides a `dma::controller::Registers` trait (defined
+//! differently for SAMD21 vs SAMD51) which abstracts over the DMAC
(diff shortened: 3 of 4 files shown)


hal/src/dma/controller.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub trait Registers {
+        // Define accessors for registers common to SAMD21 and SAMD51
+        dmac_common_registers!();
+    }
+
+    #[macro_export]
+    macro_rules! impl_dmac_registers {
+        ($dmac:ty) => {
+            impl $crate::dma::controller::Registers for $dmac {
+                impl_dmac_common_registers!();
```
tarcieri: TODO: SAMD51 DMA channels. I figured I'd cross-check this basic approach before working on those.

hal/src/dma.rs
```diff
@@ -0,0 +1,3 @@
+//! Direct Memory Access
+
+pub mod controller;
```
tarcieri: This module is empty right now, but the goal is to put the DMA driver in here. However, it needs some way of accessing the DMAC registers first.

hal/src/lib.rs
```diff
@@ -72,6 +72,8 @@ pub mod pwm;
 pub mod pwm;
 
 pub mod delay;
+#[cfg(feature="unproven")]
+pub mod dma;
```
tarcieri: Gated this under `unproven` for now, but perhaps it could use its own cargo feature, e.g. `dma`

sajattack:
Why are you defining things at the board level? Shouldn't it all be HAL level? Also, I wonder if we should modify the SVD to define the registers, so they are PAC level.

tarcieri:
@sajattack the DMAC registers are defined by the PAC. It'd be great if we could do this at the PAC level, but as far as I can tell the PAC is 100% `svd2rust`, and doesn't directly depend on the HAL

boards/trellis_m4/src/lib.rs
```diff
@@ -29,6 +31,11 @@ use keypad::{keypad_new, keypad_struct};
 /// Number of Neopixels on the device
 pub const NEOPIXEL_COUNT: usize = 32;
 
+/// Define the `DmacRegisters` type which impl's the `hal::dma::controller::Registers`
+/// trait and can be used with a DMA driver.
+#[cfg(feature = "unproven")]
+decl_dmac_registers!(DmacRegisters, dmac::DMAC, dmac::RegisterBlock);
```
tarcieri: This macro is defining a newtype, `DmacRegisters`, that wraps the `svd2rust`-generated DMAC registers, and then impls the `hal::dma::controller::Registers` trait.

It can't impl the trait directly on `DMAC`, because `DMAC` is defined in the PAC and the trait is defined in the HAL.

Issue #25, opened 2018-11-03
Implement DMA abstraction


sajattack:
Title kinda speaks for itself. I'll try to work on this but will probably need help.

sajattack:
Resources:
http://ww1.microchip.com/downloads/en/AppNotes/Atmel-42257-SAM-Direct-Memory-Access-Controller-Driver-DMAC_ApplicationNote_AT07683.pdf
https://github.com/adafruit/adafruit_zerodma
https://github.com/rust-embedded/embedded-hal/issues/37

sajattack:
How closely should we follow the ASF4 API?

wez:
I haven't really looked at DMA on this device in detail.  I recall reading the blog posts referenced from the embedded-hal issue you referenced, and have been soft-following that issue.

My currently under-educated opinion is that we should feel free to produce an implementation that works for us (eg: we don't have to wait for a trait to appear in embedded-hal), and that implementation should be reasonably rust idiomatic and safe rather than mirroring an unsafe C API that exists elsewhere.

That said, I haven't really looked at the DMA APIs for C/C++ on this device so I don't know whether I'm being too much of a rust snob, and it can be better to start out by building something (or *anything*!) that does the job so that there is a concrete implementation as a starting point that we can then comment on as a PR and give suggestions in review, rather than blocking on consensus in this issue.

After skimming a couple of things, I have a couple of questions on my mind that would influence how I might choose to design the API, so I'm going to ask them here; I'd like to stress that I still don't know enough to have a strong influence on direction.

### DMA Channels

* Is a channel allocated for a specific purpose for the lifetime of the firmware, or is it allocated for a small number (which may be 1) requests before being released and being eligible for re-purposing for different transfers?
* In light of the above, how do we safely and conveniently obtain a channel and release it?

Our clock allocation is pretty robust and type safe and is one model that we could potentially apply to this, but I don't know if is convenient enough to translate to DMA.

### Callbacks

* Do we *have* to use callbacks?  I can see a bunch of them in https://gist.github.com/sajattack/5080311aa5085f1a9748c711bea78140
[...]

PR #60, opened 2019-05-08
Port Adafruit ZeroDMA to Rust

tarcieri:
This is a translation of [Adafruit ZeroDMA] (C++)  into Rust.

It's not a completely faithful translation, and includes a lot of guesswork, but should provide a starting point for getting an initial DMA implementation working.

[Adafruit ZeroDMA]: https://github.com/adafruit/Adafruit_ZeroDMA


hal/Cargo.toml
```diff
@@ -16,7 +16,7 @@ embedded-hal = "~0.2"
 nb = "~0.1"
 paste = "~0.1"
 bitfield = "~0.13"
-vcell = "~0.1"
+vcell = { version = "~0.1", features = ["const-fn"], git = "https://github.com/japaric/vcell.git" }
```
sajattack: git deps aren't allowed on crates.io

tarcieri: Yeah, I was originally using the git version because `const-fn` support hasn't been published yet, however that was before I was able to leverage the DMAC registers defined in the PAC, so this should be completely unnecessary. I just have to wire everything up (at which point I can remove this)

tarcieri:
Note that there's a lot of low-hanging refactoring work that could happen here, particularly around code that looks like this:

```rust
        // Reset the allocated channel
        #[cfg(feature = "samd51")]
        {
            self.dmac
                .chctrla(channel)
                .write(|reg| reg.enable().clear_bit().swrst().set_bit());
        }

         #[cfg(not(feature = "samd51"))]
        unsafe {
            self.dmac.chid.write(|reg| reg.id().bits(channel));
            self.dmac
                .chctrla
                .write(|reg| reg.enable().clear_bit().swrst().set_bit());
        }
```

I've left it like this deliberately so as to keep the translated code as structurally similar as possible to the real code for review (allowing easy side-by-side review of the translation).

After we're sure the code is working, it seems like it'd be good to extract a `struct Controller` or thereabouts with a separate SAMD21 and SAMD51 implementation with named functions for each of these things that has a varying implementation.

tarcieri:
I'm calling this ready for initial review. It definitely doesn't work, but at least it compiles!

I'd suggest looking at 23b803cc6fa0ee64ca661e9ee68a723cc52b564e for a side-by-side view of the original C++ and my translation.

I will attempt to flag outstanding problems with review notes.

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+    trigger_action: TRIGACTW,
+
+    /// Callbacks
+    callbacks: Callbacks,
+
+    /// Descriptor list
+    descriptor_list: &'desc DescriptorList,
+}
+
+pub struct InterruptState {
```
tarcieri: This was all (somewhat necessarily) global state in ZeroDMA. Any ideas what the Rust code should do? I shoved it all in a struct for now, but it then becomes a parameter to every method that needs to acquire a critical section.

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+            trigger_action: TRIGACTW::TRANSACTION,
+            callbacks: Callbacks::default(),
+            descriptor_list,
+        }
+    }
+
+    // TODO(adafruit): Drop impl? Should stop job, delete descriptors, free channel.
+
+    /// IRQ handler
+    pub fn irq_handler(&mut self, flags: u8) {
```
tarcieri: ZeroDMA has a global interrupt handler which looks up which channel received an interrupt. I think we'll need one too... this is a janky one that will only work for one channel (if that!).

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+                reg.baseaddr()
+                    .bits(self.descriptor_list.descriptor.as_ptr() as u32)
+            });
+
+            self.dmac.wrbaddr.write(|reg| unsafe {
+                reg.wrbaddr()
+                    .bits(self.descriptor_list.writeback.as_ptr() as u32)
+            });
+
+            // TODO(tarcieri): Wipe decriptor and writeback tables
```
tarcieri: This most likely needs to be implemented

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+            .swtrigctrl
+            .modify(|reg, value| unsafe { value.bits(reg.bits() & !(1 << channel)) });
+
+        // This type unfortunately does not impl Copy or Clone
+        let peripheral_trigger = match self.peripheral_trigger {
+            TRIGSRCW::DISABLE => TRIGSRCW::DISABLE,
+        };
+
+        // Ditto for this type
+        let trigger_action = match self.trigger_action {
```
tarcieri: These are annoying and ugly as `TRIGSRCW` and `TRIGACTW` don't impl `Copy` or `Clone`. Any ideas how to work around that? A newtype maybe?

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+    /// to be called before job is started.
+    pub fn set_callback(&mut self, cb_type: CallbackType, cb: fn(&mut DMA)) {
+        match cb_type {
+            CallbackType::TransferError => self.callbacks.transfer_error = Some(cb),
+            CallbackType::TransferDone => self.callbacks.transfer_done = Some(cb),
+            CallbackType::ChannelSuspend => self.callbacks.channel_suspend = Some(cb),
+        }
+    }
+
+    // TODO(adafruit): Suspend/resume don't quite do what I thought -- avoid using for now.
```
tarcieri: Should we even keep these around if they're not recommended to be used?

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+
+        // Scan descriptor list to find last entry.  If an entry's
+        // DESCADDR value is 0, that's the end of the list and it's
+        // currently un-looped.  If the DESCADDR value is the same
+        // as the first entry, that's the end of the list and it's
+        // looped.  Either way, set the last entry's DESCADDR value
+        // to the new descriptor, and the descriptor's own DESCADDR
+        // will be set later either to 0 or the list head.
+        let desc = if self.has_descriptors {
+            // TODO(tarcieri): allocate new descriptor
```
tarcieri: No idea what to do here...

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+                            + (desc.btcnt.get() * bytes_per_beat) as u32 * (1 << step_size),
+                    );
+                } else {
+                    desc.dstaddr
+                        .set(desc.dstaddr.get() + (desc.btcnt.get() * bytes_per_beat) as u32);
+                }
+            }
+        }
+
+        // TODO(adafruit): I think this code is here by accident -- disabling for now.
```
tarcieri: Should we just delete this code as well?

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+    pub fn is_active(&self) -> bool {
+        if let Some(channel) = self.channel {
+            self.descriptor_list.writeback(channel).btctrl.get() & BTCTRL_VALID != 0
+        } else {
+            false
+        }
+    }
+}
+
+/// Thunk for `__get_PRIMASK()`
```
tarcieri: TODO: `__get_PRIMASK` equivalent

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+fn get_primask() -> u32 {
+    // TODO(tarcieri): implement this!
+    unimplemented!();
+}
+
+/// Thunk for `__enable_irq()`
```
tarcieri: TODO: `__enable_irq()` equivalent

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+fn enable_irq() {
+    // TODO(tarcieri): implement this!
+    unimplemented!();
+}
+
+/// Thunk for `__disable_irq()`
```
tarcieri: TODO: `__disable_irq()` equivalent

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+fn disable_irq() {
+    // TODO(tarcieri): implement this!
+    unimplemented!();
+}
+
+/// Thunk for `__DMB()`
```
tarcieri: TODO: `__DMB()` equivalent

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+fn dmb() {
+    // TODO(tarcieri): implement this!
+    unimplemented!();
+}
+
+/// Thunk for `NVIC_EnableIRQ`
```
tarcieri: TODO: `NVIC_EnableIRQ` equivalent

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+fn nvic_enable_irq(_irq: IRQn) {
+    // TODO(tarcieri): implement this!
+    unimplemented!();
+}
+
+/// Thunk for `NVIC_DisableIRQ`
```
tarcieri: TODO: `NVIC_DisableIRQ` equivalent

hal/src/dma.rs
```diff
@@ -0,0 +1,1075 @@
...
+fn nvic_disable_irq(_irq: IRQn) {
+    // TODO(tarcieri): implement this!
+    unimplemented!();
+}
+
+/// Thunk for `NVIC_SetPriority`
```
tarcieri: TODO: `NVIC_SetPriority` equivalent

hal/src/dma/controller.rs
```diff
@@ -0,0 +1,270 @@
+//! DMA controller (DMAC)
+
+use vcell::VolatileCell;
+
+/// (DMAC_BTCTRL) Source Address Increment Enable
+pub(super) const BTCTRL_SRCINC: u16 = 1 << 10;
```
tarcieri: It'd be nice to get rid of all of these consts and replace them with things from the SVD, or failing that proper `bitfield` types.

A lot of them are for DMAC descriptor fields.

hal/src/dma/controller.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// 16-bit
+    HWord = 1,
+
+    /// 32-bit
+    Word = 2,
+}
+
+/// DMA descriptor
+#[repr(align(16))]
+pub struct Descriptor {
```
tarcieri: Is this the best way to define the DMAC descriptor struct, or can it be defined via SVD somehow?

hal/src/dma/controller.rs
```diff
@@ -0,0 +1,270 @@
...
+            srcaddr: VolatileCell::new(0),
+            dstaddr: VolatileCell::new(0),
+            descaddr: VolatileCell::new(0),
+        }
+    }
+}
+
+/// Descriptor and writeback sections for all DMAC channels
+#[derive(Default)]
+pub struct DescriptorList {
```
tarcieri: TODO: Map DMAC descriptors ala https://github.com/sajattack/atsamd/blob/9102fe969e6d85639ec70faf2c544fe4691ffc2c/hal/src/dma.rs#L229

hal/src/dma.rs
```diff
@@ -26,721 +26,1050 @@
...
+    pub fn allocate(
+        &mut self,
+        clock_source: &mut ClockSource,
+        int_state: &InterruptState,
+    ) -> Status {
+        if self.channel.is_some() {
+            return Status::Ok; // Already alloc'd!
+        }
+
+        // Find index of first free DMA channel.  As currently written,
```
sajattack: This part of automatically finding a free channel felt a bit adafruit-y to me. Maybe for more control we should let the user figure it out? Idk, see what other rust dma implementations are doing.

hal/src/dma.rs
sajattack: https://atsamd-rs.github.io/atsamd/atsamd51j19a/atsamd51j19a/struct.NVIC.html

sajattack: https://docs.rs/cortex-m/0.2.12/cortex_m/interrupt/index.html

hal/src/dma.rs
sajattack: https://docs.rs/cortex-m/0.2.12/cortex_m/asm/fn.dmb.html

hal/src/dma.rs
sajattack: https://docs.rs/cortex-m/0.6.0/cortex_m/register/primask/index.html

hal/src/dma.rs
tarcieri: Aha, in that case, would it be better to substitute a Rust memory fence instead?

According to Godbolt[1] this program:

```
#![no_std]

use core::sync::atomic;

pub fn test() {
    atomic::fence(atomic::Ordering::SeqCst);
}
```

...generates this thumbv7 assembly:

```
example::test:
        dmb     sy
        bx      lr
```

[1]: https://godbolt.org/z/2SnyjA

tarcieri: I get the same ASM for:

```
atomic::fence(atomic::Ordering::Acquire);
```

...which I'm already using elsewhere (cargo culted from the ST DMA driver)

hal/src/dma/controller.rs
sajattack: https://github.com/sajattack/atsamd/blob/9102fe969e6d85639ec70faf2c544fe4691ffc2c/hal/src/dma.rs#L183

hal/src/dma.rs
tarcieri: Updated to use these

hal/src/dma.rs
tarcieri: Changed to take an explicit `channel_id`, and return `Err(Error::Busy)` if it's in use

hal/src/dma.rs
tarcieri: Switched to cortex_m::interrupt::Mutex

hal/src/dma.rs
tarcieri: Deleted (but in the Git history for posterity)

hal/src/dma.rs
tarcieri: Deleted (but kept in git history)

hal/src/dma.rs
tarcieri: The need for this was eliminated by switching to `cortex_m::interrupt::Mutex`

tarcieri:
This still needs to get wired up to a `cortex_m` crate `#[interrupt]` handler, but is pretty close to "complete" (although almost certainly buggy/broken).

Any ideas how to test it? I guess the easiest might be a memory-to-memory transfer.

sajattack:
Is it possible to just let the user define the interrupt handler? I would try DMA'ing a big chunk of data over a slow spi and flashing leds at the same time or something. So you know it's async.

hal/src/dma/channel.rs
```diff
@@ -0,0 +1,584 @@
...
+        self.job_status.set(Err(Error::Aborted));
+    }
+
+    /// Set DMA peripheral trigger.
+    /// This can be done before or after channel is allocated.
+    pub fn set_trigger(&mut self, trigger: TRIGSRCW) {
+        let dmac = self.controller.lock();
+
+        // Save value for `allocate()`
+        self.peripheral_trigger = match self.peripheral_trigger {
```
Gentoli, line 327:
```suggestion
        self.peripheral_trigger = match trigger {
```

also on `master`, in the newly generated pac `TRIGSRCW` -> `TRIGSRC_A` and has copy

tarcieri: @Gentoli thanks. this PR is super out of date now though! Any interest in helping pick it back up?

tarcieri:
If anyone else is interested in picking this back up, here's some earlier work I did on SERCOM DMA support:

https://gist.github.com/tarcieri/61d5789e369cf5e25edc4248d26b525a

Never got this merged into the Adafruit translation.

tarcieri:
Seems like DMA support should likely leverage this in some way: https://github.com/rust-embedded/embedded-dma

sajattack:
Should this PR be closed you think @tarcieri ?

tarcieri:
@sajattack yeah, probably. It's way too out of date, but hopefully this PR and https://gist.github.com/tarcieri/61d5789e369cf5e25edc4248d26b525a could help guide someone else who wants to write a DMA implementation

jessebraham:
Any further discussion can take place here: https://github.com/atsamd-rs/atsamd/issues/25

PR #62, opened 2019-05-10
Move the build-all logic into .travis.yml config

keithduncan:
To help find test failures in https://github.com/atsamd-rs/atsamd/pull/61 quicker I’ve added a build matrix to the travis config.

This builds each board in parallel with stable and nightly rust using the same features and examples that were used in the `build-all.sh` script.


keithduncan:
I’ve seen linking fail in some of these test runs:

```
rust-lld: error: 
          There can't be more than 32 interrupt handlers. This may be a bug in
          your device crate, or you may have registered more than 32 interrupt
          handlers.
```

But I haven’t made any code changes, does anyone know what would cause that?

wez:
It's valuable to be able to build all the boards locally.  With the script removed, that now becomes more difficult.

As a counter proposal, how about changing the build-all script so that it has a flag that will update the travis config with the equivalent set of builds that build-all would run?

That would allow for both local and CI workflows to run with the same single list of builds

wez:
Also, FWIW, Travis has fairly limited concurrency, so even though the matrix allows expressing parallelized build steps, it is rare to get more than 1-2 instances running on a build.  If you look at the overall build time in the build history on travis you'll see that this PR took 59 minutes, which is 15 minutes less than the cron'd builds.  It's a reduction but not a massive reduction.
It is definitely nice to be able to see the status on a per-board level though.

sajattack:
The doc build alone takes over 50 minutes, which is the travis timeout. That's why our builds have been failing.

keithduncan:
> It's valuable to be able to build all the boards locally. With the script removed, that now becomes more difficult.

That’s a good point, I don’t know of any command line tool that reads/runs travis-ci config 🤔 

> As a counter proposal, how about changing the build-all script so that it has a flag that will update the travis config with the equivalent set of builds that build-all would run?

I think this sounds good, just to check I’m following you: there would be a single source of truth for board/examples/features combos to build, build-all would use that and there’d be a way to switch the travis config from using build-all directly or having a build matrix that hits all the combos?

keithduncan:
It looks like rust nightly wasn’t looking up the directory tree for `.cargo/config`, this was causing the “more than 32 interrupts error”. I’ve added the config from root into the each board crate directory to fix that.

The remaining error on link for symbol `__aeabi_uidiv` appears in [master](https://travis-ci.org/atsamd-rs/atsamd/jobs/530858216) builds for Cortex-M0 based boards when built with nightly too.

wez:
> I think this sounds good, just to check I’m following you: there would be a single source of truth for board/examples/features combos to build, build-all would use that and there’d be a way to switch the travis config from using build-all directly or having a build matrix that hits all the combos?

My thinking was to switch build-all over to a python script that has a list of the things that we want to build (effectively what you have currently in that big `env` block in this PR), and then it could have two modes:

* The default mode is to simply iterate the list and construct a command line using eg: the `script` section in this PR as a template and `subprocess.check_call` on each one; that would replace the current local use case for build-all
* There could be an `--update-travis-config` mode that rewrites the `env` and `script` sections of the travis.yml from the list.  Each time we make a change to the build-all script, we'd want to run it in `--update-travis-config` mode so that it updates the travis config as part of that PR.  That way we get your nicer travis config and the ability to run all those same builds locally.

For managing the travis yaml, I would probably opt for something simple like marking up the file with comments:

```
### BEGIN: DO NOT EDIT THIS BLOCK BY HAND: run build-all.py --update-travis-config
env:
   - BOARD=metro_m0 EXAMPLES="--example=blinky_basic --example=blinky_rtfm"
  ...
script:
  - "cd boards/$BOARD"
  - "cargo build ${EXAMPLES:---examples} $FEATURES"
### END: DO NOT EDIT THIS BLOCK BY HAND
```

and the `--update-travis-config` script could look for the start and end delimiters when it reads the text and discard the contents, replacing them with its updated version.

keithduncan:
I’ve taken a look at a combination Python + Travis build-all.

Instead of keeping the authoritative list of builds in the python script and updating the travis file I’ve done the opposite for now, make `build-all.py` read the travis file to find crates to build. I think this is neater since there’s less to keep up to date, the build list can’t get out of sync by forgetting to run an `--update-travis-config`.

I vendored the PyYaml pip for Python 2.7 and 3.7, I have no idea if this is the right way to do this since I haven’t used Python previously. 🚁 🐕 This is working locally for me with the default macOS Python install running `PYTHONPATH=vendor/lib/python2.7/site-packages/ python build-all.py` but let me know if there’s a better way I could be packaging this. I felt that using a native yaml parser instead of writing a partial yaml encoder to update the travis config without a library would have less defects over time. Let me know what you think 😄

sajattack:
I would recommend setting up a requirements.txt for all the python packages you need, and then we can set up a virtualenv.

keithduncan:
> I would recommend setting up a requirements.txt for all the python packages you need, and then we can set up a virtualenv.

I’ve replaced the vendored copies with a requirements.txt :+1:

> Also, maybe we should split up the doc builds as well?

I think this is harder to parallelise due to `ghp-import`, we’d need to collect the docs built in parallel somewhere to push the aggregate docs to the gh-pages branch.

sajattack:
Ok, my biggest concern right now is the documentation build has been failing recently due to timeouts. 

Also, several of your jobs failed (no fault on you, nightly is broken on thumbv6m currently), but the job succeeded. Maybe adjust it so it fails the whole build if a job fails?

keithduncan:
> Also, several of your jobs failed (no fault on you, nightly is broken on thumbv6m currently), but the build succeeded. Maybe adjust it so it fails the whole build if a job fails?

I added this intentionally because the `nightly` toolchain is unstable, right now the GitHub pull request commit status doesn’t show whether a particular change set is broken because of the changes being proposed or because of a breakage in `nightly`.

I think it’s useful to exclude these failures, but I’m happy to remove this failure tolerance if you want the aggregate status to reflect nightly failures too.

sajattack:
Alright, good enough for me, @wez, you good?

wez:
Looks good to me; just one easy final request: can you add a brief note to the readme that shows how to run the new `build-all.py` script with the yaml bits loaded?

Something like this inserted at the end of the section on building, but before the bit about getting code onto the device:

````
## Building everything locally

If you'd like to build all the same things that the CI would build but on
your local system, you can run:

```
$ something something virtualenv, something something pip
$ build-all.py
``` 
````

Thanks!

keithduncan:
Done!

wez:
Thanks!

PR #67, opened 2019-06-26
Adds a convenience method to enable UART on the Trellis M4

fionawhim:
Renames the I2C type to JST to reflect that it is for the NeoTrellis’s
sole output pins, which are in a JST connector. The screen printing
is still "SDA" and "SCL" so that’s still what the fields are called.

The UART methods are based on the existing i2c_master methods, though
one of them may be redundant.

If possible, this will be the basis for adding UART on its default
pins for the other boards in this repo.


sajattack:
LGTM. Are you planning to add more or should I merge now?

fionawhim:
I think merge now and I’ll follow up with more devices over the next few
days.

Issue #69, opened 2019-06-28
Upgrade digital input/output pins to embedded_hal::digital::v2


sajattack:
As of embedded_hal [0.2.3](https://docs.rs/embedded-hal/0.2.3/embedded_hal/), there is a new fallible (returns a Result) version of the [digital traits](https://docs.rs/embedded-hal/0.2.3/embedded_hal/digital/index.html) and the non-fallible version we are using is deprecated. It looks like there are compatibility shims to upgrade v1 (the version we're using) to v2, but I don't know all the details. As usual, probably a good idea to see what other HALs like [stm32f1xx](https://github.com/stm32-rs/stm32f1xx-hal) and [nrf52](https://github.com/nrf-rs/nrf52-hal) are doing.

sajattack:
cc @fionawhim

fionawhim:
As far as I can tell we’re pioneers in switching to v2. I don’t _think_ there are error states? So mostly this is just type wrangling.

BenBergman:
IIRC, the failing trait was added so that pins on IO expanders could be used interchangeably with direct IO when passed to drivers while still letting drivers fail gracefully if something wasn't right.

sajattack:
Hmm, ok. Then why is v1 deprecated?

sajattack:
The page for v2 says "Infallible implementations should set Error to !."

fionawhim:
@sajattack `!` as a type would force us into nightly from what I can tell. Would `()` be ok for `Error`?

sajattack:
I suppose so. Maybe we're jumping into this too early.

fionawhim:
Well, I can make the branch and we can see the consequences.

PR #71, opened 2019-06-29
Upgrades to digital::v2 GPIO interface

fionawhim:
New interface returns a Result, old interface was deprecated.

Fixes atsamd-rs/atsamd#69

### DIFF
--- a/boards/arduino_mkrzero/examples/blinky_basic.rs
+++ b/boards/arduino_mkrzero/examples/blinky_basic.rs
@@ -24,8 +24,8 @@ fn main() -> ! {
 
     loop {
         delay.delay_ms(200u8);
-        led.set_high();
+        led.set_high().unwrap();
         delay.delay_ms(200u8);
-        led.set_low();
+        led.set_low().unwrap();
     }
 }
--- a/boards/circuit_playground_express/examples/blinky_basic.rs
+++ b/boards/circuit_playground_express/examples/blinky_basic.rs
@@ -27,8 +27,8 @@ fn main() -> ! {
     let mut delay = Delay::new(core.SYST, &mut clocks);
     loop {
         delay.delay_ms(200u8);
-        red_led.set_high();
+        red_led.set_high().unwrap();
         delay.delay_ms(200u8);
-        red_led.set_low();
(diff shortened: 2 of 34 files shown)


sajattack:
It's generally bad practice to have unwraps in a library. Also I wonder if it's possible through one of the _compat modules of the digital module to keep the old interface exposed as well as the new one, and reduce the need to change every program that's ever used a digital pin.

fionawhim:
It’s your call re: the `unwrap`s. I suppose they could be switched to `ok`s, we just need to consume the `Result`.

It *may* be the case that we could switch from `()` as the `Error` type to `Infallible`, and use an exhaustive `match` on the result in the library, which would (from what I understand) let us not include the `Err` option since `Infallible` is an enum with no choices. That way, if we ever added an `Error` to the pins that did something, there would be a compile error that the match wasn’t exhaustive that would tell us to handle it in the library code.

I don’t think we can easily expose both interfaces at the same time, because they have the same method names. So if our pins had both traits, I think consumers would need to specify which `set_high` implementation to use, which would be a lot of boilerplate.

Fixing the broken pyportal example now. It was missing an `into`.

sajattack:
Unwraps are fine for examples. Can you change the ones in `lib.rs`to tries maybe?

fionawhim:
@sajattack My understanding is that that would mean changing all of the `flash_spi_master` _&c._ functions to return `Result`s. If we did that, we could also remove the `unwrap`s that are already in there for getting the SERCOM clocks.

Is that what you’d like?

sajattack:
I think that's probably best practice but I'd like to see if anyone else has other ideas.

sajattack:
I guess unwraps might be less hassle for the end-user, and no one has gotten back to me on whether that is egregiously bad practice so f*** it I guess. You ready?

tarcieri:
IMO `unwrap()` is fine in cases where you are completely confident they will never fail, which per the comments appears to be the case here.

I use `expect()` in cases which are more like "well it's not panicking in the tests, but I'm not completely sure this scenario will be panic-free"

fionawhim:
@sajattack Cool. I think this is ready for merge, then.

PR #73, opened 2019-07-05
Don't depend on USB feature for USB calibration data.

bjc:
These address exist in the NVM calibration area always, so they should
be referrable regardless of what features are enabled.

### DIFF
--- a/hal/src/calibration.rs
+++ b/hal/src/calibration.rs
@@ -41,19 +41,16 @@ pub fn dfll48m_coarse_cal() -> u8 {
     cal_with_errata(4, 26, 0x3f, 0x3f, 0x1f) as u8
 }
 
-#[cfg(feature = "usb")]
 /// USB TRANSN calibration value. Should be written to USB PADCAL register.
 pub fn usb_transn_cal() -> u8 {
     cal_with_errata(4, 13, 0x1f, 0x1f, 5) as u8
 }
 
-#[cfg(feature = "usb")]
 /// USB TRANSP calibration value. Should be written to USB PADCAL register.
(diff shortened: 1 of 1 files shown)


sajattack:
Are they needed outside of USB though?

bjc:
No, but I needed access to them for USB stuff without pulling in the rest of the stuff behind the USB feature. Not that there's a way to do it without modifying the Cargo.toml, and addititonally, the USB stuff appears to be feature-gated because it's not working.

In general, the feature gates that I've seen weren't set up to lock out which parts of the hardware you can access but to pull in additional stuff on top of raw hardware access. I would guess that the USB feature, in particular, only exists because the stuff behind it isn't finished yet. Once it's working, I'd hope that it's no longer feature flagged, and is instead available for general use like everything else.

But for simply accessing the NVM calibration data, I don't think the feature flag is necessary or warranted.

sajattack:
Yeah, USB is feature-flagged out because it doesn't work. I think I'll leave these flags in place until USB is working.

bjc:
Sorry to necro this, but I've recently finished up a first pass on a USB host driver for the SAMD21 (https://github.com/bjc/atsamd-usb-host), and in order to publish it I need this patch as well as https://github.com/atsamd-rs/atsamd/pull/79 to expose the USB pins from the MCU.

While I can duplicate the `padcal` library, I don't think it makes a lot of sense to.

sajattack:
It won't let me reopen so I think you'll have to make a new PR

bjc:
Done: https://github.com/atsamd-rs/atsamd/pull/80

Thanks.

PR #74, opened 2019-07-06
Modify the $pincfg register to retain a pullen config for alternate modes

keithduncan:
Using `write` clears this register which means an alternate function like extint cannot use a pull up/down. Replacing this with a `modify` operation ensures the pullen config remains after moving the pin into an alternate function.

I yet haven’t reasoned through how this will interact with other values that may be set in the $pincfg register and whether the hal was relying on these being cleared.


sajattack:
image

PR #75, opened 2019-07-21
point to master svd2rust

wez:
Upstream finally merged support for SVD features used by the
ATSAMD family of parts, so that we don't need my fork any longer.
They haven't released this to crates.io yet, but we can still pull
it down from master.

This commit is large and hard to review because most of it is visual
noise.  The gist of the changes are:

* Point at svd2rust master rather than my fork
* Remove a doc warning hack
* `is_async_` got renamed to `is_async`...
* ... so update pac and hal version numbers
* Lots of `const`s for MASK and OFFSET got folded into accessors
* `::vcell` got replaced with `vcell` (eg: no colons)
* Lost of constant numbers changed from decimal to easier to grok hex


sajattack:
Finally! lol

PR #97, opened 2019-09-19
Feather m0 examples

spicyjack:
Two "GraphicsMode" demos and one "TerminalMode" demo that use an Adafruit Feather M0 with the Adafruit OLED Featherwing, which uses a SSD1306 chipset and has a resolution of 128x32 pixels.  The demos can also be run with generic SSD1306 modules that are correctly connected to the Feather M0 board.


boards/feather_m0/examples/ssd1306_terminalmode_128x32_i2c.rs
```diff
@@ -0,0 +1,111 @@
...
+        &mut clocks,
+        KiloHertz(400),
+        peripherals.SERCOM3,
+        &mut peripherals.PM,
+        pins.sda,
+        pins.scl,
+        &mut pins.port,
+    );
+
+    // NOTE the `DisplaySize` enum comes from the ssd1306 package,
```
jamwaffles: Are there other sizes in the wild that the ssd1306 crate doesn't support, or is this just a comment to clarify the code? Please [open an issue](https://github.com/jamwaffles/ssd1306/issues/new) if the former!

boards/feather_m0/examples/ssd1306_graphicsmode_128x32_i2c.rs
```diff
@@ -0,0 +1,156 @@
...
+    let y_max = 31;
+
+    // screen outline
+    disp.draw(
+        Rectangle::new(Point::new(0, 0), Point::new(x_max, y_max))
+            .stroke(Some(BinaryColor::On))
+            .into_iter(),
+    );
+
+    // triangle
```
jamwaffles: Embedded Graphics has a new [`Triangle` primitive](https://docs.rs/embedded-graphics/0.6.0-alpha.2/embedded_graphics/primitives/triangle/struct.Triangle.html) which could replace the 3 lines drawn below FWIW.

boards/feather_m0/examples/ssd1306_graphicsmode_128x64_i2c.rs
```diff
@@ -0,0 +1,158 @@
...
+    );
+
+    // default DisplaySize: 128x64 pixels; see the
+    // ssd1306::builder::Builder::new() method definition for more info
+    let mut disp: GraphicsMode<_> = Builder::new().connect_i2c(i2c).into();
+
+    disp.init().unwrap();
+    disp.flush().unwrap();
+
+    // triangle
```
jamwaffles: See other triangle comment

spicyjack:
> Looks good, although aside from the other comments I'd be wary of using an alpha release of both `ssd1306` and `embedded-graphics`, but that's up to the atsamd maintainers to decide whether to accept or not 🙂

The name of the `size` method to the _Builder_ struct changed in `ssd1306` between 0.2.6 and 0.3.x, just trying to future-proof the demo as much as possible.  The `embedded-graphics` dep version comes from the version of `ssd1306` that was used for these demos.

boards/feather_m0/examples/ssd1306_terminalmode_128x32_i2c.rs
spicyjack: The `ssd1306` driver currently only accepts `DisplaySize` enum values for display sizes, you can't just feed that driver an `x`/`y` value, so this would be a comment to clarify the code, and/or level-set expectations of what's possible with the `ssd1306` driver as it's currently written.

boards/feather_m0/examples/ssd1306_graphicsmode_128x32_i2c.rs
spicyjack: I'll take a poke at it and see what happens.

boards/feather_m0/examples/ssd1306_graphicsmode_128x64_i2c.rs
spicyjack: More poking.  Thanks for the feedback!

boards/feather_m0/examples/ssd1306_terminalmode_128x32_i2c.rs
jamwaffles: Sounds good 👍

sajattack:
I don't really care what version you use, and I commend your attempts to future-proof. Incorporate the triangle primitive and we should be good to go.

jamwaffles:
> I don't really care what version you use

I brought it up as a point for consideration - there are still many breaking changes coming in future alphas of embedded-graphics which might cause some maintenance pain, negating the future proofing. If you're happy with that, then 0.6.0-alpha.2 is fine :slightly_smiling_face:

spicyjack:
@jamwaffles How far away is a non-alpha release of `ssd1306` and `embedded_graphics`?  I think for these small examples, following breaking changes should not be too much of a burden... I'm willing to do the work as needed, and I'd like to also start submitting PRs for updating primitives in the [ssd1306 examples](https://github.com/jamwaffles/ssd1306/tree/master/examples).

jamwaffles:
> How far away is a non-alpha release of ssd1306 and embedded_graphics?

Hard to say, but not that soon as I can only work on it in my limited spare time.

>  I think for these small examples, following breaking changes should not be too much of a burden. 

Yeah good point, shouldn't be hard to keep up with e-g releases with this small amount of code 🙂 

> I'd like to also start submitting PRs for updating primitives in the ssd1306 examples.

Thanks!

PR #99, opened 2019-10-11
samd51-hal: Implement watchdog, serial number, reset cause

twitchyliquid64:


hal/src/samd51/mod.rs
```diff
@@ -4,8 +4,12 @@ pub mod pwm;
 pub mod sercom;
 pub mod timer;
 pub mod trng;
+pub mod watchdog;
 
-#[cfg(feature="unproven")]
+mod serial_number;
```
sajattack: Why are you doing this instead of just putting `pub mod serial_number`?

twitchyliquid64: I was under the impression that `pub mod serial_number` would mean consumers would need to do `hal::serial_number::serial_number()` instead of `hal::serial_number()`, which stutters?

sajattack: oh yeah I guess

PR #102, opened 2019-10-14
correct adc gain (fixes #101)

### LINKED ISSUE #101: SAMD21 ADC returning values 2x more than expected, except for fully saturated
I'm using a feather_m0 board and the [AD8495](https://www.adafruit.com/product/1778) thermocouple amplifier. The ADC HAL seems to be working great, except that it returns values exactly 2x of what is expected.

I've tried debugging this, but haven't gotten far. The gain settings are being written in the HAL code, and I don't see any other obvious ways we can get a 2x gain error. I'm reading from pin A5 for the record (pin B2 it seems).

Here's a trimmed down version of my code:

```rust
#![no_std]
#![no_main]

const ADC_FULLSCALE: u32 = 4095; // 12 bit ADC

extern crate cortex_m;
extern crate panic_halt; // panic handler

use feather_m0 as hal;
use hal::adc::Adc;
use hal::delay::Delay;
use hal::entry;
use hal::pac::{CorePeripherals, Peripherals};
use hal::prelude::*;

#[entry]
fn main() -> ! {
    // cortex_m::Peripherals::take().unwrap()
        peripherals.GCLK,
        &mut peripherals.PM,
        &mut peripherals.SYSCTRL,
        &mut peripherals.NVMCTRL,
    );

    let mut pins = hal::Pins::new(peripherals.PORT);

    let mut i2c = hal::i2c_master(
        &mut clocks,
        400.khz(),
        peripherals.SERCOM3,
        &mut peripherals.PM,
        pins.sda,
        pins.scl,
        &mut pins.port,
    );  // used for a display

    let mut adc = Adc::adc(peripherals.ADC, &mut peripherals.PM, &mut clocks);
    let mut a5 = pins.a5.into_function_b(&mut pins.port);

    loop {
        let data: u16 = adc.read(&mut a5).unwrap();
        let data: u16 = data / 2;
        let temp_c: f32 = (((data as f32 / ADC_FULLSCALE as f32) * 3.3) - 1.25) / 0.005;
        let temp: f32 = temp_c * (9. / 5.) + 32.;  // print to display
        delay.delay_ms(1000 as u32);
    }
}
```

sajattack:


twitchyliquid64:
Given this could break consumers who have no relied on the defaults, should we consider either exposing it behind a feature or a different API, or bumping the major version?

sajattack:
The defaults haven't been out long enough for many people to rely on them, and they were wrong to begin with.

sajattack:
Not to mention, a hal major version bump requires bumping every BSP, and this is such a small change to do that for.

twitchyliquid64:
Oky dokies

PR #103, opened 2019-10-16
Implement low-level Qspi

sajattack:
@twitchyliquid64 this should be enough to get started.

### DIFF
--- a/boards/metro_m4/examples/qspi.rs
+++ b/boards/metro_m4/examples/qspi.rs
@@ -1,0 +1,43 @@
+#![no_std]
+#![no_main]
+
+extern crate cortex_m;
+extern crate cortex_m_semihosting;
+extern crate metro_m4 as hal;
+#[cfg(not(feature = "use_semihosting"))]
+extern crate panic_halt;
+#[cfg(feature = "use_semihosting")]
+extern crate panic_semihosting;
+
+use hal::clock::GenericClockController;
+use hal::prelude::*;
+use hal::entry;
+use hal::pac::{CorePeripherals, Peripherals};
+use hal::qspi::Qspi;
+
+#[entry]
+fn main() -> ! {
+    let mut peripherals = Peripherals::take().unwrap();
+    let core = CorePeripherals::take().unwrap();
+    let clocks = GenericClockController::with_external_32kosc(
+        peripherals.GCLK,
+        &mut peripherals.MCLK,
+        &mut peripherals.OSC32KCTRL,
+        &mut peripherals.OSCCTRL,
+        &mut peripherals.NVMCTRL,
+    );
+    let mut pins = hal::Pins::new(peripherals.PORT);
+
+    let qspi = Qspi::new(
+        &mut peripherals.MCLK, 
+        &mut pins.port,
+        peripherals.QSPI, 
+        pins.flash_sck,
+        pins.flash_cs,
+        pins.flash_io0,
+        pins.flash_io1,
+        pins.flash_io2,
+        pins.flash_io3,
+    );
+    loop{}
+}
--- a/boards/metro_m4/examples/qspi.rs
+++ b/boards/metro_m4/examples/qspi.rs
@@ -14,19 +14,23 @@ use hal::prelude::*;
 use hal::entry;
 use hal::pac::{CorePeripherals, Peripherals};
 use hal::qspi::Qspi;
+use hal::qspi::Command;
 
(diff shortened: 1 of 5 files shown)


boards/metro_m4/examples/qspi.rs
```diff
@@ -0,0 +1,43 @@
...
+        &mut pins.port,
+        peripherals.QSPI, 
+        pins.flash_sck,
+        pins.flash_cs,
+        pins.flash_io0,
+        pins.flash_io1,
+        pins.flash_io2,
+        pins.flash_io3,
+    );
+    loop{}
```
twitchyliquid64: Nit: Could you make the example do something (maybe flash an LED with the results from reading the first byte) to demonstrate more than initialization?

hal/src/samd51/qspi.rs
```diff
@@ -0,0 +1,219 @@
...
+    ) -> Qspi {
+        let sck = sck.into_function_h(port);
+        let cs = cs.into_function_h(port);
+        let io0 = io0.into_function_h(port); 
+        let io1 = io1.into_function_h(port); 
+        let io2 = io2.into_function_h(port); 
+        let io3 = io3.into_function_h(port); 
+
+        mclk.apbcmask.modify(|_, w| w.qspi_().set_bit());
+        mclk.ahbmask.modify(|_, w| w.qspi_().set_bit());
```
twitchyliquid64: Could you roll the two`ahbmask` writes into one call to `modify()`? Also maybe add a comment, ie: _Enable the clocks for the qspi peripheral in single data rate mode._

hal/src/samd51/qspi.rs
```diff
@@ -0,0 +1,219 @@
...
+        self.qspi.ctrla.write(|w| {
+            w.enable().set_bit();
+            w.lastxfer().set_bit()
+        });
+
+        while self.qspi.intflag.read().instrend().bit_is_clear() {}
+        self.qspi.intflag.modify(|_, w| w.instrend().set_bit());
+    }
+
+    pub fn run_command(&self, command: Command) {
```
twitchyliquid64: Would it make sense for these to return a result?

Could we add a docstring for these public methods?

sajattack:
There are some safety issues in this as it stands, the functions should probably check that the commands sent to it can be performed in the given modes.

boards/metro_m4/examples/qspi.rs
sajattack: I made the example erase the chip, write something, read it back, and make the led solid if it matches, blinky if it doesn't. It blinked :'(

jacobrosenthal:
I tried to build it as is, and I get a red flashing led so indeed were not reading what were writing.

I have a salea here, and it seems theres a qspi analyzer someone wrote
https://github.com/YADRO-KNS/saleae_qspi/tree/extra_commands

Ill let you know what I see, but feel free to give any pointers or ideas what you were going to do next

Further i see this no_std fat12 library in crates, not sure if that was around last time you looked
https://github.com/rafalh/rust-fatfs

jacobrosenthal:
One interesting fact, the chip does response with correct ids.. but during debug only not during release.

Im working on pygamer, samd51 if it matters, GD25Q64C
```
    let mut read_buf = [0u8; 3];
    // #define GD25Q64C                                                               \
    // .manufacturer_id = 0xc8,.memory_type = 0x40, .capacity = 0x17,                                     \
    // $1 = [23, 64, 200]
    let dev_buf = [0x17, 0x40, 0xc8];
    qspi.read_command(Command::ReadId, &mut read_buf).unwrap();
    if read_buf != dev_buf {
        loop {
            red_led.set_high().unwrap();
            delay.delay_ms(100u8);
            red_led.set_low().unwrap();
            delay.delay_ms(100u8);
        }
    }
```

debug doesnt make the existing read/ write work though so no gain there

sajattack:
Yeah it will need some heavy debugging I think.

jacobrosenthal:
Just wanted to say I am, occasionally, working on this. This upstream work is mostly working, I needed to add while waits on some status bits mainly. Though I have to say I  really dislike the command enum pattern. 

I have actually mostly gotten data back over the wire in quad mode, but its had like the first bit corrupted so still some weirdness. 

My branch is here. 
https://github.com/jacobrosenthal/atsamd/commits/qspi

sajattack:
I'm not particularly attached to the way it's written if you want to change it.

PR #104, opened 2019-10-19
Pygamer usb

jacobrosenthal:
Neopixel example wasn't working for me at all.. I reved the underlying deps and now it works for 1 led but not more. I'll pr that work over there as well.

Then I brought over usb from metro_m4. (likewise its stuck on 1 lets as well)
The only problem there is uart_debug.rs  is pretty coupled and needs a better solution than this.

### DIFF
--- a/boards/pygamer/Cargo.toml
+++ b/boards/pygamer/Cargo.toml
@@ -42,11 +42,21 @@ ws2812-spi = "0.2.0"
 [dev-dependencies.ws2812-timer-delay]
 version = "~0.2"
 
+[dependencies.usb-device]
+version = "~0.2"
+optional = true
+
+[dependencies.usbd-serial]
+version = "~0.1"
+optional = true
+
 [features]
 # ask the HAL to enable atsamd51j20a support
 default = ["rt", "atsamd-hal/samd51j19a", "atsamd-hal/samd51"]
 rt = ["cortex-m-rt", "atsamd-hal/samd51j19a-rt"]
 unproven = ["atsamd-hal/unproven"]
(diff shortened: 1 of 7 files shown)


twitchyliquid64:
Yeah uart_debug is :fire: :wastebasket: - couldn't work out a better way to do it :( feel free to revise or maybe even rip it out.

Maybe it makes more sense being in the uart code? That way there can be an invariant for each kind of uart rather than the one on the itsybitsy_m4 pins.

twitchyliquid64:
> now it works for 1 led but not more.

Neopixel or apa102 LED?

twitchyliquid64:
I didn't think neopixels were compatible with SPI? I thought they had to be bitbanged?

jacobrosenthal:
Neopixels. And yes theyre being bitbanged using the same library as the itsy bitsy

jacobrosenthal:
Rather the feather m4 and metro m4

jacobrosenthal:
RE uart_debug, the main issue is I think my stuff is on sercom5 instead of 3, so hardcoded pins need changing. I suppose we can store via traits somehow. What about static dyn like the singleton pattern docs https://docs.rust-embedded.org/embedonomicon/singleton.html

jacobrosenthal:
Ive pulled all the smart-leds neopixel code out of here and its already been merged, so now this is just the usb work. 

I pulled the uart_debug feature because I dont know how to fix it yet either.. 

Technically this could merge any time. It works as well as the other implementation

despite the existing bug https://github.com/atsamd-rs/atsamd/issues/105

PR #106, opened 2019-10-28
Adc helper methods

TDHolmes:
Implement some helper methods on the ADC HAL to change things like the gain, reference source, and number of samples. Tested on a feather M0 but no SAMD51 targets. Also, the gain functionality is omitted from the SAMD51 target since it's so different than SAMD21 and there is no existing use of the gain correction feature.

First commit is functional changes, the second is rust formatting changes on those two files.

### DIFF
--- a/hal/src/samd21/adc.rs
+++ b/hal/src/samd21/adc.rs
@@ -1,5 +1,5 @@
 use crate::hal::adc::{Channel, OneShot};
-use crate::target_device::{ADC, PM};
+use crate::target_device::{adc, ADC, PM};
 use crate::clock::GenericClockController;
 use crate::gpio::{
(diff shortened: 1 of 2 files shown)


sajattack:
The reason I was dividing is so that we were referenced to 3v3.

hal/src/samd21/adc.rs
```diff
@@ -26,18 +26,37 @@ impl Adc<ADC> {
...
+
+        newadc
+    }
+
+    pub fn samples(&mut self, samples: adc::avgctrl::SAMPLENUM_A) {
+        self.adc.avgctrl.modify(|_, w| {
+            w.samplenum().variant(samples);
+            // I don't see any reason to divide ourselves. peripheral will automatically
+            // shift as needed
+            unsafe { w.adjres().bits(0) }
```
TDHolmes: @sajattack are you talking about this comment? I was just saying I feel we don't need to expose an API to adjust `adjres` since that does additional divides if you oversample, but the peripheral automatically divides as needed if the resulting resolution is >16 bits. I left your gain adjustment as is in the default case

PR #110, opened 2019-11-18
pygamer neopixels

jacobrosenthal:
This goes on top of #107 

I cant find any configuration that works with slow feature on timer. It only lights 1 led.
Also I cant find a configuration that works with lto=false

* nop - I cant get a working config for nop yet if anyone has ideas. Otherwise Ill delete it.
* spintimer 4-11 works `let timer = SpinTimer::new(4);`
* Timer, supposed to be 3mhz but 3.1-3.8 works `timer.start(3_100_000u32.hz());`

Then when this is merged,  #104 can be separated to only usb based changes and be rebased on top of this

### DIFF
--- a/boards/metro_m4/examples/spi.rs
+++ b/boards/metro_m4/examples/spi.rs
@@ -4,17 +4,17 @@
 extern crate cortex_m;
 extern crate cortex_m_semihosting;
 extern crate metro_m4 as hal;
+extern crate nb;
 #[cfg(not(feature = "use_semihosting"))]
 extern crate panic_halt;
 #[cfg(feature = "use_semihosting")]
 extern crate panic_semihosting;
-extern crate nb;
 
 use hal::clock::GenericClockController;
 use hal::delay::Delay;
-use hal::prelude::*;
 use hal::entry;
 use hal::pac::{CorePeripherals, Peripherals};
+use hal::prelude::*;
 use hal::sercom::PadPin;
 use nb::block;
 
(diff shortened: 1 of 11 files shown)


sajattack:
Thanks for all your hard work

jacobrosenthal:
Thank you all for all this to build on top of. Any thoughts on the nop neopixel, that doesnt actually function reliably. I was thinking of deleting or replacing it with one of the others.
And nobody cares that this re enabled lto otherwise, right?

sajattack:
Yeah it's better to have something that works with lto. Nop is pretty flakey and busted. I've tried to mess with it a couple times. It's just an option if people are out of spi and timers.

tarcieri:
Is this an actually stable neopixel driver?! 🤩

jacobrosenthal:
Its still just all the smart-leds org's existing work, but I have working
implementations for pygamer of
* ws2812-timer-delay with asm_delay which is probably still going to be
flaky as a nop
* ws2812-timer-delay with an actual timer. Still dont understand why I need
to 'overclock' it.. Medium flaky in future?
* ws2812-spi-rs which seems rock solid, but will pointlessly eat up the
pygamers i2c pins sadly

tarcieri:
@jacobrosenthal I'm mostly curious because I had a lot of issues with `lto` and `debug`/`release` profiles in general with @sajattack's nopper driver before (double checking now it required , and even when I had had a working binary, occasionally it would boot up out of sync for indeterminate reasons.

I scoped it and sent my results to @sajattack, but we never tracked down the root cause.

For context I'm trying to drive... a lot of neopixels, and for this many the timing seems particularly sensitive:

blinky

So far this has only worked with `debug = true` and `lto = false`:

https://github.com/atsamd-rs/atsamd/blob/master/boards/trellis_m4/Cargo.toml#L62

Changing either of these values has, historically and typically, caused the neopixels to stop working completely. In some cases they were somewhat on but the timing was obviously glitched. Unfortunately everything I want to actually do with this device could *really* benefit from both LTO and release builds

tl;dr: I'm asking does this mean I can flip on `release` builds and `lto` and the timing will still be correct?

jacobrosenthal:
Yeah try one of the timer based drivers or if your pins work out (which it
sounds like they didn't in our last convo) or you can choose different
sacrificial Pins try the spi based one

jacobrosenthal:
@tarcieri Oh hey, I found out why timer delay drive was funky. thats rock solid for me too now with this patch https://github.com/smart-leds-rs/ws2812-timer-delay-rs/pull/1 That seems like your best bet

tarcieri:
Woohoo, fantastic! I will try it out again.

PR #113, opened 2019-11-28
samd51 adc: init defines generic clock so adc0 and adc1 can be used at the same time

jacobrosenthal:
samd51 adc was hardcoding a single clock and crashing when both adc0 and adc1 were initialized at same time.

Moves clock selection to the adc init function instead. But this isnt very typestatey or safe and just moves the problem upstream. Perhaps we need a clock tree function call to replace  configure_gclk_divider_and_source to hand out generic clocks so users or subsystems cant reuse.

### DIFF
--- a/boards/metro_m4/examples/adc.rs
+++ b/boards/metro_m4/examples/adc.rs
@@ -1,18 +1,19 @@
 #![no_std]
 #![no_main]
 
-extern crate panic_semihosting;
 extern crate cortex_m_rt;
-extern crate metro_m4 as hal;
 extern crate embedded_hal;
+extern crate metro_m4 as hal;
+extern crate panic_semihosting;
 
 use core::fmt::Write;
 
-use hal::prelude::*;
+use hal::adc::Adc;
 use hal::clock::GenericClockController;
 use hal::entry;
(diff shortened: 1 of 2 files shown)


jacobrosenthal:
Offtopic but in same code, isnt panic formatting still a little troublesome to be guaranteed to be stripped. Should we have .expects over unwrap?

sajattack:
I'm never sure when to use expect or unwrap or whatever. I've been told unwrap is a bit of a sin but maybe that's wrong.

sajattack:
It's a little hairy to just assume the user passes in a 48mhz clock, I'm not sure what's better though.

sajattack:
Maybe something like timerparams that outputs a particular samplerate the user requests.

jacobrosenthal:
I think we need to impl all 11 clocks just with the GENA enums. Then have
user take them from clock tree and parcel them out to something like adc to
own, which can then some facade over configure_gclk_divider_and_source

sajattack:
Sure, that works, but configurable samplerate would be good too.

jacobrosenthal:
Opened a separate issue for the bigger discussion, I think this can be merged (when it passes travis)

PR #117, opened 2019-12-12
Edgebadge

jacobrosenthal:
Adafruit and friends are handing a ton of these out at conferences right now it would be good to support them.

Very close to a pygamer, differences include:
* 2mb instead of 8mb qspi flash
* no sd card
* 4 more buttons instead of joystick
* built in buzzer speaker
* built in PDM microphone

Mainly I just implemented the buttons fix and stripped sdcard from pygamer.


sajattack:
Have you tested this on appropriate hardware?

jacobrosenthal:
Yeah have access to a few. Can put my name on upkeeping it.

sajattack:
Cool, thanks

PR #125, opened 2019-12-22
Update samd21 usb

keithduncan:
I have a [usb serial echo example](https://gist.github.com/keithduncan/016c6c86b8397e0be63dcfc4468efe2c) working for my Arduino NB1500 using these changes.

I don’t have any of the Cortex-M0(+) boards in our example set to verify otherwise I’d contribute that gist as an example too.

### DIFF
--- a/hal/src/lib.rs
+++ b/hal/src/lib.rs
@@ -64,8 +64,3 @@ pub use self::samd21::*;
 pub mod samd11;
 #[cfg(feature = "samd11c14a")]
 pub use self::samd11::*;
-
-#[cfg(all(feature = "usb", feature = "samd21"))]
-pub use self::samd21::usb;
-#[cfg(all(feature = "usb", feature = "samd51"))]
-pub use self::samd51::usb;
--- a/hal/src/samd21/usb/bus.rs
+++ b/hal/src/samd21/usb/bus.rs
@@ -712,12 +712,12 @@ impl Inner {
             w.uprsm().set_bit()
         });
 
-        // TODO this should be done when an EIC for VBUS is triggered
+        self.flush_eps();
+
         self.usb().ctrlb.modify(|_, w| w.detach().clear_bit());
     }
 
-    fn reset(&self) {
-        dbgprint!("reset");
+    fn flush_eps(&self) {
         for idx in 0..8 {
             let cfg = self.epcfg(idx);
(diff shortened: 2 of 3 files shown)


hal/src/samd21/usb/bus.rs
```diff
@@ -761,21 +711,24 @@ impl Inner {
             w.eorsm().set_bit();
             w.uprsm().set_bit()
         });
+
+        self.flush_eps();
```
keithduncan: This part is cargoculted from the samd51 usb module

hal/src/samd21/usb/bus.rs
```diff
@@ -218,15 +168,15 @@ impl BufferAllocator {
...
         }
 
         let end_addr = unsafe { start_addr.offset(size as isize) };
         if end_addr > buf_end {
-            return Err(UsbError::SizeOverflow);
+            return Err(UsbError::EndpointMemoryOverflow);
         }
 
-        self.next_buf = unsafe { end_addr.offset_from(self.buffers.as_ptr()) as u16 };
+        self.next_buf = unsafe { end_addr.sub(self.buffers.as_ptr() as usize) as u16 };
```
keithduncan: This is also copied from the samd51 example and avoids using an unstable feature

hal/src/samd21/usb/bus.rs
```diff
@@ -761,21 +711,24 @@ impl Inner {
             w.eorsm().set_bit();
             w.uprsm().set_bit()
         });
+
+        self.flush_eps();
+
+        self.usb().ctrlb.modify(|_, w| w.detach().clear_bit());
```
keithduncan: This part was preventing the device from showing up on my host's bus scan

hal/src/lib.rs
```diff
@@ -64,8 +64,3 @@ pub use self::samd21::*;
 pub mod samd11;
 #[cfg(feature = "samd11c14a")]
 pub use self::samd11::*;
-
-#[cfg(all(feature = "usb", feature = "samd21"))]
```
keithduncan: Looking for some :eyes: on this change, it appears to me that these are already re-exported by the family globs `samd21::*` etc

sajattack:
LGTM

PR #127, opened 2019-12-22
Add device family-specific feature gating

jessebraham:
I've added features for each device family from which the other features "inherit" from. With more devices being introduced there is the potential for some messy `#[cfg]` attributes.

This shouldn't cause any issues AFAICT, but if I'm missing something or causing problems by doing this feel free to just close.


sajattack:
Thanks!

PR #128, opened 2019-12-23
Update M0 USB examples

BenBergman:
This is still a work in progress.

Building on the great updates from #125, I'm trying to get the M0 USB examples updated. So far I've got the Metro M0 example compiling but I don't have a Metro M0 to test with. I do have a couple other atsamd21 boards around that I'm planning to port the example to.

Extra sets of eyes on here would be great as I did a lot of copy-paste programming to get it in line with the M4 examples.

### DIFF
--- a/boards/metro_m0/examples/usb_serial.rs
+++ b/boards/metro_m0/examples/usb_serial.rs
@@ -6,167 +6,23 @@ extern crate panic_rtt;
 
 use cortex_m_rt::{exception, ExceptionFrame};
 use hal::clock::GenericClockController;
-use hal::hal::usb::usb_device::bus::UsbBusWrapper;
-use hal::hal::usb::usb_device::prelude::*;
 use hal::prelude::*;
-use hal::{dbgprint, entry};
-use hal::{interrupt, CorePeripherals, Peripherals};
+use hal::{entry};
+use hal::pac::{interrupt, CorePeripherals, Peripherals};
 
-static mut USB_BUS: Option<UsbBusWrapper<hal::UsbBus>> = None;
-static mut USB_DEV: Option<UsbDevice<hal::UsbBus>> = None;
-static mut USB_SERIAL: Option<cdc_acm::SerialPort<hal::UsbBus>> = None;
+use hal::usb::UsbBus;
+use hal::usb_allocator;
+use usb_device::bus::UsbBusAllocator;
+use usb_device::prelude::*;
+use usbd_serial::{SerialPort, USB_CLASS_CDC};
 
-// Minimal CDC-ACM implementation
-mod cdc_acm {
-    use core::cmp::min;
-    use hal::hal::usb::usb_device::class_prelude::*;
-    use hal::hal::usb::usb_device::utils::AtomicMutex;
-    use hal::hal::usb::usb_device::Result;
-
-    pub const USB_CLASS_CDC: u8 = 0x02;
-    const USB_CLASS_DATA: u8 = 0x0a;
-    const CDC_SUBCLASS_ACM: u8 = 0x02;
-    const CDC_PROTOCOL_AT: u8 = 0x01;
-
-    const CS_INTERFACE: u8 = 0x24;
-    const CDC_TYPE_HEADER: u8 = 0x00;
-    const CDC_TYPE_CALL_MANAGEMENT: u8 = 0x01;
-    const CDC_TYPE_ACM: u8 = 0x02;
-    const CDC_TYPE_UNION: u8 = 0x06;
-
-    const REQ_SET_LINE_CODING: u8 = 0x20;
-    const REQ_SET_CONTROL_LINE_STATE: u8 = 0x22;
-
-    const BUFSIZ: usize = 16;
-
-    struct Buf {
-        buf: [u8; BUFSIZ],
-        len: usize,
-    }
-
-    pub struct SerialPort<'a, B: 'a + UsbBus + Sync> {
-        comm_if: InterfaceNumber,
-        comm_ep: EndpointIn<'a, B>,
-        data_if: InterfaceNumber,
-        read_ep: EndpointOut<'a, B>,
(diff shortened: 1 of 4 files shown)


boards/metro_m0/src/lib.rs
```diff
@@ -230,23 +232,24 @@ pub fn uart<F: Into<Hertz>>(
...
-        usb_clock,
-        pm,
-        dm.into_function(port),
-        dp.into_function(port),
-        usb,
-    ))
+    ) -> UsbBusAllocator<UsbBus> {
+    clocks.configure_gclk_divider_and_source(GEN_A::GCLK2, 1, SRC_A::DFLL48M, false);
+    let usb_gclk = clocks.get_gclk(GEN_A::GCLK2).unwrap();
+    let usb_clock = &clocks.usb(&usb_gclk).unwrap();
```
keithduncan: Do you think this should configure gclk2 or take the usb clock as an argument instead? An app on top of this board crate could want to use gclk2 for something else?

boards/metro_m0/examples/usb_serial.rs
```diff
@@ -181,75 +37,74 @@ fn main() -> ! {
...
-                .device_class(cdc_acm::USB_CLASS_CDC)
-                .build(&[serial]),
-        );
-    }
+        USB_SERIAL = Some(SerialPort::new(&bus_allocator));
+    };
 
     unsafe {
-        core.NVIC.set_priority(hal::Interrupt::USB, 0);
+        core.NVIC.set_priority(interrupt::USB, 0);
```
keithduncan: I’m a little concerned about the dual case import continuing to work, is it guaranteed somewhere 🤔

boards/metro_m0/examples/usb_serial.rs
```diff
@@ -181,75 +37,74 @@ fn main() -> ! {
...
-            let serial = unsafe { USB_SERIAL.as_ref().unwrap() };
-
-            match serial.read(&mut buf) {
-                Ok(count) if count > 0 => {
-                    // Echo back in upper case
-                    for c in buf[0..count].iter_mut() {
-                        if 0x61 <= *c && *c <= 0x7a {
-                            *c &= !0x20;
+    unsafe {
+        USB_BUS.as_mut().map(|usb_dev| {
```
keithduncan: What do you think about a `match` on USB_BUS and USB_SERIAL instead of two `map` calls to have less nesting?

boards/metro_m0/src/lib.rs
BenBergman: Yeah, taking the USB clock probably makes sense. This was one of the major copy-paste points where I wasn't really sure about the right approach.

boards/metro_m0/examples/usb_serial.rs
BenBergman: Yeah, I have no idea about this one. Another copy-paste 🤞

boards/metro_m0/examples/usb_serial.rs
BenBergman: Sure, let me see what I can do with this one.

BenBergman: I'm not sure if this is what you had in mind, but I changed the maps to a match in 8d3fedb.

sajattack:
I'll let @keithduncan handle this one because he's the expert here.

boards/metro_m0/src/lib.rs
keithduncan: :+1: lets switch to that

boards/metro_m0/examples/usb_serial.rs
keithduncan: :+1: lets switch to the capitalised `Interrupt` since that’s the type name in the PAC

boards/metro_m0/examples/usb_serial.rs
keithduncan: :+1: that’s what I had in mind 😄

boards/metro_m0/examples/usb_serial.rs
BenBergman: Sounds good. It looks like it is [aliased to the lower case for when rt is enabled](https://github.com/atsamd-rs/atsamd/blob/master/pac/atsamd21g18a/src/lib.rs#L169), but upper case does feel more correct here anyway.

boards/metro_m0/src/lib.rs
BenBergman: I'm not totally sure what to do for this. I poked through the pac and hal for a bit and didn't really see anything obvious to use here. I skimmed the [data sheet](http://ww1.microchip.com/downloads/en/Appnotes/Atmel-42261-SAM-D21-USB_Application-Note_AT06475.pdf) a bit and noticed on page 11 it recommends using gclk0 as the default for USB unless you specifically need something else, so I'll switch it to gclk0.

keithduncan: That sounds good, `GenericClockController` already initialises gclk0 to 48MHz and the example is using the external oscillator :+1:

BenBergman:
Since I don't have a Metro M0 to test with, I started porting it to the Circuit Playground Express. I don't yet have an SWD debugger so I'm a bit stuck at the moment. I set the on-board LED to blink a pattern at each step of the initialization and it seems to just be stopping in one of the last initialization steps.

I've already mentioned this in the gitter, but for anyone wanting to help diagnose this I've gist-ed my CPX code (based on top of this PR's branch) [here](https://gist.github.com/BenBergman/161432bde25f633d1cb4c61027f790a7).

BenBergman:
For additional reference, I just noticed @twitchyliquid64 is also working on M0 USB support [here](https://github.com/twitchyliquid64/atsamd/commit/73134f4a547f535e2971cd50e93131ef4c61907c).

twitchyliquid64:
I got it to compile, but I'm having issues getting the USB to work properly. I'll probably have another go tonight or tomorrow.

Issue #121, opened 2019-12-15
ATSAMD11 PWM pins?


wyager:
Hello all, 

I'm not sure if I'm missing something, but I'm having trouble setting up multiple PWM pins on an ATSAMD11C14. I need a total of 4 PWM pins. Would this require adding support for TC2, or maybe TCC0?

sajattack:
@jessebraham

jessebraham:
Just got back from out of town, I'll take a look at it in the next couple days here.

jessebraham:
I've added (and tested) `Pwm2` to my fork (36afab5bac753b26dd8af4ad7cd5cc1d520e0225) which makes use of `TC2`.

As far as using multiple pins, the datasheet seems to imply it's possible:
> Timer/Counter -> 24.8 Signal Description
>
> Refer to “I/O Multiplexing and Considerations” on page 13 for details on the pin mapping for this peripheral. **One signal can be mapped on several pins.**

I believe it is possible to use `TCC0` as well. However, unless I'm mistaken, in both cases our `pwm!` macro needs to be modified as it currently expects a single `TCxPinout` as well as a `TCx` parameter.

@sajattack , am I on track here? The `pinout` member in the `$TYPE` struct does seem kind of like magic and doesn't appear to be read anywhere yet is still working as expected, so I'm clearly missing something there.

wyager:
> One signal can be mapped on several pins.

The thing I'm aiming for is to have 4 separate PWM counters. From a not-too-careful reading of the datasheet, it seems like this might be possible with TC{1,2}/WO{0,1}. I might be misunderstanding the way the WO channels work - the datasheet says `Waveform output channels for TC : 2`, whereas it says `Waveform output channels per TCC : 8`. I'm not sure if this means 2 WOs per TC or 2WOs total shared across both TCs. If it's the latter, I think I would have to use TCC0.

sajattack:
The mode we use the WO in uses 2 channels for one pwm output.

wyager:
One thing that's confusing me about the rust API is that the definition of `atsamd11c14a::tcc0` seems to be missing some stuff compared to what's mentioned in the datasheet/C snippets. E.g. it seems like `cc` should have some additional structure to account for the fact that there are 4x compare channels. All of this stuff is autogenerated right?

sajattack:
Could've been missing from the SVD

wyager:
[...]

PR #129, opened 2019-12-24
PWM HAL updates for samd11 & samd21 devices

jessebraham:
This is a continuation of #121. Thanks to some help from @wyager I was able to get PWM working using `TCC0`. I tested multiple channels and things seem to be working as expected.

I've also removed the `pinout` parameter of the `pwm!` macro, as it wasn't actually being used by anything. If it was being provided for some other reason just let me know and I can change it back, removing it just simplifies things.


wyager:
I think you can use e.g. `while tcc.syncbusy.read().swrst().bit()  {}` instead of manually &-ing.

jessebraham:
@wyager You are correct, thanks!

sajattack:
Could you add TCC for other families as well?

jessebraham:
Sure thing

jessebraham:
Alright I think that should mostly do it. May be some minor errors and/or a few missing `Pwm` definitions, but I think I got most of it.

I re-tested the `samd11` and tested the `samd21` using a Trinket M0 (the included exampled). I don't own any `samd51` devices so I can't test them.

jacobrosenthal:
Hey nice job! Im happy to test samd51 boards.

I started down the tcc path in order to get 24bit timers a while ago but never finished. 

I copied the metro_m4 pwm example to do a pwm on d13, it builds but didnt seem to work on board. But I didnt spend much time on it yet. 

Then I tried  timer swap for tcc0 on pygamer neopixel_rainbow_timer and noticed we dont have tcc for timers on samd51?  Maybe I can help

jessebraham:
That would be fantastic if you'd be able to do some testing, thanks. I'm not certain that the implementation for the `samd51` is accurate at this point, but without a device to test with I can't really determine this.

I have tested `TC{1,2}` and `TCC0` on the `samd11`, and have tested a variety of `TCx` and `TCCx` on the `samd21` and experienced no issues. I plan on doing some more thorough testing on the `samd21` as well.

jessebraham:
Sorry I forgot to mention that naming for the `samd51` PWMs are a bit janky currently. None of the `samd11` or `samd21` devices had naming collisions. For the time being, on the `samd51` the PWMs for `TCC{0,1,2}` are named `Pwm{00,01,02}` respectively. Not sure how we want to go about naming these properly.

jacobrosenthal:
The example Im working on for samd51 is in this branch here

https://github.com/jacobrosenthal/atsamd/commit/624bb7f150c1950dae2da069198bbbde7bdc19b7

Going off this datasheet here
http://ww1.microchip.com/downloads/en/DeviceDoc/60001507E.pdf

Im just starting to play with the branch but I dont think its working as is.

It might be out of the scope of this issue, as it seems more fundamental to the pwm trait:
Perhaps that should be one of the first traits to be pulled out of the embedded hal and be worked on a bit more? If you look at how many notes I had to get together just to start finding the right clocks and functions and pins. Perhaps we should consume the gpio pin we intend to output on to both make sure theyre not doing anything with it and also to help indicate if you're actually setting up the pin you think you are. The first time through even I wasn't and I vaguely know what I'm doing.

jessebraham:
Thanks for taking a look, this PR kind up blew up in scope (as they tend to).

I have tested and confirmed `TCx` and `TCCx` on `samd11` devices as previously stated, and additionally have done some minimal testing for `samd21` (need more dev boards!), which was successful but far from comprehensive. At this time I'm reasonably confident that the implementation is mostly correct for these device families.

The aforementioned devices are `thumbv6m`, whereas the `samd51` are `thumbv7em`, so that may or may not have something to do with the issues. Support for `same54` devices has been merged since I opened this PR as well, so that will likely suffer from the same issues.

I agree the current implementation could probably provide a better interface, and that the pins assignment is less than ideal. Not sure how best to address that at this point.

At this point I'm thinking we can either:
1. Revert the changes for `samd51` for the time being and merge what has been confirmed working for `samd11` and `samd21` devices. `samd51` and `same54` can be dealt with in another PR.
2. Rebase on `master`, and have another contributor help work on the `samd51` and `same54` bits (I don't have any boards using these otherwise I'd be happy to slog through it).

sajattack:
Feel free to revert samd51 support.

jessebraham:
This PR now applies *only* to the `samd11` and `samd21` device families. The `samd51` and `same54` will need to be handled separately.

wyager:
> Perhaps we should consume the gpio pin we intend to output on

This is how I did it in my FFI implementation, but I'm actually torn on whether this is a good idea. The problem is that there's not a straightforward mapping between e.g. duty cycle registers and pins. You can have 0,1, or 2 (or more?) pins mapped to the same PWM gizmo, and different PWM gizmos control a given pin depending on the pin's mode. You also probably want the ability to release the pin. Coming up with a good API for this might be tricky.

wyager:
I'm getting

```
error[E0432]: unresolved import `crate::hal::Pwm`
 --> /Users/will/.cargo/git/checkouts/atsamd-d272bcb1b045fea7/5b8be2d/hal/src/samd11/pwm.rs:2:18
  |
2 | use crate::hal::{Pwm, PwmPin};
  |                  ^^^ no `Pwm` in the root
```

when I try to update to using this commit. Any ideas?

jessebraham:
Building using `--features unproven` should fix that

wyager:
That seems to work, thanks. Is there any way to put that in the cargo config? I already have 

```
[features]
```
But I guess that doesn't do this.

jessebraham:
I think you can specify default features [based on the Cargo docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section) but I've never actually done this myself.

PR #123, opened 2019-12-21
pygamer: remove wheel function for smart_leds hsv conversion

jacobrosenthal:
Closes #111 

Just realized we have hsv conversion in the smart_leds library were already bringing in. Anyone against removing the copy pasted wheel function everywhere? Starting in pygamer for this PR.

PR #130, opened 2019-12-26
update smart-leds, remove wheel function for smart_leds hsv conversion (non pygamer boards only)

jacobrosenthal:
More  #123

Could probably use a spot check on these boards as I dont own them.

### DIFF
--- a/boards/feather_m4/Cargo.toml
+++ b/boards/feather_m4/Cargo.toml
@@ -26,15 +26,7 @@ default-features = false
 [dev-dependencies]
 panic-halt = "~0.2"
 panic-semihosting = "~0.5"
-panic_rtt = "~0.2"
 cortex-m-semihosting = "~0.3"
-cortex-m-rtfm = "~0.4"
-smart-leds = "~0.1"
-smart-leds-trait = "~0.1"
-
-[dev-dependencies.ws2812-timer-delay]
-version = "~0.1"
-features = ["slow"]
 
 [features]
 # ask the HAL to enable atsamd51j19a support
@@ -46,9 +38,9 @@ unproven = ["atsamd-hal/unproven"]
 incremental = false
 codegen-units = 1
 debug = true
-lto = false
+lto = true
 
 [profile.release]
 debug = true
-lto = false
+lto = true
 opt-level = "s"
--- a/boards/itsybitsy_m4/Cargo.toml
+++ b/boards/itsybitsy_m4/Cargo.toml
@@ -18,8 +18,7 @@ cortex-m = "~0.6"
 embedded-hal = "~0.2.3"
 bitbang-hal = "~0.3"
 apa102-spi = "~0.3"
-smart-leds = "~0.2"
-smart-leds-trait = "~0.2"
+smart-leds = "~0.3"
 nb = "~0.1"
 
 [dependencies.cortex-m-rt]
@@ -58,11 +57,11 @@ use_uart_debug = ["atsamd-hal/use_uart_debug"]
 incremental = false
 codegen-units = 1
 debug = true
-lto = false
+lto = true
 
 [profile.release]
 debug = true
-lto = false
+lto = true
 opt-level = "s"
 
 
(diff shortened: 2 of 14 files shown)


sajattack:
I'll try my Metro M4 when I'm back from Christmas holidays, we might want to switch to a different ws2812 impl than the nop one.

tarcieri:
I can try the Trellis M4 when I'm back from the Christmas holidays as well

sajattack:
SpinTimer is similar to nop but works better.

jacobrosenthal:
Yeah Ive consolidated on the spintimer for all my pygamer examples

jacobrosenthal:
Ping on this in case anyone is back from holidays.

sajattack:
Pygamer build failed.

jacobrosenthal:
Thanks for the ping. Fixed now

sajattack:
Neopixels not working on metro m4.

jacobrosenthal:
and to confirm, upstream master did work.

jacobrosenthal:
Oh pygamer is on newest version of of smartleds using ws2812 timer (without
slow) and with lto re-enabled. I found this to be far more stable on
pygamer. Ill push an update for trellis and metro m4 since thats what this
touches.

sajattack:
Not sure if it was working lately, it's probably lto and stuff breaking nop, SpinTimer is preferable.

sajattack:
Now rainbow works and blink doesn't.

jacobrosenthal:
Im not sure, I can peel off those commits or maybe you can take a look. Im not partial to any of these changes, especially if it blocks this merge. Im just trying to get the offer the other boards the cleanups Ive been doing in pygamer.

jacobrosenthal:
Note, The sodaq_one, metro_m0 and feather_m0 are still lto=false for debug only

Is there a reason to update those as well? And or, should we be using lto=false for debug?

sajattack:
Well, the m0 boards can't use spintimer, so might be best to leave those.

sajattack:
I'll try to fix neopixel blink and if I can't we might as well delete it.

jacobrosenthal:
Yeah I was hoping to avoid those boards as I cant confirm a best path for them. However, when I looked, I didnt find an m0 board here that has neopixel examples anyway? I think people are just copy pasting the lto settings from other boards

jacobrosenthal:
Metro m4 neopixel blink has a TimerCounter going in it for no reason.. that could be removed too

sajattack:
Yes I see that and am deleting it.

sajattack:
@tarcieri you around to try your boards?

tarcieri:
I am around but my boards are thousands of miles away. I can check in a week.

jacobrosenthal:
@tarcieri ping check those boards?

tarcieri:
I fly home today and can check very soon!

jacobrosenthal:
Another friendly ping :)

sajattack:
@tarcieri ping

tarcieri:
sorry about the delay! will check tomorrow

sajattack:
@jacobrosenthal, it looks like the unnecessary timer on metro m4 snuck back in.

jacobrosenthal:
Got it, thanks

tarcieri:
Just tried this out on the Trellis M4 and I can't get it working in either debug or release mode.

The resulting `neopixel_rainbow.bin` seems significantly larger than it used to be (~200kB instead of 20kB).

If I go back a bit before this PR though, the binaries are over 1MB (and therefore unflashable).

I was able to build from the last known working release (`trellis_m4` v0.1.2) and that works. I suppose I can try to bisect...

sajattack:
I think there's something weird with your setup. I run `arm-none-eabi-size` on neopixel_rainbow and it reports 1832 bytes.

tarcieri:
Seems I was missing the `-O binary` flag on `arm-none-eabi-objcopy`

After adding that it appears to be working! 🎉

PR #131, opened 2019-12-29
ATSAME54P20A support

mvniekerk:
ATSAME54 and ATSAME54P20A support.
Datasheet:
http://ww1.microchip.com/downloads/en/DeviceDoc/60001507E.pdf


mvniekerk:
I tried to glean from existing code any order of boards and features (sometimes 51 is before 21, so I would have put 54 above 51, vice versa)

PR #138, opened 2020-01-02
Fix initial value of used_clock bits in GenericClockController

keithduncan:
I noticed this while trying to enable the `atsamd21g18a::gclk::clkctrl::ID_A::EVSYS_0` clock with a method generated from `clock_generator!`.

`EVSYS_0` has a value of `7` and the generated method was returning `None`. I checked where `used_clocks` was modified and tracked down the seventh bit being set on initialisation in `GenericClockController::new`.

I think this should be setting the first bit, `ID_A::DFLL48` instead of the seventh bit `SRC_A::DFLL48M`. I’ve added type guards to ensure the constants passed in are from the `ClockId` type aliased to `ID_A`.

---

I wasn’t sure what to do for the samd51 or same54 hal clock modules, I think they probably want to use `ClockId::FDPLL0` and set `0b10` instead of `SRC_A::DPLL0` and set `0b10000000` and could use some :eyes: on the TODO I left in the changes.

### DIFF
--- a/hal/src/samd11/clock.rs
+++ b/hal/src/samd11/clock.rs
@@ -171,7 +171,7 @@ impl GenericClockController {
                 Hertz(0),
                 Hertz(0),
             ],
-            used_clocks: 1u64 << u8::from(DFLL48M),
+            used_clocks: 1u64 << u8::from(ClockId::DFLL48),
         }
     }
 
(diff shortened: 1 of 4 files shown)


sajattack:
The datasheet is unclear on if FDPLL0 and DPLL0 are the same.

keithduncan:
I’ve taken another look at this and I’m confident that this is correct.

In `samd51` and `same54`, the initial value of `used_clocks` should reflect the modification to `gclk.pchctrl[ClockId::FDPLL0]` in `configure_and_enable_dpll0`.

Using `DPLL0` from `genctrl::SRC_A` has a value of `7` which collides with our `ClockId::SERCOM0_CORE` and getting that clock would return `None`.

sajattack:
Build is failing.

PR #143, opened 2020-01-11
Watchdog updates

jessebraham:
I went ahead an implemented all traits currently available for the Watchdog Timer from `embedded-hal`. I initially did this for `samd11`/`samd21` devices, but have since ported the changes to the `samd51`/`same54` as well. I have tested and confirmed functionality on a `samd21`.

The previous implementation had enumerated the timeout values in milliseconds, but this is not quite correct. The timeout values are actually in clock cycles according to the various datasheets, so I have updated the enumerations to reflect such. I'm not crazy about the current naming for the enum values, but it's not the worst also.

I also worked on improving HAL parity by copying the `reset_cause` and `serial_number` functions over to the `samd11`/`samd21`, as they were low-hanging fruit. These were tested and confirmed working as well.

Just as a note, this PR introduces breaking API changes.

As usual if you'd like any additional changes made just let me know. I think we could probably pull this all into the `common/` module as well, there are only a handful of lines that would require feature gating, so let me know if you'd like that done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
   - CRATE=boards/pyportal FEATURES="--features=unproven"
   - CRATE=boards/trellis_m4 FEATURES="--features=keypad-unproven"
   - CRATE=boards/pygamer FEATURES="--features=unproven,usb,ws2812-timer,math"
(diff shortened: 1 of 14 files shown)


sajattack:
Looks great thanks!

PR #144, opened 2020-01-13
Pin exact versions of the ssd1306 and embedded-graphics crates in feather_m0

keithduncan:
These are pre-release and have breaking changes in later versions. A fresh build pulls the latest alpha versions which doesn’t compile, this change should ensure the build works fresh i.e. when generating a new Cargo.lock file.


jacobrosenthal:
I have a 0.6 alpha branch that im maintaining when it's time to fix all examples.

PR #149, opened 2020-01-14
even more pwm fallout

jacobrosenthal:
Blah this gift keeps giving. 

Pygamer is still broken because ferris_img example doesnt have unproven feature.

But instead Im going to default unproven for all the pyboards ive been managing. Basically all the examples use it at this point, and im stripping the ws2812-timer feature in https://github.com/atsamd-rs/atsamd/pull/130 which cleans up a ton of complexity

### DIFF
--- a/boards/edgebadge/Cargo.toml
+++ b/boards/edgebadge/Cargo.toml
@@ -53,7 +53,7 @@ optional = true
 
 [features]
 # ask the HAL to enable atsamd51j20a support
-default = ["rt", "atsamd-hal/samd51j19a", "atsamd-hal/samd51"]
+default = ["rt", "atsamd-hal/samd51j19a", "atsamd-hal/samd51", "unproven"]
 rt = ["cortex-m-rt", "atsamd-hal/samd51j19a-rt"]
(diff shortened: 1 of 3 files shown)


sajattack:
Rather than this, you can assign a feature to an example. https://doc.rust-lang.org/cargo/reference/manifest.html#the-required-features-field-optional

jacobrosenthal:
If I understand you correctly, yeah Im going to stop using that. For
pyboards almost every example has unproven. The screen requires it for pwm,
as do joystick, light sensor, battery reading for adc.
So its I think non negotiable for this specific board. I understand if
other boards dont think so.

sajattack:
Well, if someone downstream wants to not use unproven features for whatever reason, they have to `--no-default-features` the way you have it. What's wrong with adding it to the examples?

jacobrosenthal:
Trying to clean up user experience in general. Also unproven is basically
meaningless at this point and on its way out.
https://github.com/rust-embedded/embedded-hal/issues/163

sajattack:
Ok good enough for me.

PR #150, opened 2020-01-14
Update Cargo.toml for pwm move to unproven

tekulvw:
I added a `required-features` attribute(?) to the pwm example section. This forces the compiler to at least display a message saying that you must pass `--features="unproven"` when attempting to build the pwm example. Without this, the error isn't super intuitive and requires some source diving for new users to figure out.

### DIFF
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -76,6 +76,8 @@ name = "blinky_basic"
 
 [[example]]
 name = "pwm"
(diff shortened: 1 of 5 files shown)


jacobrosenthal:
Thanks. 
Can you do the same for trinket m0, metro m4, samd11_bare and serpente

tekulvw:
Added, the serpente board didn't have any examples specified so I added blinky_basic and pwm.

PR #151, opened 2020-01-14
Remove rtfm dep from feather bsp

tekulvw:
The current version (0.3) won't build correctly on stable due to old requirements for some sub dependencies.

Should this be updated for all board crates?

### DIFF
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -39,10 +39,6 @@ optional = true
 version = "~0.1"
 optional = true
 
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
Feather doesnt even have any rtfm usage, so that should be removed entirely. Theres some other bsps that have unused deps too.

Note the pygamer and edgebadge and metro have an rtfm example on the last versions of rtfm and Id like to see them updated.

tekulvw:
Removed that dep, do you want me to look at any other boards and clean them up?

PR #153, opened 2020-01-19
samd51: Fix spurious hangs for OUT packets in UsbBus

### LINKED ISSUE #105: atsamd usb serial not receiving characters until close
I have a usb commit in #104 here for pygamer When you open the port with `screen /dev/blah 115200` and type nothing is printed on screen and Ive checked and the characters arent even received. I thought it was maybe interrupts, but I implemented the polling to upper case echo example and it has the same behavior. 

So maybe my implementation is broken, but wait I noticed as you close screen, the led example will trigger and In the case of the polling to upper case example when you reopen screen you'll see your previous chars printed to screen.

So the act of closing is triggering for some reason. I can't confirm on itsy but I presume its the same. So maybe its my setup for some reason. I checked with [stm32-usb](https://github.com/stm32-rs/stm32-usbd) and a [upper case polling bluepill example](https://github.com/stm32-rs/stm32-usbd-examples/tree/master/example-stm32f103c8
) and can confirm those characters are echoed immediately using screen so its not my setup.

I have a feeling using the stty as given in directions in Itsy, opens and closes and is thus triggering a flush somehow and so this hasnt been noticed until now.

Im not yet familiar with the usb device implementation but I think it would lie in there somewhere. Any thoughts @sajattack ?

twitchyliquid64:
Fixes #105 

I spent hours on this - I can't work out why interrupt line 80 isn't firing in the flood of interrupt URBs that the bug demonstrates.

The fix is to enable the interrupt on the start-of-frame, and use that to periodically poll the handler to handle pending packets.

 I was not able to find any open-source C implementation that did not use the start-of-frame interrupt.

I've also gotten rid of suspend/wakeup interrupt handling, as thats not necessary on samd51 and is kind of broken for self-powered devices anyway.

### DIFF
--- a/boards/itsybitsy_m4/examples/usb_serial.rs
+++ b/boards/itsybitsy_m4/examples/usb_serial.rs
@@ -89,10 +89,11 @@ fn main() -> ! {
         core.NVIC.set_priority(interrupt::USB_OTHER, 1);
         core.NVIC.set_priority(interrupt::USB_TRCPT0, 1);
         core.NVIC.set_priority(interrupt::USB_TRCPT1, 1);
-        core.NVIC.set_priority(interrupt::USB_TRCPT1, 1);
+        core.NVIC.set_priority(interrupt::USB_SOF_HSOF, 2);
         NVIC::unmask(interrupt::USB_OTHER);
         NVIC::unmask(interrupt::USB_TRCPT0);
         NVIC::unmask(interrupt::USB_TRCPT1);
+        NVIC::unmask(interrupt::USB_SOF_HSOF);
     }
 
     loop {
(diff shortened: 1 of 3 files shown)


jacobrosenthal:
usb-device says poll:

> Must be called at least once every 10 milliseconds while connected to the USB host to be USB compliant.

Maybe SOF is just the correct way to do this. Any idea what the sof generation freq is, how did you come to /4?

twitchyliquid64:
> Any idea what the sof generation freq is, how did you come to /4?

Theres 1k USB frames per second, so polling USB every 4 frames seems like a good balance between:

1. Polling periodically to make sure the peripheral doesn't get stuck
2. Letting the other interrupts handle the bulk of the work
3. Servicing usb-device's requirement to poll at least once every 10ms (we do it every 4ms)

twitchyliquid64:
Got this working without using SOF, I'll close this PR and open a new one.

PR #154, opened 2020-01-20
samd51 UsbBus: Fix interrupts not enabling & missing poll events around USB reset

### LINKED ISSUE #105: atsamd usb serial not receiving characters until close

twitchyliquid64:
Fixes #105.
Co-authored-by: Jacob Rosenthal <jacobrosenthal@gmail.com>

So it looks like the issue came down to our writes to `EPINTENSET[n]` not sticking. I found one really important sentence:

> This [EPINTENSET] register is cleared by USB reset or when EPEN[n] is zero.

-- Page 1164 / 2129 of the SAM D5x/E5x Family Data Sheet

There is no register or other reference to `EPEN`. I've taken it to mean a non-zero value for `EPCFG[n].EPTYPE[0/1]`, but to be sure we configure the endpoint interrupts last.

The remaining changes are the result of an iterative audit of registers, ordering, and minimizing unnecessary writes so the driver is as reliable as possible.

### DIFF
--- a/hal/src/samd51/usb/bus.rs
+++ b/hal/src/samd51/usb/bus.rs
@@ -312,9 +312,10 @@ impl<'a> Bank<'a, InBank> {
         self.epintflag(self.index()).read().trfail1().bit()
     }
 
-    /// Resets the state of the endpoint as expected for a USB
-    /// protocol reset.
-    fn reset(&mut self) {
+    /// Writes the state of the endpoint to registers and its corresponding
+    /// in-memory descriptor. Status & flag registers are not cleared, except
+    /// for endpoint 0.
+    fn flush_config(&mut self) {
         let idx = self.index();
         let config = self.config().clone();
         {
@@ -326,10 +327,18 @@ impl<'a> Bank<'a, InBank> {
             desc.set_byte_count(0);
         }
 
-        self.epstatusclr(idx)
-            .write(|w| w.stallrq1().set_bit().dtglin().set_bit().bk1rdy().set_bit());
-        self.epintenset(idx).write(|w| w.trcpt1().set_bit());
-        self.clear_transfer_complete();
+        if idx == 0 {
+            self.epstatusclr(idx)
+                .write(|w| w.stallrq1().set_bit().dtglin().set_bit().bk1rdy().set_bit());
+            self.clear_transfer_complete();
+        }
+    }
+
+    /// Enables endpoint-specific interrupts. This is separate from reset()
+    /// as the EPINTENSET[n] documentation is ambiguous on writes sticking
+    /// when EPEN[n] is zero. As such, these interrupts are configured last.
+    fn setup_ep_interrupts(&mut self) {
+        self.epintenset(self.index()).write(|w| w.trcpt1().set_bit());
     }
 
     /// Prepares to transfer a series of bytes by copying the data into the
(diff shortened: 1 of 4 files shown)


jacobrosenthal:
Man. Nice job. Confirm fix.
Not sure if you saw it but the zephyr implementation has something akin to this.  I tried a few implementations but never got it working
https://github.com/zephyrproject-rtos/zephyr/blob/master/drivers/usb/device/usb_dc_sam0.c#L98

jacobrosenthal:
Wanna inspect these logs before we merge
https://gist.github.com/jacobrosenthal/cf4e909558c4f3ef4c03fa378d64e39e

This is upstream vs this patch, but with the WRITE FAIL debug line added to the transfer complete that has commented out


Shows some nice removals, but also shows were tending to now respond to the 'wrong' interrupts and getting a bunch of write fails we never used to

hal/src/samd51/usb/bus.rs
```diff
@@ -326,10 +327,18 @@ impl<'a> Bank<'a, InBank> {
             desc.set_byte_count(0);
         }
 
-        self.epstatusclr(idx)
-            .write(|w| w.stallrq1().set_bit().dtglin().set_bit().bk1rdy().set_bit());
-        self.epintenset(idx).write(|w| w.trcpt1().set_bit());
-        self.clear_transfer_complete();
+        if idx == 0 {
```
jacobrosenthal: Is this scope necessary at all? From manual


        //The USB bus reset is initiated by a connected host and MANAGED BY HARDWARE.
        // During USB reset the following registers are cleared:
        // Device Endpoint Configuration (EPCFG) register - except for Endpoint 0
        // Device Frame Number (FNUM) register
        // Device Address (DADD) register
        // Device Endpoint Interrupt Enable Clear/Set (EPINTENCLR/SET) register
        // Device Endpoint Interrupt Flag (EPINTFLAG) register
        // Transmit Stall 0 bit in the Endpoint Status register (EPSTATUS.STALLRQ0)
        // Transmit Stall 1 bit in the Endpoint Status register (EPSTATUS.STALLRQ1)
        // Endpoint Interrupt Summary (EPINTSMRY) register

hal/src/samd51/usb/bus.rs
```diff
@@ -441,15 +451,23 @@ impl<'a> Bank<'a, OutBank> {
...
-        self.epstatusclr(idx).write(|w| {
-            w.stallrq0()
-                .set_bit()
-                .dtglout()
-                .set_bit()
-                .bk0rdy()
-                .set_bit()
-        });
-        self.epintenset(idx)
+        if idx == 0 {
```
jacobrosenthal: Same

hal/src/samd51/usb/bus.rs
twitchyliquid64: I think so, I put that conditional in so that we reset those bits on endpoint zero only, which is the endpoint which does not get cleared.

twitchyliquid64: Wait did I read that wrong? The EPSTATUS register is reset for all endpoints?

jacobrosenthal:
LGTM

sajattack:
Thanks guys.

PR #155, opened 2020-01-23
samd21 UsbBus: Fix implementation & get itsybitsy_m0 example working

twitchyliquid64:
The main issue was the clocking: Driving a 48Mhz closed loop from a 32khz oscillator is a 15k multiplier, and with usb correction enabled, it was really throwing it off. Every time the USB bus would reset, the UART would stop working for a little thats how bad the timing got.

Aside from that, I've brought the samd21 implementation in line with the samd51 one, caveat minor hardware differences (ie: clocking).

### DIFF
--- a/hal/src/samd21/usb/bus.rs
+++ b/hal/src/samd21/usb/bus.rs
@@ -19,6 +19,10 @@ use usb_device::bus::PollResult;
 use usb_device::endpoint::{EndpointAddress, EndpointType};
 use usb_device::{Result as UsbResult, UsbDirection, UsbError};
 
+use crate::dbgprint;
+#[cfg(feature = "use_uart_debug")]
+use crate::uart_debug;
+
 /// EndpointTypeBits represents valid values for the EPTYPE fields in
 /// the EPCFGn registers.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
--- a/hal/src/samd21/usb/bus.rs
+++ b/hal/src/samd21/usb/bus.rs
@@ -19,10 +19,6 @@ use usb_device::bus::PollResult;
 use usb_device::endpoint::{EndpointAddress, EndpointType};
 use usb_device::{Result as UsbResult, UsbDirection, UsbError};
 
-use crate::dbgprint;
-#[cfg(feature = "use_uart_debug")]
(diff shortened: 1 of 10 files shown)


hal/src/samd21/usb/bus.rs
```diff
@@ -15,6 +19,12 @@ use usb_device::bus::PollResult;
 use usb_device::endpoint::{EndpointAddress, EndpointType};
 use usb_device::{Result as UsbResult, UsbDirection, UsbError};
 
+use crate::dbgprint;
+#[cfg(feature = "use_uart_debug")]
+use crate::uart_debug;
```
sajattack: I thought you were going to remove this?

twitchyliquid64: Ah yes, nice catch.

lopsided98:
I was investigating USB clock correction and figured out that the problem is that DFLLMUL.MUL needs to be set to 48000 (48 MHz / 1kHz), rather than 1464 (48 MHz / 32768 Hz). When USBCRM is enabled, the 32 kHz reference clock is ignored. If USB is connected, the DFLL runs in closed-loop mode using the 1 kHz SOF signal, otherwise it runs in open loop mode. This is described in section 16.6.7.2 (Atmel datasheet) or 17.6.7.2.2 (Microchip datasheet) of the datasheet.

I tested various configurations on my itsybitsy_m0. The frequencies I measured depend on many external factors, but they may still give some insights. Enabling USBCRM with the current multiplier results in a frequency of around 52 Mhz, which explains why things didn't work. With the correct multiplier, I get a frequency of 47.98 MHz. In open loop mode (USB disconnected), I measured 46.50 Mhz. In closed loop mode with the 32 kHz oscillator, I measured 47.12 MHz.

Perhaps there should be an option to choose between open-loop/USBCRM or closed-loop with 32 kHz reference. I would expect that closed loop mode has better accuracy, even with the internal oscillator, so there may be some users who would prefer one over the other.

bradleyharden:
@lopsided98, could you open an issue? We have a full featured clock module coming in #450 (although only for the thumbv7em chips right now). I'd like to make sure we model this correctly there. Tagging @vccggorski and @vcchtjader

PR #161, opened 2020-01-25
Add ability to usb_echo to reset directly to bootloader

twitchyliquid64:
This is a nice feature to have & to be able to demo, I'm not 100% sure where it should go though?

### DIFF
--- a/boards/itsybitsy_m4/Cargo.toml
+++ b/boards/itsybitsy_m4/Cargo.toml
@@ -16,10 +16,14 @@ documentation = "https://atsamd-rs.github.io/atsamd/atsamd51g19a/itsybitsy_m4/"
 [dependencies]
 cortex-m = "~0.6"
 embedded-hal = "~0.2.3"
-bitbang-hal = "~0.3"
 apa102-spi = "~0.3"
 smart-leds = "~0.3"
 nb = "~0.1"
+
+[dependencies.bitbang-hal]
+version = "~0.3"
+path = "../../../bitbang-hal"
+
 
 [dependencies.cortex-m-rt]
 version = "~0.6.12"
--- a/boards/itsybitsy_m4/src/pins.rs
+++ b/boards/itsybitsy_m4/src/pins.rs
@@ -233,6 +233,38 @@ impl SPI {
             ),
         )
     }
+
+    /// Convenience for setting up the labelled pins to operate as an
+    /// SPI master in half-duplex (transmit only) mode, running at
+    /// the specified frequency.
+    pub fn spi_master_half_duplex<F: Into<Hertz>>(
+        self,
+        clocks: &mut GenericClockController,
+        bus_speed: F,
+        sercom1: SERCOM1,
+        mclk: &mut MCLK,
+        port: &mut Port,
+    ) -> SPIMaster1<
+        hal::sercom::Sercom1Pad3<gpio::Pb23<gpio::PfC>>,
+        hal::sercom::Sercom1Pad0<gpio::Pa0<gpio::PfD>>,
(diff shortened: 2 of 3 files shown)


boards/itsybitsy_m0/src/lib.rs
```diff
@@ -245,3 +245,16 @@ pub fn usb_allocator(
         usb,
     ))
 }
+
+
+// uf2 bootloaders look for this value to determine if they should stay in the bootloader.
+const ENTER_BOOTLOADER_MAGIC: u32 = 0xf01669ef;
+// The magic value for remaining in the bootloader is stored in the last 4 bytes
+// of RAM. The itsybitsy_m0 has 32k of ram & starts at 0x20000000.
+const ITSYBITSY_BOOTLOADER_MAGIC_ADDR: *mut u32 = (0x20000000 + (32 * 1024) - 4) as *mut _;
```
jacobrosenthal: we can add something next to _stack_start in memory.x to point to flash_start or something

jacobrosenthal:
its a bsp capability for adafruit boards, so probably in all (sigh) the bsps that have the bootloader. Though I dont think an example is necessary in all

twitchyliquid64:
What if I add `reset_into_bootloader` into the itsybitsy_m4 crate but only have the example in itsybitsy_m0? That way we have the two methods ready to use and if ppl search github for examples theres at least one

jacobrosenthal:
Sounds good to me. Im going to like wise thinking about combining and culling pygamer examples in the same way as Ive gained quite a few over time

jacobrosenthal:
I have a usb poll which is basically your usb echo. Ill probably bring it in there too, good place to tuck it in.

boards/itsybitsy_m0/src/lib.rs
jacobrosenthal: I dont have a working verison, but memory.x like this would make it less fragile and portable
```
__config__   = ORIGIN(RAM) + LENGTH(RAM) - 4;
SECTIONS
{
  .uf2Segment __config__ : {KEEP(*(.uf2Section))} > RAM
}
```

Then could make it a struct, but then have to init to to something?
maybe DBL_TAP_MAGIC_QUICK_BOOT is the default?
```
pub const DBL_TAP_MAGIC_QUICK_BOOT: u32 = 0xf02669ef;
#[repr(packed)]
pub struct Uf2Config {
    pub magic: u32,
}
#[link_section = ".uf2Section"]
static CONFIG: Uf2Config = Uf2Config {
    magic: DBL_TAP_MAGIC_QUICK_BOOT,
};
```
Or uninit?

```
#[link_section = ".uf2Section"]
static mut CONFIG: MaybeUninit<[u32; 1]> = MaybeUninit::uninit();
pub unsafe fn reset_into_bootloader() {
    CONFIG = MaybeUninit::new([DBL_TAP_MAGIC]);
    cortex_m::peripheral::SCB::sys_reset();
}
```

twitchyliquid64: I spent some time trying to get this to work using the linker script, which is made harder by not being allowed custom MEMORY sections. Let me know if you get it working, otherwise would it be okay to stick to mashing pointers at runtime (same way the generated svd code does it?)

jacobrosenthal: Not that Im in charge, but yeah I guess theres no other data held there right? So its not buying much.

twitchyliquid64: Not in charge either but I trust your intuition as an embedded person / responsible adult lol

jacobrosenthal: Heh. Ill take it. My adult says lets get a comment line there saying the 0x20000000 comes from and should be kept up to date with memory.x

PR #162, opened 2020-01-31
Feather M0 USB Example

tekulvw:
Copied the example from itsybitsy, got code hooked together, added new deps and have a working echo!


twitchyliquid64:
Nice! Could you update the `.travis.yaml` [here](https://github.com/atsamd-rs/atsamd/blob/master/.travis.yml#L12) so it compiles the example + compiles with the `usb` feature? That way travis can will fail if the example breaks compilation.

sajattack:
@tekulvw could you rebase this on master please?

WyseNynja:
What can I run on the host machine to test this?

twitchyliquid64:
The device emulates a serial port: you should be able to open it and have it echo back characters

PR #167, opened 2020-02-02
cortex-m-rt fix ARM.extab

jacobrosenthal:
https://github.com/rust-embedded/cortex-m-rt/issues/157


sajattack:
What does it do in plain english?

jacobrosenthal:
An upstream rust change broke cortex-m-rt  a while back. Anyone fresh building will eval up to latest and not notice anything, but people who have a lock file might be stuck unless we push the minimum up above 0.6.10 I think? but I just set it to 0.6.12.

sajattack:
Github says it needs a rebase.

PR #163, opened 2020-01-31
Upgrade embedded-graphics to 0.5, st7735-lcd to 0.6

johnbatty:
Move pygamer up to a newer version of embedded-graphics.
This also requires upgrading the st7735-lcd driver.

PR #173, opened 2020-03-28
Embedded graphics 0.6

jacobrosenthal:
Supersedes #163

Itsy and Feather need testing.. Make sure I didnt fill a primitive that was supposed to be only stroked or something too.


twitchyliquid64:
> Feather looks good to me. Anyone else have itsy? @twitchyliquid64 ?

I have em Itsy's but no display, so not sure that testing them would be much better than checking that it compiles.

sajattack:
Right, ok. I guess I'll yolo it then.

PR #181, opened 2020-04-01
Create serial.rs

danielpi:
Adding an example for configuring a uart serial connection of the feather M4 board


sajattack:
Thanks!

PR #185, opened 2020-04-12
True Random Number Generator example for the feather_m4

danielpi:
Adding an example of using the True Random Number Generator for the feather_m4 board


danielpi:
Sorry, still trying to figure out github.

PR #193, opened 2020-04-22
boards: Add Seeeduino XIAO support

gkelly:
Add Seeed Studio Seeedunio XIAO support. The `blink` and `usb_serial`
examples were tested on hardware.

---
The board is commonly called Seeedunio XIAO, but there's references to it being called XIAO M0 in their Arduino support packages so I went with it here in case there's a different variant in the works.


sajattack:
Looks great to me, thanks for your contribution.

Issue #195, opened 2020-05-11
Issues blocking `update.sh` on master


gkelly:
I recently started adding support for ATSAMD51P19A. While I was doing this I ran `update.sh` and noticed that at least one other ATSAMD51 doesn't actually build at head. I started digging in to this and discovered a few issues.

- [x] `svd2rust` 0.17.0 doesn't support dimable names (`CHANNEL[%s]`, for example). This is because in December of 2019 the `svd-parser` package was patched to enforce naming rules, but it uses the pre-CMSIS 1.3.3 naming rules.
  - Fixed in https://github.com/rust-embedded/svd/pull/119.
- [x] `svd2rust` is missing adding some `use` declarations in some modules.
  - Fixed in https://github.com/rust-embedded/svd2rust/pull/444.
- [x] Some register reset values generated by `svd2rust` are bigger than their containing types. This doesn't get caught until after `svd patch`, `svd2rust`, and `form` are run. This doesn't _block_ getting `update.sh` running again, but it's a quality-of-life improvement that would help catch breakages early.
  - Fixed in https://github.com/rust-embedded/svd/pull/120.
- [x] Update `update.sh` to install the release of `svd2rust` that includes the above changes.
  - Proposed https://github.com/atsamd-rs/atsamd/pull/197 to address.
- [ ] Add SVD patch yamls for devices that are failing to generate.
  - [X] `ATSAMD5X`-series devices.
    - Fixed in https://github.com/atsamd-rs/atsamd/pull/196.
  - [ ] `ATSAMD11C14A`. `USB.DEVICE` cluster has wrong register size.
- [ ] Run `update.sh` and merge results as a PR.

jessebraham:
Thanks for opening this issue. I am currently working on at least some related problems, just haven't had a ton of time for this project lately unfortunately. Probably should have opened an issue myself, so thanks for taking care of that.

I have a few patches to be applied to the SVDs locally, however I need to patch `svdtools` one more time to allow modifying nested clusters; once there is support for that, the SVDs can be fully patched (as far as I know) and that will help at least some issues.

Regarding the clusters, AFAIK you are correct in that `svd2rust` has some issues processing them currently, so we're sort of blocked on that too.

I will try to list out what has been done and what needs to be done (other than what you've posted) later tonight or this week when I get a bit more time.

gkelly:
[...]

PR #197, opened 2020-05-14
Use bleeding-edge `svd2rust`

gkelly:
A pair of changes (rust-embedded/svd#119, rust-embedded/svd2rust#444)
recently went into the `svd2rust` and `svd-parser` crates. Until these
changes are part of a release have `update.sh` pull a fork of `svd2rust`
that uses the master version of `svd-parser`.

See #195.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -5,7 +5,7 @@ set -xe
 
 # INSTALL DEPENDENCIES
 
-cargo install --force --version 0.17.0 svd2rust
+cargo install --force --git https://github.com/gkelly/svd2rust --branch bleeding-edge svd2rust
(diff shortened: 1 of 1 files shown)


sajattack:
Could you maybe pin the commit hash so we don't get other unexpected bleeding edge changes?

gkelly:
Just pushed a commit that pins `svd2rust` to https://github.com/gkelly/svd2rust/commit/2bbb60590096bcb67c91f38bedd1f63f98132abe, which itself includes an explicit rev for `rust-embedded/svd`, so no change in either repository can get accidentally used.

PR #198, opened 2020-05-18
Add initial support for SAMD10C14A (SAMD10 family)

twopoint718:
- Add SVD files for samd10c14a and generate PAC

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ In addition to the PACs and HAL, there numerous **B**oard **S**upport **P**ackag
 
 | Crate | Version | Board Support Packages |
 |:------|:--------|:-----------------------|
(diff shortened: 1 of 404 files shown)


sajattack:
Sorry, the SVD parser is currently being worked on so this isn't quite correct. Also, if you choose to redo this PR after #195 is closed, please add necessary changes to the HAL at the same time. Not much point in a PAC without a HAL here.

twopoint718:
Ah okay, that's good to know. Yes, I will resubmit once #195 is closed. I'm still learning Rust and so the HAL is a little bit beyond me at the moment, but I'll see if I can find a collaborator. Thanks for the feedback!

sajattack:
We are now ready to accept new chips.

twopoint718:
🎉 I'll start working on it tonight. Thanks!

PR #199, opened 2020-05-19
Switch from svdtools to XSLT

gkelly:
Switch the SVD modifications from being made through svdtools' `svd
patch` to using XSLT. Since SVD files are already XML files, XSLT is
well-suited to modifying them.

This change also includes fixes for the `ATSAMD11C14A` device which are
not currently possible with `svdtools`.

See #195.


wez:
(This is the only time I've ever been happy to see xsltproc in use :-p)

gkelly:
> (This is the only time I've ever been happy to see xsltproc in use :-p)

I definitely had a bit of trepidation as I wrote this change. I know some people are outright allergic to xsltproc. 😛

sajattack:
I will admit I am somewhat hesitant to diverge too far from svdtools because it's what the rest of the ecosystem is using, but if it's the only way we can get this to work and consensus is fine with it, I won't stop this PR from going through.

gkelly:
> I will admit I am somewhat hesitant to diverge too far from svdtools because it's what the rest of the ecosystem is using, but if it's the only way we can get this to work and consensus is fine with it, I won't stop this PR from going through.

To be fair, it's not the only way to make this work. It's not even the only way to make this work without `svdtools`, you could just use `patch` to add corrections to the SVDs. However, making this work with `svdtools` is a losing game, in my opinion. Each new SVD I see is a new way that a vendor has managed to generate a weird broken SVD. `svdtools` (and `svd-parser`) are both designed to handle more-or-less well-formed SVDs. Using XSLT has the advantage that the only thing an SVD author needed to get right was that the SVD has to parse as XML. Everything else we can fix with XSLT. Other changes will inevitably require changes to `svdtools` to be able to reach into specific nodes, or even bigger changes if it needs to modify things like registers inside clusters inside devices (like `ATSAMD11C14A`).

The cons are, as you say, that there has been ecosystem adoption for `svdtools`. However, I do think the flexibility of having a tool _now_ that can do every SVD modification we could need is a very strong pro.

sajattack:
Makes sense. I'd like to get a bit more feedback from community members before merging. Particularly @jessebraham

PR #202, opened 2020-06-05
32k clocks are 32768 Hz, not 32000 Hz

TDHolmes:
Looking at the data sheets, it seems like all of the 32k clocks are 32768, not 32000 as in the code. Is there any reason it's not 32768?

I also applied code formatting in the `same54` file since my IDE did it. I can revert the commit if desired.

image


jacobrosenthal:
Interesting. I had problems with timing on neopixel cycle delays way back
when.  Wonder if this was it

PR #205, opened 2020-06-13
Delay implementation that uses a timer to WFI sleep

TDHolmes:
Using any timer as an interrupt source, delay for the given amount of time but WFI for improved power performance. On my feather_m0, I went from ~11mA to ~5mA consumption. I'd expect even less if you had a more efficient regulator or were more careful about peripherals being powered.

Currently the added example doesn't compile due to [this issue](https://github.com/rust-embedded/cortex-m/issues/221). TLDR, the `unmask` NVIC method isn't available when compiling the example. Everything works fine in [my project](https://github.com/TDHolmes/oven-temp-rs/blob/master/src/main.rs#L96) though, as well as [this minimal reproduction](https://github.com/TDHolmes/nvic-error).

- [x] ~~Fix example so it compiles - help on this would be appreciated~~ seems like this was fixed after updating my nightly compiler
- [x] Fix examples that fail to compile due to `InterruptDrivenTimer` trait not being in scope
- [x] ~~Potentially hide this behind some sort of feature? (`unproven`?)~~ probably not needed?
- [x] Allow for >1s delays

### DIFF
--- a/hal/src/common/sleeping_delay.rs
+++ b/hal/src/common/sleeping_delay.rs
@@ -6,6 +6,8 @@ use crate::time::U32Ext;
 use crate::timer_traits::InterruptDrivenTimer;
 use hal::blocking::delay::{DelayMs, DelayUs};
 
+const NUM_US_IN_S: u32 = 1_000_000;
+
 /// Delay and sleep while we do (WFI) using a timer
 pub struct SleepingDelay<TIM> {
     timer: TIM,
@@ -39,13 +41,28 @@ where
     TYPE: Into<u32>,
 {
     fn delay_us(&mut self, us: TYPE) {
-        self.timer.start((1_000_000_u32 / us.into()).hz());
+        let mut us: u32 = us.into();
+        if us == 0 {
+            panic!("Invalid delay duration");
+        }
+
+        // Determine how many cycles we need to run for this delay, if any
+        let mut count: u32 = 1 + us / NUM_US_IN_S;
+        if count > 1 {
+            us /= count - 1;
+        }
+
+        // Start the timer and sleep!
+        self.timer.start((NUM_US_IN_S / us).hz());
         self.timer.enable_interrupt();
         loop {
             asm::wfi();
             if self.timer.wait().is_ok() || self.interrupt_fired.load(atomic::Ordering::Relaxed) {
                 self.interrupt_fired.store(false, atomic::Ordering::Relaxed);
-                break;
+                count -= 1;
+                if count == 0 {
+                    break;
+                }
             }
         }
         self.timer.disable_interrupt();
--- a/hal/src/common/sleeping_delay.rs
+++ b/hal/src/common/sleeping_delay.rs
@@ -33,12 +33,13 @@ where
     }
 }
 
-impl<TIM> DelayUs<u32> for SleepingDelay<TIM>
+impl<TIM, TYPE> DelayUs<TYPE> for SleepingDelay<TIM>
 where
     TIM: InterruptDrivenTimer,
+    TYPE: Into<u32>,
 {
-    fn delay_us(&mut self, us: u32) {
-        self.timer.start((1_000_000 / us).hz());
+    fn delay_us(&mut self, us: TYPE) {
+        self.timer.start((1_000_000_u32 / us.into()).hz());
         self.timer.enable_interrupt();
         loop {
             asm::wfi();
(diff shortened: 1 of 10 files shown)


TDHolmes:
To fix the compilation errors like this:
```rust
error[E0599]: no method named `enable_interrupt` found for struct `atsamd_hal::samd21::timer::TimerCounter<atsamd21g18a::TC3>` in the current scope
  --> examples/blinky_rtfm.rs:70:13
   |
70 |         tc3.enable_interrupt();
   |             ^^^^^^^^^^^^^^^^ method not found in `atsamd_hal::samd21::timer::TimerCounter<atsamd21g18a::TC3>`
   |
   = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
   |
4  | use atsamd_hal::common::timer_traits::InterruptDrivenTimer;
   |
```

would it be acceptable to add
```rust
pub use crate::timer_traits::InterruptDrivenTimer as _atsamd_hal_timer_traits_InterruptDrivenTimer;
```

to the prelude? I don't want to break people

sajattack:
Yeah, adding to the prelude is fine.

hal/src/common/sleeping_delay.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+impl<TIM, TYPE> DelayUs<TYPE> for SleepingDelay<TIM>
+where
+    TIM: InterruptDrivenTimer,
+    TYPE: Into<u32>,
+{
+    fn delay_us(&mut self, us: TYPE) {
+        self.timer.start((1_000_000_u32 / us.into()).hz());
```
BroderickCarlin, line 42: Recommend using `checked_div()` here to avoid potential issues if `0` is passed in

hal/src/common/timer_traits.rs
```diff
@@ -0,0 +1,12 @@
+use crate::hal::timer::{CountDown, Periodic};
+use crate::time;
+
+/// Trait for timers that can enable & disable an interrupt the fires
```
BroderickCarlin, line 4: nit: `enable & disable an interrupt that fires`

hal/src/common/sleeping_delay.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+impl<TIM, TYPE> DelayMs<TYPE> for SleepingDelay<TIM>
+where
+    TIM: InterruptDrivenTimer,
+    TYPE: Into<u32>,
+{
+    fn delay_ms(&mut self, ms: TYPE) {
+        self.delay_us(ms.into() * 1_000_u32);
+    }
```
BroderickCarlin, line 62: Ref: https://github.com/atsamd-rs/atsamd/pull/205/files#r443156053

If `delay_ms` is just a wrapper around `delay_us` the 1sec limitation of `delay_us` should be addressed

TDHolmes:
> With the current implementation, the accuracy of the `delay_us()` falls of drastically the longer the delay actually is because of rounding errors in the division.

I don't think this can be avoided since the underlying timers I'm using use `Hertz` to keep track of time, so delays between ~ 1s and 100ms are gonna have pretty bad resolution. I could break it down into smaller chunks and wake up more frequently, but that kind of defeats the purpose of a sleeping delay too. Ideally the timer implementation would be period based instead of frequency based

sajattack:
Can you convert this PR to draft until it is ready?

hal/src/common/sleeping_delay.rs
TDHolmes: Not an issue with the new period-based timer approach

hal/src/common/sleeping_delay.rs
TDHolmes: multiple second delays are implemented

hal/src/common/timer_traits.rs
TDHolmes: fixed

TDHolmes:
Everything now works (example verified in (this repo)[https://github.com/TDHolmes/nvic-error])). I cannot figure out why the example fails to compile. A binary crate is able to compile the example code just fine but when it's an example, it can't find the `unmask` function. Any expertise there would be appreciated.

sajattack:
Not sure what's up there.

TDHolmes:
After doing some `cargo clean`'s and updating my nightly compiler, both examples compile fine now... 🤷🏼‍♂️ This PR should be 100% good to go now.

sajattack:
LGTM. Thanks.

sajattack:
Possible future improvement https://github.com/FluenTech/embedded-time

PR #206, opened 2020-06-16
examples: rtfm - rtic rename

jacobrosenthal:
I only tested pygamer, but change is very minimal for everything except samd21_mini


sajattack:
The CI script for metro_m0 needs updating with the rename.

PR #207, opened 2020-06-27
Convert timers to using period instead of frequency

TDHolmes:
By using frequency to track a timer's period, you limit yourself in the possible timer options. For one, a timer longer than 1s is not possible, and many fractional periods are not possible. There are some instances where frequency works a little better, for example 3 Hz which has a rounding error if you convert to microseconds, but that error is extremely small.

I was able to test this on a feather_m0 board, but not the other boards.

This change came out of a discussion in #205 around the accuracy of timers, and the ability to set timers for > 1s

### DIFF
--- a/hal/src/samd11/timer.rs
+++ b/hal/src/samd11/timer.rs
@@ -177,12 +177,12 @@ impl TimerParams {
         Self::new_from_ticks(ticks)
     }
 
-    pub fn new_us<T>(timeout: T, src_freq: Hertz) -> Self
+    pub fn new_us<T>(timeout: T, src_freq: u32) -> Self
     where
         T: Into<Microseconds>,
     {
         let timeout = timeout.into();
-        let ticks: u32 = (timeout.0 as u64 * src_freq.0 as u64 / 1_000_000_u64) as u32;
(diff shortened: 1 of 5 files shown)


hal/src/samd11/timer.rs
```diff
@@ -174,6 +174,20 @@ impl TimerParams {
...
         let ticks: u32 = src_freq / timeout.0.max(1);
+        Self::new_from_ticks(ticks)
+    }
+
+    pub fn new_us<T>(timeout: T, src_freq: Hertz) -> Self
+    where
+        T: Into<Microseconds>,
+    {
+        let timeout = timeout.into();
+        let ticks: u32 = (timeout.0 as u64 * src_freq.0 as u64 / 1_000_000_u64) as u32;
```
TDHolmes, line 185: I went back and forth on this one. I initially had:
```rust
let src_frq_mhz: MegaHertz = src_freq.into();
let ticks: u32 = timeout.0 * src_freq_mhz.0;
```
which avoids `u64` shenanigans (to avoid overflow), but for any source frequencies less than 1 MHz, or even non-multiples of 1 MHz, this introduces rounding errors. Personally, doing a little bit of `u64` math isn't too bad, but curious what people think. Currently I don't think it's even possible to have a non-48 MHz clock :p

PR #209, opened 2020-07-01
Add Arduino Nano 33 IOT board

guswynn:
This is very similar to the mkr zero board, but ends up being somewhat different.

Couple notes:

- I don't really have a way right now of testing every pin, just the led and usb, so there may be inaccuracies. As far as I can tell, this is considered somewhat understandable in the embedded rust world, but it would be nice if someone could cross check: https://github.com/arduino/ArduinoCore-samd/blob/master/variants/nano_33_iot/variant.cpp against my pins.
- I found that variant.cpp file hard to read, so some of my comments about TC vs TCC on pins may be wrong.
- **Many pins double or triple up in functionality on this board, so I need help with naming**
- This board is highly under-documented, and for the life of me I couldn't find a reference on its bootloader size. I used the same value as the mkrzero, and it seemed to work
- You can see here: https://content.arduino.cc/assets/NANO33IoTV2.0_sch.pdf that this chip, unlike the mkrzero doesn't have an external clock on xin and xout. This is curious, but requires the use of with_external for the clock. That took me a while to figure out lol, I don't have a debugger so panic's are basically "the led didn't blink"
- I opted for an example of usb serial for logging, as that's what I need to debug in the future, and it was slightly subtle, with the "maybe not needed but better" `free` block, and the need to read blind data in the interrupt to make `picocom` cooperate.
- I added some notes and improvements to the mkrzero and nano33iot docs to make it easier for people to pick up.
- This board has no documentation other than forum posts on the double-press to get it into a flashable bootloader mode, so I added it to the docs.

In the future, I plan on using https://lib.rs/crates/wifi-nina to get wifi to work (maybe with tls, we shall see) as well as write my own driver crate for the lcm imu on board that there seems to be no support for in rust-embedded yet.

### DIFF
--- a/boards/arduino_nano33iot/README.md
+++ b/boards/arduino_nano33iot/README.md
@@ -20,6 +20,8 @@ cargo build --release --example blinky_basic
 arm-none-eabi-objcopy -O binary target/thumbv6m-none-eabi/release/examples/blinky_basic target/blinky_basic.bin
 bossac -i -d -U true -i -e -w -v target/blinky_basic.bin -R
 ```
+(You may need to use `--port` with something like `/dev/ttyACM0` or `/dev/ttyACM1`
 
 #### Notes
- - It may help to double-press the center button to reset when re-flashing the device
+ - It may help to double-press the center button to reset when re-flashing the device. This sets the device in a bootloader mode.
+
--- a/boards/arduino_nano33iot/README.md
+++ b/boards/arduino_nano33iot/README.md
@@ -1,4 +1,4 @@
-# Arduino Mkrzero Board Support Crate
+# Arduino Nano 33 IOT Board Support Crate
 
 This crate provides a type-safe API for working with the [Arduino nano 33 IOT board](https://store.arduino.cc/usa/nano-33-iot).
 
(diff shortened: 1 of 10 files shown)


guswynn:
Should I be the author of the crate?

sajattack:
Yes, you should.

boards/arduino_nano33iot/README.md
```diff
@@ -0,0 +1,28 @@
+# Arduino Mkrzero Board Support Crate
```
sajattack, line 1: Forgot to update the title here.

guswynn:
@sajattack sorry for the delayed fix, let me know if anything else is needed to merge!

sajattack:
Can you make an account on crates.io so I can add you as an owner?

guswynn:
@sajattack just made an account and linked my github!

guswynn:
@sajattack quick question as Im new to crates.io
I see you added more owners to the crate (including this repo, nice!)
Is that accomplished by locally adding owners to the Cargo.toml and using `cargo publish` or is there some other way?

sajattack:
`cargo owner --add`

PR #211, opened 2020-07-19
SPI: set baud, phase, polarity and disable/enable

mvniekerk:
On SERCOM spi, add methods to set the baud rate and phase/polarity of the SPI. Also add enable/disable methods.


sajattack:
good job

PR #217, opened 2020-07-29
Update PACs for all devices

jessebraham:
Generated new PACs using `./update.sh` and bumped crate versions to `0.7.0`.


jessebraham:
A second PR based on this one will be opened which updates the HAL and board crates, which should make Travis happy.

PR #218, opened 2020-07-29
Update HAL to use new PACs

jessebraham:
This updates the HAL to use the new PACs generated in #217.

The structure of a number of a peripherals have changed with these new PACs, but seemingly only really for the ATSAMD51/ATSAME54 devices. Because of this, I have had to refactor the common module to further make a distinction between `thumbv6m` and `thumbv7em` targets; this applies mostly to `gpio.rs`. While I was doing this, I took the time to refactor out some more low-hanging fruit in the same manner.

I have confirmed the HAL and all BSPs build on my local system, and have run a couple of tests on an ATSAMD21 device. We will need some people with ATSAMD51/ATSAME54 devices to do some further testings to verify no regressions have been introduced. Also I'd be surprised if I managed to make this many changes without any mistakes, so some code review would be appreciated.


jacobrosenthal:
Tried a bunch of stuff on pygamer d51 seems good
(neopixel timing is bad on debug again, but thats in master as well)
adc, buttons, leds, neopixels, spi, timer, i2c

sajattack:
Why did you choose to split up common rather than move to src/samxxx?

jessebraham:
The D11/D21 implementations were the same, and the D51/E54 implementations were the same; I honestly just didn't know how else to split them up. I'm happy to change it if you feel there's a better way to structure things.

sajattack:
Ah ok, I guess that makes sense then.

PR #220, opened 2020-07-30
timer: nanosecond base

jacobrosenthal:
as of https://github.com/atsamd-rs/atsamd/pull/207 clocks were based on microsecond base with integer storage 
https://github.com/atsamd-rs/atsamd/pull/207/files#diff-50f81082e17d3ce037e13f0f58cf454fR174
which means 1_000_000/2_000_000 = .5 :(

To support both 1hz and 1mhz I think wed have to base on a nanosecond instead of microsecond
1000mhz is 1ns, so were good up to that
1hz in nanoseconds is 1_000_000_000 and is still under u32 max 4_294_967_295u32 
so I think were ok?

### DIFF
--- a/hal/src/common/time.rs
+++ b/hal/src/common/time.rs
@@ -32,6 +32,10 @@ pub struct Miliseconds(pub u32);
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub struct Microseconds(pub u32);
 
+/// Nanoseconds
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct Nanoseconds(pub u32);
+
 /// Extension trait that adds convenience methods to the `u32` type
 pub trait U32Ext {
     /// Wrap in `Bps`
@@ -54,6 +58,9 @@ pub trait U32Ext {
 
     /// Wrap in `Microseconds`
     fn us(self) -> Microseconds;
+
+    /// Wrap in `NanoSeconds`
+    fn ns(self) -> Nanoseconds;
 }
 
 impl U32Ext for u32 {
@@ -87,6 +94,10 @@ impl U32Ext for u32 {
 
     fn us(self) -> Microseconds {
         Microseconds(self)
+    }
+
+    fn ns(self) -> Nanoseconds {
+        Nanoseconds(self)
     }
 }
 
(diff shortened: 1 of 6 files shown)


jacobrosenthal:
cc @TDHolmes

WyseNynja:
I was about to open an issue, but I think this PR fixes it.

I had a timer that was being setup with `timer.start(3.mhz())`, but with https://github.com/atsamd-rs/atsamd/pull/207, I get this error:

```
the trait bound `hal::time::Microseconds: core::convert::From<hal::time::MegaHertz>` is not satisfied
required because of the requirements on the impl of `core::convert::Into<hal::time::Microseconds>` for `hal::time::MegaHertz`
```

How should I fix this? Will nanoseconds work? Will that have a rounding error?

jacobrosenthal:
I think this this will fix that, but only secondarily, because I implemented megahz into nanoseconds. Give it a shot

jacobrosenthal:
if you want to keep using upstream I think 3_000_000.hz() or 3_000.khz() will work

sajattack:
I need you to rebase again.

jacobrosenthal:
done

sajattack:
Thanks!

WyseNynja:
This PR is working for me! Thanks!

Both my timers are working now.

```
light_timer.start(3.mhz());
elapsed_ms_timer.start(1.ms());
```

TDHolmes:
didn't consider people doing megahertz timers... my bad! I think this change makes sense.

TDHolmes:
The only thought that comes to mind is this limits the timer to ~4s durations, however for atsamd running at 48 MHz, the HW is limited to ~1.3s, so not really a big deal. If people clocked the micro at a lower frequency however, they'd hit the container limitation. Perhaps `u64` container is in order?

jacobrosenthal:
I think timer math could happen really often or in interrupts contexts. u64
math during these is probably not tolerable for most?
Nice catch though.
The fact is one abstraction might not be able to cover everything. Maybe we
need a high resolution and low resolution timers instead of one.

TDHolmes:
Yeah good point. I personally don’t have a use case for such a timer so
keeping it simple for now makes sense to me. Good abstractions for all are
definitely tricky!

PR #221, opened 2020-07-31
Pygamer and edge neopixels (probably the rest too)

jacobrosenthal:
Draft to discuss if someone else can do better than me.

Sigh neopixels are strange again. Both the timer version and the spintimer show white on debug.Still fine on release.

I can sort of hack them working with slow feature on ws2812 timer and optimizations, but then they look bad on release mode. And I dont feel great about optimizing all debug builds? Or maybe it doesnt matter?

I tried every combination of dev dependencies and I can not find a dep override that works!
```
 [profile.dev.package.micromath]
 opt-level = "s"

 [profile.dev.package.ws2812-timer-delay]
 opt-level = "s"

 [profile.dev.package.smart-leds]
 opt-level = "s"
```
```
[profile.dev.package."*"]
opt-level = 3 # enable optimizations for all dependencies for neopixel
```
But no go

Thoughts? Im leaving everyone elses boards to them.. who knows maybe yours are fine?

### DIFF
--- a/boards/edgebadge/Cargo.toml
+++ b/boards/edgebadge/Cargo.toml
@@ -1,9 +1,7 @@
 [package]
 name = "edgebadge"
 version = "0.5.0"
-authors = [
-    "Jacob Rosenthal <@jacobrosenthal>",
-]
+authors = ["Jacob Rosenthal <@jacobrosenthal>"]
 description = "Board Support crate for the Adafruit EdgeBadge"
 keywords = ["no-std", "arm", "cortex-m", "embedded-hal"]
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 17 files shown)


sajattack:
The timer version is broken?? That seems strange. Prevously it was mainly the nop ones that broke on optimizations. Has @david-sawatzke changed anything lately?

jacobrosenthal:
It's less.its broken as I can't find a working setup for.both debug and
release any more

david-sawatzke:
No idea, haven't changed anything. Probably compiler & llvm changes in debug mode. Might be a good idea to add a less-hacky, more hardware specific version to atsamd-hal, for example a pwm-based one with less timing issues, similar to https://github.com/jamesmunns/nrf-smartled

tarcieri:
Should I check out the Trellis M4 again? The neopixels, with either driver, have always been "work most of the time but somewhat wonky"

jacobrosenthal:
updated opt to 3. This combo is working decent for me these days with timer and spintimer

Issue #214, opened 2020-07-20
Refactor Pin and Pad implementation


BroderickCarlin:
I wanted to open this issue to discuss the extent that this crate uses macros and possible alternatives. Macro's tend to be more difficult to read and reason about while also not being as well supported by Rust's documentation system. It is my belief that because of this it may be beneficial to explore moving to code generation using some external scripts (such as Python) to generate the valid Rust code but I am curious to hear others thoughts of if this has been attempted previously

wez:
While it is true that macros can be hard to read, and that it is easy to write macros that don't play as nicely with rust's documentation system, it is also true that they can be made clear and made to pass through documentation attributes and generate useful, meaningful documentation.

Rather than having a philosophical discussion, I think it would be useful if you could enumerate concrete examples of problems with macro usage in this crate.  That would give a good starting point for an objective evaluation of how well things are working out!

bradleyharden:
When adding support for a new chip (#243) this past weekend, I also struggled to wrap my head around what was going on with the `SERCOM` instances. I'm very new to Rust in general, so I wasn't sure if it was just me, or if the code could be better structured. It sounds like it's not just me.

It's all still very fuzzy to me, but it seems like the `Pins` trait is what we're missing compared to the `stm3214xx` HAL. On top of that, we also have to deal with the different IOSETs, which is what @blm768 is referring to. Did I get all of that right, or am I misunderstanding?

Before even seeing this thread, I was going to ask if it would make sense to introduce an IOSET trait. If I understand correctly, it's currently possible to mix pins from different IOSETs without the compiler complaining, right?

keithduncan:
I added some of the macros in the sercom module to support more typestate programming at the sercom pad level. I agree they are complex, they took me a while to wrap my head around, and I’d be open to ideas to make them easier to understand while preserving the type guarantees currently supported 😄 if more traits or docs would help let me know.

---

As an example I have the following code in a board support crate I’ve made for the Arduino MKR NB1500:

```
// SERCOM4 => UART => Sara

pub extern crate sara_r4;

pub type SaraRx  = mcu::gpio::Pa13<mcu::gpio::PfD>;
pub type SaraTx  = mcu::gpio::Pa12<mcu::gpio::PfD>;
pub type SaraRts = mcu::gpio::Pa14<mcu::gpio::PfD>;
pub type SaraCts = mcu::gpio::Pa15<mcu::gpio::PfD>;

[...]

PR #222, opened 2020-07-31
add feather_m0 pins d3, d4, d8

WyseNynja:
I have a feather_m0 LoRa and it uses pins 3, 4, and 8 for the Lora Module Control: https://learn.adafruit.com/assets/46254

I haven't yet tested this on actual hardware, but I'm pretty sure it is correct since it matches adafruit's pinout diagram.


sajattack:
My feather m0 doesn't have these pins exposed.

WyseNynja:
Yeah, not all of them do. I also have adafruit's feather m0 basic proto and it doesn't expose them: https://learn.adafruit.com/assets/46244

Should these maybe be behind a feature?

rubberduck203:
I feel like it would be good to put them behind a feature.

sajattack:
I'm not sure if that's possible with our current macro for pins.

TDHolmes:
I don't see the harm in having the "superset" of feather_m0 pins 🤷🏼‍♂️ Might as well keep it simple and have the superset, so long as nothing collides.

WyseNynja:
I could add a comment saying something like "not exposed on all models"


Bryan

sajattack:
I suppose that's reasonable, but some people might feel like it violates rust's safety guidelines.

WyseNynja:
Is using a non-existent pin actually unsafe?

I need these pins for my project, but features aren't an option. So what's the alternative?

TDHolmes:
Agreed. A pin will toggle on a disconnected pin of the micro. Seems fine to
me.

sajattack:
Yeah I guess it's the best we can do. Please add a comment at least.

rubberduck203:
> I'm not sure if that's possible with our current macro for pins.

@sakattack what makes you unsure? Has anyone tried?

sajattack:
@rubberduck203 macros require fixed tokens for syntax and I don't think #[cfg(feature="whatever")] would be part of the macro without modification.

jessebraham:
I had to modify `pin!()` recently to accept `#[cfg()]`, it's not too difficult. See lines `16` and `36` here:
https://github.com/jessebraham/atsamd/blob/8812146ab697d769e52f4383b821810a6d3e18d6/hal/src/common/pad.rs

jessebraham:
While working on an unrelated task I noticed that `define_pins` does seem to support feature gating:
https://github.com/atsamd-rs/atsamd/blob/master/hal/src/common/thumbv6m/gpio.rs#L569-L602

Reformatting the macro arguments makes this a bit more clear:

```rust
#[macro_export]
macro_rules! define_pins {
    (
        $(#[$topattr:meta])*
        struct $Type:ident,
        target_device: $target_device:ident,
        $(
            $(#[$attr:meta])*
            pin $name:ident = $pin_ident:ident
        ),+ ,
    ) => {

$crate::paste::item! {
    $(#[$topattr])*
    pub struct $Type {
        /// Opaque port reference
        pub port: Port,

        $(
            $(#[$attr])*
            pub $name: gpio::[<P $pin_ident>]<Input<Floating>>
        ),+
    }
}
```

Notice the `$(#[$attr:meta])*` and `$(#[$attr])*` for each pin definition. Anyways, just FYI.

TDHolmes:
macro-confusion kind of makes this related to #214

henrikssn:
I have the same problem for the RFM69 version of the Feather M0 board.

ianrrees:
I've made https://github.com/atsamd-rs/atsamd/pull/319 , which adds these pins behind an "rfm" feature flag as suggested above

Issue #213, opened 2020-07-20
Instantiating sercom with frequency is misleading at best and broken at worst


BroderickCarlin:
After spending some time attempting to debug a SPI issue I came to the realization that the instantiated SPI bus's frequency was not inline with what had been specified during instantiation. Experimenting with passing in different `freq` values during initialization resulted in non-trivial changes to the realized SPI bus's frequency. Digging into [the code](https://github.com/atsamd-rs/atsamd/blob/48b6624c0ea51d9ca568d5b206165ece45a31c6a/hal/src/samd21/sercom/spi.rs#L136) it became evident that the frequency parameter passed in was being converted to a baud rate internally. This hidden conversion does not appear to be doing proper bound checking and as such seemingly valid inputs result in non-discriminate outputs. 

For example, let's imagine a user wanted to have an exceptionally slow clock for some sort of visualization of SPI traffic. The code as written today would entirely allow for a frequency of `1.hz()` to be passed in. However, if we work through the math of how this value would be interpreted we can see that it would not actually yield a bus frequency of, or near, 1hz. 

My suggestion would be to revisit this math and add proper bounds checks

sajattack:
Sounds like a good idea.

sajattack:
@TDHolmes since you have the most recent experience mucking with the timers, would you be willing to take this on? I'm thinking making TimerParams fallible (return a Result) is probably a good approach.

BroderickCarlin:
As a sort of later thought on this, it may be beneficial to have an accessor method to see what the actual frequency is that gets set. Adding bounds checks, or making it fallible, in conjunction with allowing a developer to access what the actual clock speed is that gets set would be a huge improvement imho. 

@TDHolmes, thoughts?

sajattack:
Haha I just realized you were talking about sercoms not timers.

TDHolmes:
haha, I mean it is somewhat related :p The proposed solution makes sense. I'm down to take a crack at it

TDHolmes:
Looks like we could use [TryFrom](https://doc.rust-lang.org/beta/core/convert/trait.TryFrom.html) to attempt the conversion, and error out is it won't fit. Do we have a MSRV? Looks like this was added in 1.34

sajattack:
I'm actually not sure if an error is the best fit. Maybe just let the user know what they're actually getting.

BroderickCarlin:
yeah; personally I think just letting the user know what frequency was actually set would be very beneficial

[...]

PR #224, opened 2020-08-02
Add ability to query frequency of configured SPI peripheral

TDHolmes:
Partially addresses #213 by adding a helper method to query the current configured frequency of the SPI peripheral. I also reduced the scope of some `unsafe` blocks, and made the usage of the `spi()` helper method more consistent in the code, as well as distinguishing between mutable/immutable access there.


TDHolmes:
@BroderickCarlin can you try this branch out? I don't have any SPI stuff laying about to test this.

TDHolmes:
I'm curious what you guys think of `spi()` and `spi_mut()`. Both return `&SERCOM`, but borrow `self` differently to show intent, as you can modify state with `&SERCOM`. I made the two variants so I could do
```rust
pub fn freq(&self, src_clock: &clock::$clock) -> Hertz { ... }
```

instead of
```rust
pub fn freq(&mut self, src_clock: &clock::$clock) -> Hertz { ... }
```

TDHolmes:
@sajattack sorry for the churn, started playing around with refactoring common SPI logic out and I think I came up with a pretty good solution. I can pull this out into a separate later review if you want though. 

I wanted to move `set_baud` out too, but didn't want to change the existing API to make it happen (pass in source clock frequency instead of a reference to the clock)

TDHolmes:
Fair enough, how about `CommonSpiInterface` or something?

sajattack:
A bit long. Maybe Spi or CommonSpi

PR #227, opened 2020-08-07
Update uf2conv references to match its documentation

ianrrees:
Just a minor change - both the [referenced source](https://github.com/sajattack/uf2conv-rs) and the uf2conv installed by Cargo on my system reference a `uf2conv` not `uf2conv-rs`.


sajattack:
Thanks!

PR #231, opened 2020-08-11
Refactor ADC, PWM and uart_debug modules into common

jessebraham:
(Just as a quick note, there is still a fair bit more refactoring that can be done, I just chose to split it up since this is going to be a bit annoying to review as it is)

While doing some other work I noticed a fair bit of duplicated code that had accumulated. I've extracted some of it into the common module:

* For the ATSAMD11/ATSAMD21 the ADC and PWM modules were essentially identical, differing only in imports and macro invocations, both of which were easily solved using `#[cfg()]`
* For the ATSAMD51/ATSAME54 the ADC and PWM modules were essentially identical, differing only in imports and macro invocations, both of which were easily solved using `#[cfg()]`. The `uart_debug` functionality was exactly the same, including the pins/pads used for the debug UART.

Diffing the current `samd11/adc.rs`/`samd21/adc.rs` etc. shows quite clearly how similar these modules are, and should (hopefully) aid in the review process. I think that I managed to get things `cfg`-gated properly, but let me know if I've missed anything!


jacobrosenthal:
I checked pwm and adc examples on samd51 and theyre working as expected. Glanced at code didnt see anything surprising.

jessebraham:
Thanks for checking! Forgot to mention I did do a basic PWM test on my Trinket M0 as well, will do more testing when I have time.

jacobrosenthal:
Im trying to remember, but i think uart_debug is identical, because it was copy pasted with hardcoded pins and never worked on samd51. I could be wrong but thats my memory.

jessebraham:
Ahh that could be the case, I'm not sure to be honest :) If somebody is able to confirm I'm happy to change/revert if needed.

jacobrosenthal:
Or maybe rather, maybe it worked great for whoever wrote it for that specific set of Sercom pads. But Ive never seen it in use since, since Ive never had that specific set of pads exposed as uart.

jacobrosenthal:
I would say go ahead and combine it for now. Its the same (limited) functionality it had before the refactor. Just trying to provide context :)

jessebraham:
I've done a bit of additional testing of ADC/PWM on my Trinket M0, things seem to be in order as far as I can tell.

PR #232, opened 2020-08-11
panic_led feature to light d13 on panic

jacobrosenthal:
I often have a debugger but folks in users groups dont and this keeps coming up. 

I keep thinking of doing a panic to screen, or use panic persist and then read to screen on next reboot, but that would sacrifice memory and a non standard memory.x just doesnt fit in for general usage. I think this is the next best thing. 

I thought maybe I could put this in a crate but invariably it needs to have access to pac or hal, and I was getting duplicated interrupts and things. Maybe its solvable? Also it would have to keep in sync with versions. So maybe just in a bsp is a good place after all?

Also Id be happy to use the hal to set the led if I could find a way to conjure the gpio pin. Anyone know a way to do that?

### DIFF
--- a/boards/pygamer/Cargo.toml
+++ b/boards/pygamer/Cargo.toml
@@ -62,6 +62,7 @@ unproven = ["atsamd-hal/unproven"]
 usb = ["atsamd-hal/usb", "usb-device", "usbd-serial"]
 sd-card = ["embedded-sdmmc"]
 math = ["micromath"]
+panic_led = []
 
 [profile.dev]
 # opt-level = 2 # uncomment for neopixel functionality during debug
--- a/boards/pygamer/examples/blinky_basic.rs
+++ b/boards/pygamer/examples/blinky_basic.rs
@@ -3,6 +3,7 @@
 #![no_std]
 #![no_main]
 
+#[cfg(not(feature = "panic_led"))]
 use panic_halt as _;
 use pygamer as hal;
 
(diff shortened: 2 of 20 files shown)


sajattack:
Why not an external crate that takes an embedded_hal::OutputPin?

jacobrosenthal:
Maybe more general, but feels worse. Youd have to have some init macro? in addition to the import.

jacobrosenthal:
as discussed in matrix, while an led panic handler over gpio would be a nice crate, I sitll wouldn't want to use that here, because it adds something like this to already complex examples
```
#[cfg(not(feature = "panic_led"))]
use panic_halt as _;
...
    let mut pins = hal::Pins::new(peripherals.PORT);
    let mut red_led = pins.d13.into_open_drain_output(&mut pins.port);

    #[cfg(feature = "panic_led")]
    panic_init(red_led);
```
I think the bsp is exactly for making opinionated decisions based on the hardware so I think this is a nicer solution

```
#[cfg(not(feature = "panic_led"))]
use panic_halt as _;
```

sajattack:
I'm still not very convinced that the setup is too complicated, and I think it would be more idiomatic and beneficial as an external crate.

jacobrosenthal:
As an external crate it cant be used in any of the examples here, because then dual hal/pacs collide :(

sajattack:
I thought you could have an initializer that takes an embedded_hal::OutputPin and carry on from there but I guess I'm the one that's missing something.

jacobrosenthal:
Still testing on hardware again dont merge quite yet.

Excited for the fabled eh-1.0 when i intend to make the feature default and clear the examples up a bit more

PR #233, opened 2020-08-13
Add USB support to the mkrzero board

djmcgill:
This adds the USB feature to the board, and an example for using it. Also bumps the version number. Mostly copied from the other boards. I have tested the example on a real device:
https://i.imgur.com/tGtNPPz.png


sajattack:
I've got one of these too :smile:  https://github.com/atsamd-rs/atsamd/pull/234

sajattack:
You should probably add the usb feature to .travis.yml so it's tested.

djmcgill:
Okay the travis now runs with the usb and unproven features, and I added a (tested) PWM example too.

PR #235, opened 2020-08-13
Fix typo in pin assignment on nano33

jakewins:
Noted this while hacking on adding the mkrwifi1010; I think the current pin assignment has a typo?

According to page 3 on https://content.arduino.cc/assets/Pinout-NANO33IoT_latest.pdf, nina_resetn is at b8, not a8

### DIFF
--- a/boards/arduino_nano33iot/src/lib.rs
+++ b/boards/arduino_nano33iot/src/lib.rs
@@ -108,7 +108,7 @@ define_pins!(
     /// SPI (Lefacy ICSP) 4 / NINA SCK
     pin nina_sck = a15,
     pin nina_gpio0 = a27,
-    pin nina_resetn = a8,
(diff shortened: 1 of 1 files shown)


jessebraham:
Thanks for this! `b8` seems to be assigned to `sda` already. Unfortunately only one alias can be assigned to a pin at this time. Not sure on the best path forward, maybe we should just document this?

jakewins:
Oh, I see. And I guess hrm, removing the assignment might break someones build downstream.. I don't know that there's any other issue with the crate, I was just glancing at this file while writing the other PR and happened to notice this. 

Where would be a good place to document it?

jessebraham:
Just updating the docstring for the `sda` pin assignment to mention `nina_resetn` as well should probably be fine I think.

PR #239, opened 2020-08-18
I2c bus hangs

jacobrosenthal:
Im having issues with the lis3dh where its freezing up the bus. Im not sure its recoverable, I'll keep working on it. But it shouldnt just hang the device forever. I looked and sure enough adafruit's arduino core has a similar checking which I've cribbed here.

https://github.com/adafruit/ArduinoCore-samd/blob/master/cores/arduino/SERCOM.cpp#L545

Probably should go on all devices, but ill let the bikeshedding start first on if Buserror is the right error, etc first.

### DIFF
--- a/hal/src/samd11/sercom/i2c.rs
+++ b/hal/src/samd11/sercom/i2c.rs
@@ -6,8 +6,10 @@ use crate::target_device::sercom0::I2CM;
 use crate::target_device::{PM, SERCOM0, SERCOM1};
 use crate::time::Hertz;
 
+const BUS_STATE_UNKNOWN: u8 = 0;
 const BUS_STATE_IDLE: u8 = 1;
 const BUS_STATE_OWNED: u8 = 2;
+const BUS_STATE_BUSY: u8 = 3;
 
 const MASTER_ACT_READ: u8 = 2;
 const MASTER_ACT_STOP: u8 = 3;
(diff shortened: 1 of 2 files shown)


jacobrosenthal:
With this all I need is a clean way to signal via led that theres a panic so the user can click the rest button :)

jacobrosenthal:
The stuff to bikeshed here is, this changes waiting forever for the bus to be available to making the user look for.. say a BusError to retry in user code. Im not sure why, but hal could retry 5 or 100 or 1000 times for the user first for some kind of middle ground

sajattack:
I think just returning an error is fine. User code can choose to retry.

jacobrosenthal:
Rebased

I still dont have a feel for if we consider this breaking. There could be a device out there that was idle or busy for 1 or 1000 cycles through the loop but then recovered. So were breaking those devices by erroring on the first check.

Side note: for me, this error is unrecoverable. Were already resetting the i2c sercom on new and thats not shaking up the device. The pygamer doesnt put the device's power pin to a gpio so I cant reset it otherwise. In practice this fits well with my panic_led feature because the device will basically come up panic'ed which is shown on the led and I have users just toggle the power switch when that happens.

twitchyliquid64 (approved):

PR #240, opened 2020-08-21
Allow splitting USARTs

blm768:
This PR implements a `split` method on `USARTX`, much like in [other HAL crates](https://docs.rs/stm32l4xx-hal/0.5.0/stm32l4xx_hal/serial/struct.Serial.html).

Currently, only `samd51` is implemented. Once we're confident that this implementation is good, I can generalize it to the other variants.

### DIFF
--- a/hal/src/common/thumbv6m/sercom/v1/uart.rs
+++ b/hal/src/common/thumbv6m/sercom/v1/uart.rs
@@ -10,6 +10,7 @@ use crate::target_device::{SERCOM2, SERCOM3};
 use crate::target_device::{SERCOM4, SERCOM5};
 use crate::time::Hertz;
 use core::fmt;
+use core::marker::PhantomData;
 
 /// The RxpoTxpo trait defines a way to get the data in and data out pin out
 /// values for a given UARTXPadout configuration. You should not implement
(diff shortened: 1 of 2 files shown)


jessebraham:
I briefly reviewed this and everything seems pretty reasonable. I noticed a few `TODOS`, etc. in there but I'm sure you're aware, not sure how many pre-date your changes.

I did some testing on the ATSAMD51P19A, and (at least for the things I tried) it seems to be working! I'd like to get another set of eyes on this if possible, but so far things are looking good.

jessebraham:
What's the status of this? No rush or anything, I just wanted to check if you are still able/willing to complete this. No worries if not, of course.

blm768:
Oh, I was originally thinking I'd wait to finish until the USART stuff was refactored to use the new GPIO interface, and then I kinda forgot about this for a while. I'm still around and willing to finish this, though.

blm768:
Haven't tested this on hardware (my Wio Terminal has a flaky USART, and I'm not even sure where I put it at the moment), but I think it's ready.

sajattack:
Oops, forgot about this PR sitting at the bottom of the list. Would you mind making a quick example I can test for metro m4?

blm768:
OK, I just added a method to allow re-joining the split halves. I think that's everything for this PR.

PR #243, opened 2020-08-23
Add support for the ATSAMD51N20A

bradleyharden:
Hi,

I'd like to add support for the ATSAMD51N20A. I don't have much experience with Rust or with this series, but I took a stab at it anyway. I think it's complete, and `build-all.py` ran successfully on my system, so I'm crossing my fingers. Could you please review?

There were one or two areas where I wasn't sure exactly what to do. I'll add comments there.

Also, I think I found two missing SERCOM pads in SERCOM7. I added them, but I'm not sure if that belongs in this PR or a separate one. I'll point them out in a comment as well.


bradleyharden:
It looks like some refactoring was done in the last few days. I'll need to review it to see if I need to make any changes.

hal/src/samd51/sercom/pads.rs
```diff
@@ -240,37 +240,42 @@ pad!(Sercom5Pad3 {
 // sercom6[2]:  PC06:C   PC10:C   PC14:D   PC18:C   PD10:D
 // sercom6[3]:  PC07:C   PC11:C   PC15:D   PC19:C   PD11:D
 
-#[cfg(feature = "samd51p19a")]
+#[cfg(any(feature = "samd51p19a", feature = "samd51n20a"))]
```
bradleyharden, line 243: I think this is the correct way to handle this. Both N and P variants have the extra SERCOMs, but the N variant is missing some pins. Some of these pins should only apply when targeting a P variant. Will the attributes within the `pad!` macro override the one before the `pad!` macro?

hal/src/samd51/sercom/pads.rs
```diff
@@ -279,32 +284,38 @@ pad!(Sercom6Pad3 {
...
 // sercom7[3]:  PC11:D   PC15:C   PD11:C   PB19:D
 
-#[cfg(feature = "samd51p19a")]
+#[cfg(any(feature = "samd51p19a", feature = "samd51n20a"))]
 pad!(Sercom7Pad0 {
     Pb21(PfD),
     Pc12(PfC),
+    #[cfg(feature = "samd51p19a")]
     Pd8(PfC),
+    Pb30(PfC),
```
bradleyharden, line 293: If I understand the datasheet correctly, PB30 and PB31 are valid pins for SERCOM7, but they are missing here. Was that intentional or a mistake? I added them as a separate commit, so I can easily split the two commits into two PRs if required.

svd/devices/atsamd51n20a.xsl
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bradleyharden, line 1: It looked like this file was the same for each chip, so I just copied it for the new one. Is that correct?

hal/Cargo.toml
```diff
@@ -109,6 +114,8 @@ samd51j19a = ["atsamd51j19a", "samd51"]
 samd51j19a-rt = ["atsamd51j19a", "atsamd51j19a/rt", "samd51"]
 samd51j20a = ["atsamd51j20a", "samd51"]
 samd51j20a-rt = ["atsamd51j20a", "atsamd51j20a/rt", "samd51"]
+samd51n20a = ["atsamd51n20a", "samd51"]
```
bradleyharden, line 117: As a general question, would it make sense to add a set of features for `samd51g`, `samd51j`, `samd51n` and `samd51p` to cover the different pinouts?

jacobrosenthal:
Well have to defer to @BenBergman on when the right time to get this in is. hes doing a lot of refactoring right now as youve already come across once. What do you think ben, before you do more, or after?

jessebraham:
I believe Jacob meant to tag me instead. My work on that has stalled out for the time being, so feel free to merge whatever and I'll rebase my code when I get back to it.

jacobrosenthal:
Im not sure we have a test in CI, but if I run `cargo build --features=samd51n20a` locally it fails for me so check that out.

While you're at it split the PAC changes into one separate commit to clean up reviewability and lets not squash that work afterwards :)

It seems fine to me overall though.

svd/devices/atsamd51n20a.xsl
jessebraham: Yes, we just recently added the patching system so there are currently very few items. This looks correct.

hal/Cargo.toml
jessebraham: I think that might be a good idea, I was actually thinking the same the other day. @sajattack @jacobrosenthal thoughts?

hal/src/samd51/sercom/pads.rs
jessebraham: You appear to be correct, I would imagine this was a mistake.

jessebraham:
I'm also getting some build errors, specifically in `src/common/thumbv7em/gpio.rs`.

Running `build-all.py` builds all the Board Support Packages, so since no boards are currently targeting this chip the HAL was not being built for it.

To build specifically for this chip you can run:
```bash
$ cargo build --features=samd51n20a,unproven
```

bradleyharden:
Sorry, I didn't realize that the N variants don't have PORT group 3. It should be fixed now. Thanks for showing me how to build it without an example.

As a general question, how do you find out exactly which features each variant has? So far, I've been relying on connecting the dots between pinouts and other hints in the datasheet. But I haven't seen anywhere that explicitly states that "N and P variants have SERCOM 6 & 7", for example. Am I missing something?

Finally, I split out those missing SERCOM pins into a separate PR. I'll create it now.

hal/src/samd51/sercom/pads.rs
jessebraham: This is the correct approach, yes.

The top-level `#[cfg()]` attributes (that which you've quoted) will determine whether or not to invoke the `pad!` macro for the given `Sercom`/`Pad` combination. Then, within the `pad!` invocation, additional `#[cfg()]` attributes will determine whether or not to include that specific pin.

jessebraham (approved):
LGTM. Just a note that this variant is not currently checked by Travis, but I did confirm that it builds.

jessebraham:
Thanks for this!

PR #244, opened 2020-08-24
Add support for Dotstar on ItsyBitsy M0 

gordonshieh94:
Hello all!

I'd like to submit a request to add support for dotstar with an example demoing a spinning RGB color wheel. This example is identical to the CircuitPy script included when newly purchased from Adafruit. 
Compiled, tested and deployed on my board with rustc 1.40.0.

### DIFF
--- a/boards/itsybitsy_m0/Cargo.toml
+++ b/boards/itsybitsy_m0/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["Ben Bergman <ben@benbergman.ca>"]
 description = "Board Support crate for the Adafruit ItsyBitsy M0"
 keywords = ["no-std", "arm", "cortex-m", "embedded-hal"]
 license = "MIT OR Apache-2.0"
+edition = "2018"
 repository = "https://github.com/atsamd-rs/atsamd"
 readme = "README.md"
 documentation = "https://atsamd-rs.github.io/atsamd/atsamd21g18a/itsybitsy_m0/"
@@ -12,6 +13,9 @@ documentation = "https://atsamd-rs.github.io/atsamd/atsamd21g18a/itsybitsy_m0/"
 [dependencies]
 cortex-m = "~0.6"
 embedded-hal = "~0.2.3"
+bitbang-hal = "~0.3"
+apa102-spi = "~0.3"
+smart-leds = "~0.3"
 nb = "~0.1"
 
 [dependencies.cortex-m-rt]
@@ -42,7 +46,7 @@ embedded-graphics = "0.6.0"
 
 [features]
 # ask the HAL to enable atsamd21g18a support
-default = ["rt", "atsamd-hal/samd21g18a"]
+default = ["rt", "atsamd-hal/samd21g18a", "atsamd-hal/unproven"]
 rt = ["cortex-m-rt", "atsamd-hal/samd21g18a-rt"]
 usb = ["atsamd-hal/usb", "usb-device", "usbd-serial", "usbd-hid"]
 unproven = ["atsamd-hal/unproven"]
@@ -50,6 +54,9 @@ use_semihosting = []
 
 [[example]]
 name = "blinky_basic"
+
+[[example]]
+name = "dotstar"
 
 [[example]]
 name = "i2c_ssd1306"
--- a/boards/itsybitsy_m0/Cargo.toml
+++ b/boards/itsybitsy_m0/Cargo.toml
@@ -46,7 +46,7 @@ embedded-graphics = "0.6.0"
 
 [features]
 # ask the HAL to enable atsamd21g18a support
-default = ["rt", "atsamd-hal/samd21g18a", "atsamd-hal/unproven"]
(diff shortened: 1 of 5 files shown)


boards/itsybitsy_m0/examples/dotstar.rs
```diff
@@ -0,0 +1,60 @@
...
+    let mut delay = Delay::new(core.SYST, &mut clocks);
+
+    let mut pins = hal::Pins::new(peripherals.PORT).split();
+
+    let mut rgb = hal::dotstar_bitbang(pins.dotstar, &mut pins.port, SpinTimer::new(12));
+
+
+    let mut val = 0;
+    loop {
+        // Can't use because of running u8 % u16
```
jacobrosenthal, line 51: cant use what, wrapping add?

boards/itsybitsy_m0/src/lib.rs
```diff
@@ -97,6 +100,53 @@ define_pins!(
...
+}
+/// Dotstar pins
+pub struct Dotstar {
+    pub ci: Pa0<Input<Floating>>,
+    pub di: Pa1<Input<Floating>>,
+    pub nc: Pa12<Input<Floating>>,
+}
+
+impl Dotstar {
+    pub fn dotstar<T: CountDown + Periodic>(
```
jacobrosenthal, line 134: im a fan of the init nomenclature
https://github.com/atsamd-rs/atsamd/blob/master/boards/pygamer/src/pins.rs#L399

https://github.com/atsamd-rs/atsamd/blob/master/boards/pygamer/examples/neopixel_tilt.rs#L42

boards/itsybitsy_m0/examples/dotstar.rs
gordonshieh94: oops I'll revise that.

boards/itsybitsy_m0/src/lib.rs
gordonshieh94: I'll refactor and split `lib.rs` so that there's a `pins.rs` like the other boards too then.

boards/itsybitsy_m0/src/lib.rs
```diff
@@ -174,6 +107,18 @@ pub fn flash_spi_master(
     (flash, cs)
 }
 
+/// Convenience for setting up the dotstar LED using bitbang'ed
+/// SPI.
+pub fn dotstar_bitbang<T: CountDown + Periodic>(
```
jacobrosenthal, line 112: This doesnt do much anymore and can probably go right?

boards/itsybitsy_m0/Cargo.toml
```diff
@@ -42,7 +46,7 @@ embedded-graphics = "0.6.0"
 
 [features]
 # ask the HAL to enable atsamd21g18a support
-default = ["rt", "atsamd-hal/samd21g18a"]
+default = ["rt", "atsamd-hal/samd21g18a", "atsamd-hal/unproven"]
```
jacobrosenthal, line 49: im fine with this and did it for pygamer some time ago, but we should check with some other users to make sure it doesnt bother anyone. Unproven goes away entirely very shortly anyway either way.

boards/itsybitsy_m0/src/pins.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+/// Dotstar pins
+pub struct Dotstar {
+    pub ci: Pa0<Input<Floating>>,
+    pub di: Pa1<Input<Floating>>,
+    pub nc: Pb0<Input<Floating>>,
+}
+
+impl Dotstar {
```
jacobrosenthal, line 187: We dont test well for this on ci, but you're going to need to cfg unproven some of this so it builds even without unproven
Use `cargo build --no-default-features --features "rt,atsamd-hal/samd21g18a"` locally until that can build successfully

jacobrosenthal:
Feel free to @ me when you finish stuff up or I wont know.

gordonshieh94:
@jacobrosenthal made dotstar only compile with the `unproven` feature flag set only.

jacobrosenthal:
Thanks! Mind if we ping you in future for itsy related questions and testing? 
https://github.com/atsamd-rs/atsamd/wiki/Board-owner-list

gordonshieh94:
Sure, I'm okay with that!

bradleyharden:
@gordonshieh94, how did you get this to work? Pin PB00 isn't available on the SAMD21G. It's only available on the SAMD21J, according to Table 7-1 in the datasheet [here](http://ww1.microchip.com/downloads/en/DeviceDoc/SAM_D21_DA1_Family_DataSheet_DS40001882F.pdf).

bradleyharden:
I bet I know. `nc` is for not connected, and it's only there because the bit-bang SPI implementation doesn't have an optional data-in pin. It makes sense, then, why you would choose a pin that doesn't exist.

bradleyharden:
Ugh, I'm an idiot. It says that right in the comments...

PR #245, opened 2020-08-29
Add basic support for PWM using TCC on D51/E54

jessebraham:
This is almost verbatim from the `thumbv6m` module, just some minor adjustments. Initial testing looks promising, but I will be doing some more tonight. If somebody else is able to try it out that'd be appreciated!

### DIFF
--- a/boards/pygamer/Cargo.toml
+++ b/boards/pygamer/Cargo.toml
@@ -90,3 +90,11 @@ required-features = ["sd-card"]
 [[example]]
 name = "neopixel_easing"
 required-features = ["math"]
+
+[[example]]
+name = "pwm_tc4"
+required-features = ["math"]
+
+[[example]]
+name = "pwm_tcc0"
+required-features = ["math"]
--- a/boards/pygamer/examples/pwm_tc4.rs
+++ b/boards/pygamer/examples/pwm_tc4.rs
@@ -1,0 +1,62 @@
+//! Breath an led in and out without using the BSP split() method.
+
+#![no_std]
+#![no_main]
+
+use panic_halt as _;
+use pygamer as hal;
+
+use core::f32::consts::FRAC_PI_2;
+use hal::clock::GenericClockController;
+use hal::delay::Delay;
+use hal::entry;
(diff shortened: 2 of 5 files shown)


jessebraham:
@jacobrosenthal had pointed out last night that I was missing the pinout Enums for the TCC implementation, which take ownership of the pin once it has been set to the appropriate function. Should I add these in for the `thumbv6m` implementations as well, just for consistency since they don't currently exist?

jacobrosenthal:
If you can stomach it, it harmonizes all the thumb6 stuff with the thumb7 versions. Its obviously also 'safer' since we know you had to set your pin into the correct function. But more than that I was trying to implement an example to test yesterday and i had to drop to datasheet to figure out which timer offered which pin as an output, and thats discoverable by the thumbv7 implementations enums.

I updated my examples, but the pwm_tcc0 example isnt blinking my led.(where the tc0 version does) I dont have my scope here, perhaps you can see what im doing wrong?
https://github.com/jacobrosenthal/atsamd/tree/features/pwm

jessebraham:
> If you can stomach it, it harmonizes all the thumb6 stuff with the thumb7 versions. Its obviously also 'safer' since we know you had to set your pin into the correct function.

Sure thing, the `thumbv6m` chips tend to be a lot smaller anyways. I agree, it was a little awkward calling `.into_function_x()` before and just kinda not doing anything with the result.

> But more than that I was trying to implement an example to test yesterday and i had to drop to datasheet to figure out which timer offered which pin as an output, and thats discoverable by the thumbv7 implementations enums.

That's another excellent point, I found myself doing the same thing.

> I updated my examples, but the pwm_tcc0 example isnt blinking my led.(where the tc0 version does) I dont have my scope here, perhaps you can see what im doing wrong?
> https://github.com/jacobrosenthal/atsamd/tree/features/pwm

Looks reasonable to me, I may have made some errors. Let me review.

jessebraham:
@jacobrosenthal due to some difficulties in adding the enums for the `thumbv6m` I think they will be in a separate PR at this point. This should be good to merge if you're happy with it.

PR #246, opened 2020-08-30
Add missing SERCOM pins for SAMD51

bradleyharden:
Add missing pins PB30 and PB31 to the SERCOM HAL as options for SERCOM7
in SAMD51 chips


jessebraham:
LGTM, thanks!

PR #248, opened 2020-09-05
Add support for Dotstar on Trinket M0

ianmclinden:
### Summary
Adding Dotstar support & example to boards/trinket_m0.

Built and flashed to board with rustc 1.44.1

### Changelog
- Specified Rust 2018 edition (for `apa102_spi` and `embedded_hal` support)
  - Also migrated previous examples from extern to use syntax (all tested)
- Added examples/dotstar.rs based on previous work
- Updated Cargo.toml with required dependencies for dotstar
- Added Pins and convenience structs to lib.rs
  - Not separated to a pins.rs, but this could be done

### History
Based on previously accepted dotstar PR #244

### DIFF
--- a/boards/trinket_m0/Cargo.toml
+++ b/boards/trinket_m0/Cargo.toml
@@ -13,6 +13,9 @@ documentation = "https://atsamd-rs.github.io/atsamd/atsamd21e18a/trinket_m0/"
 [dependencies]
 cortex-m = "~0.6.2"
 embedded-hal = "~0.2.3"
+bitbang-hal = "~0.3"
+apa102-spi = "~0.3"
+smart-leds = "~0.3"
 nb = "~0.1"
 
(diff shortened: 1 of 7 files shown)


jessebraham:
Thanks for this!

The title says _Teensy M0_ but you've updated the _Trinket M0_. I'm assuming the title incorrect, just wanted to be sure though.

ianmclinden:
> Thanks for this!
> 
> The title says _Teensy M0_ but you've updated the _Trinket M0_. I'm assuming the title incorrect, just wanted to be sure though.

Totally correct (fixed). My head was somewhere else!

jessebraham (approved):
Not a problem, happens to me all the time 😄 

This seems to be working as intended on my Trinket, and everything is looking good to me, so I'm happy to merge it. Thanks again!

PR #252, opened 2020-09-12
fix(pwm): Limit pwm re-export to unproven

bradleyharden:
The SAMD51 and SAME54 HALs re-export the `pwm` module to avoid breaking
the previous, public API. The `pwm` module is currently unproven, but
the re-export occurred regardless, creating an error when building
without the `unproven` feature. Add a `cfg` attribute to limit the
re-export to builds with the `unproven` feature.

### DIFF
--- a/hal/src/samd51/mod.rs
+++ b/hal/src/samd51/mod.rs
@@ -7,4 +7,5 @@ pub mod usb;
 
 // This is merely to avoid breaking the public API.
 // Added after the PWM module was refactored into common.
+#[cfg(feature = "unproven")]
 pub use crate::pwm;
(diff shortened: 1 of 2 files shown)


TDHolmes:
Could we add a build configuration to CI to catch this from happening again?

bradleyharden:
I think that's a good idea, but I don't know the CI system. I also don't have any experience with Rust CI yet. I think it would be a lot easier and faster if one of the maintainers made that change.

jessebraham:
I think we probably could, but I'm not sure it's worthwhile at this point. `embedded-hal==1.0.0` gets rid of the `unproven` feature altogether.

jessebraham (approved):

PR #255, opened 2020-09-13
Add Arduino MKR1000 board support

NightWatchman:
I was able to load the blinky example on my board and verify proper operation.

I know nothing about travis builds, but I took a stab at updating `travis.yml`.


.travis.yml
```diff
@@ -17,6 +17,7 @@ env:
   - CRATE=boards/trinket_m0 FEATURES="--features=unproven"
   - CRATE=boards/samd11_bare FEATURES="--features=unproven" BUILDMODE="--release"
   - CRATE=boards/samd21_mini FEATURES="--features=unproven"
+  - CRATE=boards/arduino_mkr1000 EXAMPLES="--example=blinky_basic" FEATURES="--features=unproven,usb"
```
jessebraham, line 20:
```suggestion
  - CRATE=boards/arduino_mkr1000 FEATURES="--features=unproven,usb"
```
What you have is correct, but this change will ensure that any examples added in the future are built as well without needing to update the Travis config.

jessebraham (commented):
Thanks for this! There's one small nitpick I'd like to see addressed, but other than that it's looking pretty good.

If you're able/willing, it'd be much appreciated if you could add one or more of the below functions to `lib.rs` as well. If not it's no problem of course.

* `i2c_master`
* `spi_master`
* `uart`
* `usb_allocator`

NightWatchman:
I added a uart fn. I haven't tested it on the hardware, but the types seem to match up, and I should have the right pin numbers and such according to my pinout diagram. I'll look into implementing the other functions as well, but it'll probably take some time.

jessebraham (approved):
Great, thank you for making those changes! The `uart` function should be fine, the pins are correct and it builds so I'm assuming it's correct. I'm happy to merge this how it is now, so if you feel like adding the functions you are welcome to do so in another PR, and if not then that's no problem as well.

jessebraham:
@NightWatchman do you have any issue with me adding you to the [board owner's list](https://github.com/atsamd-rs/atsamd/wiki/Board-owner-list) in case we need somebody to test things for this board? No problem if you're not interested of cours!

NightWatchman:
> @NightWatchman do you have any issue with me adding you to the [board owner's list](https://github.com/atsamd-rs/atsamd/wiki/Board-owner-list) in case we need somebody to test things for this board? No problem if you're not interested of cours!

Sure that's no problem!

jessebraham:
@NightWatchman great, thanks! Also sorry but I totally forgot to publish this when I merged it! 😁 It's available on [crates.io](https://crates.io) now though, just FYI.

PR #265, opened 2020-09-13
SPI example with an ST7735 TFT display.

nebelgrau77:
Requires adding an 'assets' folder with the 'ferris.raw' file in it.


sajattack:
Please avoid making a new pull request for every little change.

sajattack:
Please also add the assets/ferris.raw file to this PR.

nebelgrau77:
OK, should be good now...

PR #271, opened 2020-09-19
Refactor calibration, clock, and sercom modules for SAMD51/SAME54

jessebraham:
These two device families are essentially identical other than a handful of small difference, all of which can be handled with `#[cfg()]`. This pulls out the `calibration`, `clock`, and `sercom` modules into the `common::thumbv7em` module.


jessebraham:
@mvniekerk would you be able to test this out on the SAME54 by any chance?

hal/src/common/thumbv7em/sercom/uart.rs
```diff
@@ -156,7 +156,8 @@ macro_rules! uart {
                             w.runstdby().set_bit(); // Run in standby
                             w.form().bits(0); // 0 is no parity bits
 
-                            // w.mode().usart_int_clk(); // Internal clock mode
+                            // TODO: does this need to be commented out to work on same54?
+                            w.mode().usart_int_clk(); // Internal clock mode
```
jessebraham, line 160: This line was commented out in the SAME54 implementation, but not in the SAMD51 implementation. Any ideas on why this was this case?

Sympatron:
Is something blocking this PR (apart from the merge conflict, that is not that hard to resolve)? I think it is really helpful and reduces code duplication a lot without breaking anything.
I wanted to write a PR that adds more variants of the SAME series, but that would be a PITA without this PR.

jessebraham:
I was hoping to get a test on the SAME54 for the one line mentioned above, and then forgot about it 😁 @sajattack once I rebase this are you okay with merging it as-is?

jessebraham:
Okay great, thanks. I'll try to get this done this weekend. There has been a new PAC added since I opened this PR, so unfortunately it's going to be a bit more involved than just rebasing, but still shouldn't take too long.

jessebraham:
I managed to hammer it out during my lunch break, I think it should be fine but I have yet to test on actual hardware. Will do that after work.

twitchyliquid64:
Looks good at first pass. Also spotchecked SERCOM5 and LGTM.

PR #273, opened 2020-10-04
Improve instructions for OSX users (and nudge people towards cargo-binutils)

guswynn:
I recently switched to using a macbook, so my own workflow was disrupted for my arduino_nano33iot

I updated the instructions to
1. give advice on where stuff is and how it works on macbooks (at least for arduino boards)
2. push people to use `cargo-binutils`, which is far easier to install on OSX, and somewhat nicer to use

as well as
3. fix a messed up double import for the `usb feature`

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ In addition to the PACs and HAL, there numerous **B**oard **S**upport **P**ackag
 
 Make sure that you have a new enough version of the gcc toolchain; the one installable even on recent versions of Ubuntu can fail to correctly link the vector table:
 
+**Note** you may be able to avoid this step if using `cargo-binutils`. See board specific instructions in `boards/` for more details.
+
 ```bash
 $ sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa -y
 $ sudo apt update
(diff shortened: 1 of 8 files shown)


jacobrosenthal:
Thanks. Those docs looks like they could use a refresher . I think you can clean it up even further though by using cargo objcopy For instance
https://github.com/atsamd-rs/atsamd/blob/53b3238d457864337b5c53fc01335efa5d868047/README.md

What do you think?

guswynn:
@jacobrosenthal I believe I fixed the objcopy instance in the top-level readme
are you suggesting that I just replace all instances of the gcc bin utils entirely with cargo-binutils (instead of leaving docs for either or?)

jacobrosenthal:
Yeah unless youve found it insufficient as we use it commonly in other bsps and would clean up that file a ton bringing the more cargo native approach out of comments. But also preferring 'cargo objcopy' not `rust-objcopy` as its avery clean way to  both cargo build and generate your binary.

sajattack:
@guswynn Are you still around and able to finish this up?

guswynn:
@sajattack ah I was just thinking about this! Sorry it fell off the radar, I'll try to get to it today!

guswynn:
@sajattack this should be up to date, it suggests `rust-objcopy` and `cargo-objcopy` in all places that gcc tools are in the docs, I think its reasonable to start with just documentation for them as alternatives, as this is just to unblock people using osx, and later we can default the instructions to `cargo-binutils` once we have more time to test

sajattack:
Thanks!

PR #274, opened 2020-10-08
Add support for ATSAMD51P20A chip and Adafruit Grand Central M4 Express board.

noisypebble:
I implemented support for the ATSAMD51P20A chip and the Adafruit Grand Central M4 Express board. The blinky_basic example was built and successfully flashed to my board. I split the chip and board support into two separate commits.


boards/grand_central_m4/Cargo.toml
```diff
@@ -0,0 +1,68 @@
...
+use_uart_debug = ["atsamd-hal/use_uart_debug"]
+
+[profile.dev]
+incremental = false
+codegen-units = 1
+debug = true
+lto = true
+
+[profile.release]
+debug = true
```
sajattack, line 62: I think this `debug = true` was something I left in the metro_m4 crate by accident, which presumably you copied.

boards/grand_central_m4/src/pins.rs
```diff
@@ -0,0 +1,749 @@
...
+    pin d9 = b2,
+    /// Pin 10, PWM capable
+    pin d10 = b22,
+    /// Pin 11, PWM capable
+    pin d11 = b23,
+    /// Pin 12, PWM capable
+    pin d12 = b0,
+    /// Pin 13, PWM capable. Also attached to onboard LED.
+    pin d13 = b1,
+    //pin d14 = b16,
```
sajattack, line 92: Lots of stuff commented out in this file, could you explain why?

sajattack (changes requested):

boards/grand_central_m4/Cargo.toml
noisypebble: yeah I'm still trying to get a handle on how all these pieces fit together. I've removed it in the latest commit

boards/grand_central_m4/src/pins.rs
noisypebble: Sure, the commented pins are those that are defined with friendlier names elsewhere. Looking at the other boards it seemed that most preferred the friendlier name (i.e. uart0_rx instead of d0). I'm adding comments to make this clear. I'll push a new commit soon.

noisypebble: I cleaned up the pin documentation and did some minor refactoring. I also noticed that putting Digital in the pin Set was not consistent between boards. I choose to remove it.

.travis.yml
```diff
@@ -32,6 +32,7 @@ env:
   - CRATE=boards/edgebadge FEATURES="--features=unproven"
   - CRATE=boards/xiao_m0 EXAMPLES="--example=blink --example=usb_serial" FEATURES="--features=usb"
   - CRATE=boards/wio_lite_mg126 FEATURES="--features=unproven,usb"
+  - CRATE=boards/grand_central_m4 FEATURES="--features=unproven" EXAMPLES="--example=blinky_basic"
```
sajattack, line 35: This `EXAMPLES` section is for when we want to only test certain examples out of all the ones that could ever exist. This prevents future examples from being tested so I would like it removed please.

sajattack (changes requested):

.travis.yml
sajattack: I think after this last change this will be good to go.

noisypebble: I made the requested change to .travis.yml

sajattack:
Couple of warnings from CI 
2020-10-10-190026_533x310_scrot

sajattack (approved):

twitchyliquid64:
Looks good!

I noticed theres the `usb` feature plumbed into the cargo file, and a `impl USB` that initializes it as a USB device.

If this works can we copy pasta one of the examples (like USB serial echo) from another crate + update the travis config to CI with USB enabled?

If it doesn't, can we either add a comment that its untested or remove the USB plumbing?

noisypebble:
> Looks good!
> 
> I noticed theres the `usb` feature plumbed into the cargo file, and a `impl USB` that initializes it as a USB device.
> 
> If this works can we copy pasta one of the examples (like USB serial echo) from another crate + update the travis config to CI with USB enabled?
> 
> If it doesn't, can we either add a comment that its untested or remove the USB plumbing?

Thanks @twitchyliquid64 

I copied in usb_serial.rs from pygamer and modified it accordingly. It's building successfully but I'd like to flash and test it on my board before I push a commit. I'll work on that today.

twitchyliquid64 (approved):

sajattack:
Uploaded to crates.io https://crates.io/crates/grand_central_m4 https://crates.io/crates/atsamd51p20a

Issue #278, opened 2020-10-16
Refactor SAMD51/SAME54 USB modules into `common::thumbv7em::usb`


jessebraham:
Once #271 is merged, the `samd51` and `same54` modules will consist only of the `usb` module and the PWM hack. If we can refactor the `usb` modules out to `common::thumbv7em::usb` then we can eliminate the `samd51`/`same54` modules altogether, since they share a common datasheet anyways.

__Edit:__ the aforementioned PR has been merged.

jessebraham:
Closed by #279

PR #279, opened 2020-10-16
Refactor duplicated samd5x USB into common crate

twitchyliquid64:
Looks like the same54 one was untested / out-of-date anyway.

It builds:

```shell
[xxx@xxx]:~/projects/atsamd/hal> cargo build --features "same54p20a,same54,usb,same54p20a-rt"

<snip>
warning: 10 warnings emitted

    Finished dev [unoptimized + debuginfo] target(s) in 34.78s

```

But theres no USB example to test it with, nor do I own any hardware using it.


jessebraham (approved):
Thanks! I've tested this on my Wio Terminal and things seem to be working fine, as expected.

You are correct that the `same54::usb` module was never actually used in any examples. Because of this I'm fine merging as-is, because we can't be sure of its state prior to this PR anyways.

This closes #278.

PR #281, opened 2020-10-18
Added PACs and corresponding features for all SAMD51 and SAME5x chips

Sympatron:
Since we plan on using different SAMD51 and SAME5x chips with Rust in the future, I thought it would be a good idea to have equal support for all of them.

While I was at it I added package variant features to make it easier to compile code for certain peripherals conditionally. Otherwise the filter expressions would have become unmaintainable.
I am pretty open regarding the naming scheme of these features. This (`samx5xgxxa`, etc.) is the best I could come up with, but I am very open to suggestions.

I also noticed that the previous ADC implementation did not let you use ADC1 fully for some chips and fixed that.


sajattack:
Wow, that's a big PR. When we first saw this, we were worried about ballooning build times but that doesn't seem to have happened. So, initially horrified, now optimistic. Thanks for the PR.

sajattack:
One area we're still worried about is automation of releasing all these PACs. If you could help in that area, that would be great too.

sajattack:
Another issue is we have limited resources to maintain all the features for all these chips.

sajattack:
It looks like you copied the Cargo.toml with another contributor's name on it. Could you replace this with your name?

Sympatron:
I just looked at the STM32 repositories to figure out what they did. Here are my findings:
- They have one [seperate repository](https://github.com/stm32-rs/stm32-rs) for all PACs. In my opinion a monorepo is fine too, that's just what they decided to do.
- To generate and publish the code they use a bunch of [python scripts](https://github.com/stm32-rs/stm32-rs/tree/master/scripts) and [`make`](https://github.com/stm32-rs/stm32-rs/blob/master/Makefile). Could be an easier option than github actions and the like.
- They don't differentiate by flash size, because just like Atmel SAM MCUs the registers and peripherals don't differ between versions with differents amount of flash. It is probably easier to maintain and easier for the users to do the same and switch from e.g. `samd51g18a` to just `samd51g`. The would eliminate more than half of the current PAC code.

PR #283, opened 2020-10-20
Rename PACs and features

Sympatron:
Because the last 3 characters of the chip name are only relevant for memory size and don't affect the PAC code generated by svd2rust this PR renames the PACs from e.g.  `atsamd21g18a` to `atsamd21g`. This results in 2 crates less to maintain.


sajattack:
This would be a pretty major breaking change for downstream users, and we already have the crate names published on crates.io 😕

Sympatron:
I know this not a decision to take lightly. But technically you are not at 1.0 yet, so breaking changes are to be expected.
Additionally, if users use the board crates they won't even notice, because they implicitly depend on the correct PAC.

I don't want to force this through. If the maintainers decide, they don't want to got this route that's fine by me. I just think it will be beneficial in the long run. Especially if you want to add support for other SAME5x devices.

It's also what the Rust STM32 ecosystem is doing, which is arguably the largest Embedded Rust group.

PR #289, opened 2020-10-25
wio-terminal: Add update script and firmware for ameba wifi/BLE core

twitchyliquid64:


sajattack:
Maybe the script should download these binaries on the fly so they stay up to date and don't pollute our repo.

twitchyliquid64:
I was hoping to keep copies, as this is exactly the kind of thing that disappears when authors move on. If we want to keep it out of this repo, would you be open to making a new repo like atsamd-rs/wio-terminal-tools ?

Issue #203, opened 2020-06-09
Type erasure for GPIO


AdminXVII:
As per https://rust-embedded.github.io/book/design-patterns/hal/gpio.html#pin-types-provide-methods-to-erase-pin-and-port-c-erased-pin, HAL crates should have means to downgrade the pins to a type-erased version. This is particularly useful for storing a list of pins in an array rather than a tuple.

I didn't find a way to perform this, so I'd like to request for help and/or for this feature. Thanks for your time.

jaredwolff:
Here's a link to a similar discussion. Looks like it's already implemented for some STM32 HAL: https://users.rust-lang.org/t/how-can-i-create-an-array-of-pins/44514

rubberduck203:
`stm32f3xx-hal` implements it like this. 

https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/gpio.rs#L140

https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/gpio.rs#L396

jessebraham:
Thanks for opening this issue.

I would also like to see type erasure for GPIO, as I have a handful of projects which could benefit from it myself. If somebody wants to take a swing at it that would be awesome! If not, I can take a look at implementing this feature whenever I get a chance; unfortunately I'm fairly busy with some other projects currently.

blm768:
It seems like the proposed changes for #214 would likely help with this; a common `Pin` trait seems like it would make a generic implementation of type erasure easier.

bradleyharden:
@Sympatron,

I agree with @blm768 that the changes in #214 make this easier. I didn't have any need for type-erasure myself, so I hadn't already implemented it. But since you're working on it, I thought I would demonstrate.

Here is the latest [version](https://github.com/bradleyharden/atsamd/blob/gpio_refactor/hal/src/common/thumbv7em/gpio_new.rs), with the type-erased pin found [here](https://github.com/bradleyharden/atsamd/blob/9c3cfd8dd96609ebda3fcdc15c9c9ed1bb46524f/hal/src/common/thumbv7em/gpio_new.rs#L783). I was able to easily duplicate the entire pin API for type-erased pins. You can also convert between type-level pins and value-level pins, see [here](https://github.com/bradleyharden/atsamd/blob/9c3cfd8dd96609ebda3fcdc15c9c9ed1bb46524f/hal/src/common/thumbv7em/gpio_new.rs#L1005).

[...]

PR #290, opened 2020-10-25
Gpio type erasure

Sympatron:
Implemented `downgrade` for gpio pins to solve #203.
Currently only thumbv7m support, but I think it could be ported to thumbv6m easily.

### DIFF
--- a/hal/src/common/thumbv7em/gpio.rs
+++ b/hal/src/common/thumbv7em/gpio.rs
@@ -85,6 +85,61 @@ pub struct PfM;
 /// Peripheral Function N
 pub struct PfN;
 
+pub enum Gpio {
+    PortA,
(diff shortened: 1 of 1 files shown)


sajattack:
Looks great! Can you explain the naming of the pin_x and pin_xx macros? It's not terribly clear what it means from the name.

sajattack:
Also, could you document the downgrade method since it is public?

Sympatron:
Just realized that the current implementation of `Pxx` is buggy. I'll try to fix it.

bradleyharden:
@Sympatron, now that #302 has been merged, will the `DynPin` API satisfy your needs?

Sympatron:
Yes this can be closed now.

PR #291, opened 2020-10-25
Rtc timer mode & standby operation of timers

TDHolmes:
Introduces a driver for running the RTC in timer compare mode. Also enables even lower power operation of both the RTC and the normal timer module. This tweaks the public API of `set_gclk_divider_and_source`. Tested both examples on `feather_m0` and they worked as expected in standby.

### DIFF
--- a/boards/edgebadge/src/pins.rs
+++ b/boards/edgebadge/src/pins.rs
@@ -749,8 +749,7 @@ pub struct ButtonReader {
 //tsu min setup time 55ns = 7 cycles
 //th min hold time 5ns = 1 cycles
 //tw min pulse width 36ns = 5 cycles
-//trec min recovery time 55ns, how long before you should attempt to read
-// again?
+//trec min recovery time 55ns, how long before you should attempt to read again?
 #[cfg(feature = "unproven")]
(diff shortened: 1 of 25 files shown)


Sympatron:
This contains a lot of formatting only changes. These are probably already handled by your PR #294. This makes it harder to spot the actual changes,

TDHolmes:
Yeah my editor automatically reformatted files in which I needed to change the clock API call I modified, and agreed it'll probably conflict with the other PR. I'll back out and reapply those changes to clean this PR up

TDHolmes:
cleaned up

sajattack:
Rather than changing the public API of `set_gclk_divider_and_source`, could you add another method to modify the standby mode? Also, please rebase instead of merging master into your branch.

TDHolmes:
@sajattack I'll give that a shot. It would be nice if I could avoid modifying the public API if possible.

TDHolmes:
@sajattack done & working on feather_m0

TDHolmes:
@sajattack let me know if you have anymore thoughts on this one

sajattack:
Lgtm

PR #293, opened 2020-10-25
Update build-all.py to load configuration from crates.json

### LINKED ISSUE #292: build all script no longer works without Travis CI file
Is the plan to change that script to using the crates.json file? If so I could help out with that if no one else has already done it

jessebraham:
Closes #292.


TDHolmes (approved):
🎉

PR #296, opened 2020-10-29
wio terminal: Implement SD card SPI initialization + example

twitchyliquid64:


boards/wio_terminal/src/storage.rs
```diff
@@ -42,3 +47,62 @@ pub struct SDCard {
...
+        //
+        // // Configure the backlight pin as a push-pull output; unfortunately this pin
+        // // does not appear to support PWM.
+        // //   HIGH - backlight enabled
+        // //   LOW  - backlight disabled
+        // let mut backlight = self.backlight.into_push_pull_output(port);
+        // backlight.set_high()?;
+
+        // Return a result consisting of a Tuple containing the display driver and
+        // backlight pin.
```
jessebraham, line 105:
```suggestion
```
I don't think this needs to be included 😉

boards/wio_terminal/Cargo.toml
```diff
@@ -49,6 +53,7 @@ default = ["atsamd-hal/samd51p", "rt", "unproven", "usb"]
 rt = ["atsamd-hal/samd51p-rt", "cortex-m-rt"]
 unproven = ["atsamd-hal/unproven"]
 usb = ["atsamd-hal/usb", "usb-device", "usbd-serial"]
+sdcard = ["embedded-sdmmc"]
```
jessebraham, line 56: Is there any reason `sdcard` shouldn't be a default feature?

jessebraham (commented):
Awesome, thanks!

It seems to be working as intended on my device. I just commented on a couple of little things then I think it should be good.

boards/wio_terminal/src/storage.rs
twitchyliquid64: Oooof good catch!

boards/wio_terminal/Cargo.toml
twitchyliquid64: No strong inclination - Ive removed that feature as its not really a biggie for it to be a dep.

jessebraham (approved):
Thank you for making the requested changes, and for the additional changes you made as well!

PR #298, opened 2020-11-01
Arduino Nano33 iot project fails to compile when using feature "usb"

fredpit:
When compiling projects with 'usb' feature, it fails because 'GenericClockController' is used twice.


sajattack:
Superceded by #299

PR #300, opened 2020-11-02
8 MHz RC oscillator clock option

TDHolmes:
Make an 8 MHz RC oscillator clock option. Currently there's no option other than 48 MHz clock fed by a 32k clock using a PLL. There is also an 8 MHz RC oscillator that can be used as the main clock, which is lower power due to the lower clock speed as well as not using the PLL hardware.

To still allow the use of the 32k clocks, I also made the 32k clock enable functions public


TDHolmes:
oof, failed the formatting check I added. Embarrassing.

jessebraham (approved):
LGTM, thanks!

PR #302, opened 2020-11-03
Replace GPIO modules with a new version

bradleyharden:
Create a new, unified GPIO module for all chips. The new module significantly reduces macro-use compared to the original version. This is part of the changes proposed in #214.

Remove the existing GPIO code and provide a compatibility shim to support existing code.


bradleyharden:
@Sympatron and @blm768, both of you have expressed interest in this. Could you have a look and see if you have any comments?

bradleyharden:
Hmm, how can I run rustfmt on my own? Looks like something failed there.

sajattack:
```sh
rustup component add rustfmt
cargo fmt
```

bradleyharden:
Got it. Thanks. Let me go through it and see if there's anything I don't like and want to reformat.

TDHolmes:
The format-all.sh script also works well to format everything quickly

bradleyharden:
I'm not sure what's wrong. The `format-all.sh` script produces no changes on my system.

bradleyharden:
I think the issue is a trailing blank line in the module-level documentation. Let's try again. I'm not sure why there's a difference between running here and on my machine.

hal/src/common/gpio/v2.rs
```diff
@@ -0,0 +1,1701 @@
...
+        pub pmux: [PMUX; 16],
+        pub pincfg: [PINCFG; 32],
+        _padding2: [u8; 32],
+    }
+
+    /// Trait used to implement zero-sized references to the GROUP registers
+    pub trait Group: Sealed {
+        const NUM: usize;
+        #[inline]
+        fn group() -> &'static GROUP {
```
blm768, line 722: I think this is technically `unsafe` if an `&mut` reference to the group exists when this method is called.

blm768 (commented):
I haven't looked over this super thoroughly yet, but I like the way the API looks.

I'm not 100% sure about the safety yet, but that should shake out after some further reviewing.

hal/src/common/gpio/v1.rs
```diff
@@ -0,0 +1,682 @@
...
+/// requires exclusive access to the Port hardware,
+/// which is why this isn't simply the standard `Into`
+/// trait.
+pub trait IntoFunction<T> {
+    /// Consume the pin and configure it to operate in
+    /// the mode T.
+    fn into_function(self, port: &mut Port) -> T;
+}
+
+/// TODO
```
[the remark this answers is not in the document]
bradleyharden: Oh, yeah, I didn't look through `v1` very thoroughly. That was just a reminder to add some documentation.

hal/src/common/gpio/v2.rs
```diff
@@ -0,0 +1,1701 @@
...
+    {
+        Self::convert(I::Group::group(), I::NUM);
+        PinT::new()
+    }
+    /// Convert a pin into `PinMode` `Self`.
+    ///
+    /// Users should not be able to call this function, because they
+    /// cannot access the `Group::group()` method in a private module.
+    #[inline]
+    fn convert(group: &GROUP, num: u8) {
```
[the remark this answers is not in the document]
bradleyharden: Yes, I think `#[doc(hidden)]` is a good idea. If my invariants hold, users couldn't even hope to call it though.

I don't think it should be marked `unsafe`. I don't believe there's any memory unsafety. This function only edits the registers and fields of the specific pin. And each update of those registers/fields is atomic, either through the `SET/CLR` registers, or using `WRCONFIG`. And we can guarantee that each pin is a singleton unless users call `unsafe` code.

FYI, there are some copy/paste bugs here. Notice that `PULLEN` is used for multiple incorrect fields. I just caught it tonight. I will push an update.

hal/src/common/gpio/v2.rs
bradleyharden: This function is intended only for consumption inside this module. The structure of this module guarantees that each pin accesses mutually exclusive registers and fields within the `GROUP`. Furthermore, each pin is a singleton, so there is only ever one reference to each set of those registers/fields. As a consequence, there is no chance for memory unsafety.

However, if used improperly, it could create problems. It sounds like I need to both document my assumptions and also mark the function as `unsafe` to communicate that the future authors.

hal/src/common/gpio/v2.rs
```diff
@@ -0,0 +1,1701 @@
...
+//==============================================================================
+
+#[inline]
+fn read_pin(group: &GROUP, num: u8) -> bool {
+    let mask: u32 = 1 << num;
+    (group.in_.read().bits() & mask) != 0
+}
+
+#[inline]
+fn write_pin(group: &GROUP, num: u8, bit: bool) {
```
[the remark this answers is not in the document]
bradleyharden: So I believe they are completely safe from a memory safety perspective. The `OUTSET/OUTCLR/OUTTGL` registers make these functions completely atomic.

There's a possibility for logic errors if you set the output in the wrong mode, but I don't think logic errors usually qualify for `unsafe`. They are also private, so only module authors need to understand their use.

hal/src/common/gpio/v2.rs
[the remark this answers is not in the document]
bradleyharden: Maybe the body of the function should be wrapped in `unsafe`, to communicate that the body must satisfy the invariants? I don't think the function itself needs to be unsafe though.

Right now, the function makes an implied contract that it will only modify the registers and fields of the specified pin. But I don't document that explicitly. Adding an internal `unsafe` block would communicate that and also allow the function to remain safe, because the contract is safe.

hal/src/common/gpio/v2.rs
blm768: Probably a good idea. If I'm understanding this code and Rust's safety rules correctly, just calling this while some other code is holding onto the mutable reference returned by `Pins::mut_port` would technically be undefined behavior, even if they didn't read from or write to that reference. Whether that would cause issues in practice is uncertain, I suppose.

hal/src/common/gpio/v2.rs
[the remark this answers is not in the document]
blm768: I guess if we're writing only to bits "owned" by the given `GROUP`, we're probably OK. This takes a shared reference, so I guess in theory, if we had multiple threads and they were both calling this method on the same `GROUP` (which I guess can't happen right now, but in theory, given no constraints aside from the method's signature, might be possible), that might cause issues.

hal/src/common/gpio/v2.rs
blm768: I guess the actual write happens atomically even though the bits are being "set" in separate calls, so the interactions with `WRCONFIG` are probably fine. The rest of this is probably fine too, except perhaps for the (very) theoretical multi-threading issue I noted in my other comment.

hal/src/common/gpio/v2.rs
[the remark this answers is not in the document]
blm768: I guess the actual write happens atomically even though the bits are being "set" in separate calls, so the interactions with `WRCONFIG` are probably fine. The rest of this is probably fine too, except perhaps for the (very) theoretical multi-threading issue I noted in my other comment.

Rather than wrapping the whole thing in an `unsafe` block, if you decide that the threading issue is something to actually worry about, just adding a "safety" section to the doc comment would probably be better than expanding the `unsafe` blocks. Keeping the size of `unsafe` blocks to a minimum is considered good practice.

hal/src/common/gpio/v2.rs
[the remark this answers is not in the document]
blm768: Yeah, it seems like at least the write functions would be memory-safe. I'm not totally sure about the reads; I guess it depends on the hardware semantics. Reading uninitialized memory is undefined behavior, but this is a volatile read of a memory-mapped register, so that rule might not apply here.

bradleyharden:
@blm768, I'm going to reply here, because my response cuts across several of your comments.

I've been thinking about your memory-safety points, and I reviewed the nomicon discussion on [aliasing](https://doc.rust-lang.org/nomicon/aliasing.html). I'd like to layout my understanding of memory safety with memory-mapped registers and indicate which parts of the code I think should be marked unsafe. If you agree, I can incorporate the highlights of our discussion into comments in the code.

If I understand correctly, Rust prevents aliasing so that it can allow the compiler to perform optimizations that would otherwise be impossible. As you say, it is unsafe to create an alias, even if you don't use it, because it violates that assumption, which could invalidate some of the compiler optimizations, creating undefined behavior. All of this makes perfect sense when the compiler has complete knowledge of the memory in question. The memory starts from a known, initialized state, and the aliasing rules allow the compiler to track every read and write of that memory.

However, as noted in the nomicon, these arguments do not apply to memory-mapped hardware. With normal memory, the compiler has complete knowledge of every memory access, so it can perform optimizations based on that knowledge. But memory-mapped hardware can change at any time, which means the compiler can't make any of the same assumptions. This is precisely why the PAC represents each register as a `VolatileCell`, using `ptr::read_volatile` and `ptr::write_volatile` under the hood.

As a consequence, many of the arguments against aliasing go out the window. In a sense, it's as if every hardware register is already aliased, because the hardware effectively has a static, mutable reference to that memory location and can change it at any time. As a consequence, the compiler can never perform optimizations, and volatile reads and writes can never be reordered.

I argue that the notion of memory safety depends strongly on the behavior of the individual register in question. For example, consider a write-only register (every read returns `0x00`) where every write is atomic. There is no reason you couldn't have multiple mutable references to such a register. There is no possible way for any write to interfere with or otherwise invalidate another write. Such a situation may be logically error-prone, but it is not memory-unsafe.

In this particular case, a memory unsafe operation would unexpectedly change or invalid the configuration of a pin. To prevent such situations, we want to create a singleton representation of each pin, so that we can guarantee each pin is only controlled at one location in the code at any given time. By doing so, we ensure that the behavior of a pin cannot be unexpectedly changed somewhere else in the code. However, to do so, we need to guarantee that each singleton can only modify its respective registers & bits. Specifically, we need to ensure that each singleton can only modify its corresponding PINCFG register, its corresponding bit in the SET/CLR/TGL registers, and its corresponding PMUXE/PMUXO field.

I implement this by creating static references to the entire GROUP register block and then writing only to the corresponding registers & bits for each pin. However, the GROUP references are not inherently restricted in any way. It is perfectly possible to modify the wrong registers and bits, changing or invalidating the state of other pins. Thus, as you note, creating a GROUP reference should be an unsafe operation. On the other hand, I'm not sure if each use of a GROUP reference should also be marked unsafe. Bugs in functions that use a GROUP reference could cause memory unsafety. But at the same time, you could counter that the act of creating a GROUP reference in the first place is the real origin of unsafety. What do you think?

hal/src/common/gpio/v2.rs
bradleyharden: I don't think the same aliasing arguments apply here, because we're using memory-mapped hardware. See my post in the main discussion thread.

hal/src/common/gpio/v2.rs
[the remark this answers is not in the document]
bradleyharden: This gets at the question I pose in the main discussion thread. I agree that creating a `&GROUP` reference should be unsafe. I'm less sure that every use of it should also be unsafe. If there were a bug in the `convert` function, it could cause memory unsafety. But the origin of that unsafety is really the creation of the `&GROUP` reference in the first place.

blm768:
It seems reasonable to treat the creation of the `GROUP` reference as the unsafe operation and consider the other operations to be "safe" for the most part. In the end, it's always safe code that has to handle the `unsafe` code correctly anyway.

hal/src/common/gpio/v2.rs
blm768: I think it's technically UB to hold both mutable and shared references to the struct, regardless of what's in it. But since (as I understand it) all of the registers can be mutated via a shared reference, it seems like there's little reason to take an `&mut GROUP` anyway.

bradleyharden: I ended up deleting the `Pins::mut_port` function anyway, because I don't think it's needed. There's no reason to mutate `PORT`, because each register is represented as a `VolatileCell` with interior mutability. If we find uses for it later, we can always add it back.

bradleyharden:
I made the changes discussed above. I also fixed a bug I discovered during hardware testing. It would be great if others could test this PR before it's merged.

I think we're getting pretty close. Does anyone else have any comments? @Sympatron?

bradleyharden:
@jessebraham, some of the new EIC pin mappings for the different SAMD21 variants were wrong. I fixed those here, along with the other fixes that were revealed in the switch to the new GPIO modules.

boards/wio_lite_mg126/src/lib.rs
```diff
@@ -83,7 +83,7 @@ define_pins!(
     pin d24 = b11,
 
     /// Digital 22: MISO
-    pin d22 = b12,
+    //pin d22 = b12,
```
sajattack, line 86: Is this something you accidentally included while testing?

bradleyharden: Good catch. This is something I forgot about. PB12 doesn't exist on D21G variants. I just looked up the schematic, and it looks like this must have been a typo. It should be PA12. I'll correct it and re-push.

Sympatron:
I just had time to skim over it to look into the `group` issue.
At first I thought it should by done like pointer handling in Rust in general, i.e creating a pointer is safe, but dereferencing is unsafe. But since these functions are hidden behind a private module I think it's fine the way it is.

Did we have a discussion on the naming yet? At the moment the pin types are called `PinT` and `PinV` and all other type-level API types don't have a suffix and the value-level API types have a `...V` suffix. I don't think this is particularly problematic and I am not very opinionated about that, but we should have a discussion about the final names in my opinion.

bradleyharden:
@Sympatron, I think I finally realized a better way to express the `GROUP` issue.

Each `&GROUP` is a reference to a bunch of `VoltatileCell`s. A `VolatileCell`, like any other `Cell` is `Send` but not `Sync`. It's completely safe to use a `&GROUP`, as long as you don't do so from multiple execution contexts. However, each `GROUP` represents 32 GPIO pins. We want finer-grained control than that. In particular, we want to create an API where each pin gets its own `PinT` struct that is both `Send` and `Sync`. To do so, we need to ensure that each `PinT` uses only atomic operations on the `GROUP` and only modifies its corresponding registers & bits of the `GROUP`.

From a _language_ perspective, I think it is only unsafe to create `&GROUP` in multiple execution contexts and modify them with non-atomic operations. However, within the context of this module, it is also unsafe to modify the registers & bits of a pin you don't own.

I think I will update the comments in the code to reflect some of this.

Separately, I'm glad you brought up naming. It was something I spent a lot of time thinking about. Did you see my section on naming in the module-level documentation? I'm not in love with the current naming scheme, but it was the best I could come up with. I'm interested to see if anyone has a better proposal.

Sympatron:
Is there any reason not to call `PinT` just `Pin` since it is the default type to use and corresponds to the previous zero-size pin types.

bradleyharden:
We could. I'm not opposed to that.

There's a natural competition between the trait and the type for the same name. It is common in other languages to suffix types with `_t`. Rust prefers `PascalCase` for types, which translates to a `T` suffix. I thought we could use that convention to break the competition, making the type `PinT`. That also makes it natural to use a `V` suffix for the value-level, type-erased pin type.

I end up using both `PinT` for concrete instances and `Pin` as a trait bound frequently enough that either could really lay claim to the "best" name of `Pin`. Concrete pins are probably a little more common though, which might argue for your position.

If you want to use `Pin` for the type, what would you suggest for the trait? And how would you rename the value-level types?

blm768:
I guess there's the option of splitting the traits, type-level structs, and value-level structs into their own submodules. That would be more verbose in code that uses two or three of them, but perhaps it would be more explicit to the reader.

bradleyharden:
I would be ok with moving the value-level API to its own module, but I don't think it makes sense to separate the traits.

Here's one idea. Right now, there is an `Pin` trait, an `OptionalPin` trait and a `SomePin` trait. `SomePin` is essentially identical to `Pin`; the only difference is that `OptionalPin` is a super trait. We could get rid of `Pin` completely and use `SomePin` instead, which would free up the name `Pin` for the type-level struct. My only complaint is that you might occasionally use `SomePin` outside of the `OptionalPin`/`SomePin` pattern, which might be confusing. But that's a pretty minor complaint. Would this be an acceptable alternative?

Sympatron:
Now that I looked into it a little deeper I think naming the trait `Pin` is fine. This is probably used a lot in generics.
Splitting the value- and type-level APIs in different modules might be worth looking into, but I haven't used it enough to have a meaningful opinion about that. If you think it makes more sense / is more ergonomic this way, I am fine with it.

blm768:
I'm honestly not sure which way is better. Either seems fine.

sajattack:
This PR adds a significant amount of code, complexity, and a fair bit of indirection (though it is internal to the module). I'm referring to the (3?) levels of indirection to set a pin high. Just trying to weigh the pros and cons. Could you elaborate a bit on what we gain from this 1700 line module that the older version didn't do?

bradleyharden:
Sure. I'll try to cover each of your points.

### Code

Yes, `gpio::v2` is 1700 lines long and over twice the length of the current module. However, I don't think it's an apples to apples comparison.

- About 300 lines of the new module is dedicated to the type-erased API, which does not exist in the current API.
- I also implement other features that don't exist in the current API. For example, the current module doesn't provide a way to put pins in any of the three disabled modes (see the datasheet). I also implement `core::convert::From` between all pin modes.
- The current module doesn't feature-gate pin definitions by chip variant as well as it could. It allows pins to exist in the code that don't exist on the chip. I have feature gates on almost every pin, which adds another 100 lines or so. This has already found a number of mistakes throughout the code base. I found a typo in one BSP and identified a few other BSPs that were using non-existent pins (albiet intentionally). It also found incorrectly feature-gated SERCOM pad definitions and EIC definitions.
- My module covers all SAMD11, SAMD21 and SAMx5x chips in one module. But to do so, it has to paper over some of the differences in pin count and PAC structure. That takes extra lines.
- I added 200 lines of module-level documentation. If you don't think that's useful, I can remove it.

I don't really see this as a bloated module. There's a lot of code, but it does a lot.

### Indirection

This is fully intentional.

The current module has a lot of code duplication, especially when reading and writing PAC registers. Every pin mode change and every pin read/write is repeated at almost every call site. If we were to ever discover a bug or want to change that code, it becomes tedious and error-prone to chase down all instances.

I aimed to consolidate every unique operation on PAC registers into a single function. There are exactly 5 functions in the module that actually write to hardware registers. I think that makes it much easier to fix bugs or make changes in the future.

Sure, that approach adds some indirection, but the current approach is essentially manually inlining. I don't see the need for that when we can let the compiler do it for us, and be sure it got it correct.

### Complexity

I would argue it's just a different kind of complexity.

Issue #214 identified the prevalence of macros as a barrier to understanding the code base. Based on some of the other references I mentioned in that thread, my aim was to reduce macro usage and also provide a more idiomatic and powerful structure.

With the current approach, every pin is a completely separate type, generated by a macro. Only the pin modes are shared. As a result, any trait implemented on all pins MUST use a macro to do so. `PinPad` is a great example.

Instead, I created a universal `Pin` type that uses type parameters for both its pin identity and its pin mode. And just like the current module, I use marker types constrained by trait bounds as type arguments. With the new approach, you can implement a trait directly on `Pin`, letting it be generic over all `PinId`s.

Personally, I think the resulting code is much more readable and more idiomatic Rust. It takes a little bit of effort to understand all of the trait bounds, but I find that far easier than wading through macros.

Furthermore, the `PinId` marker types can be used outside this module. For example, in my project, I needed to use an external interrupt. I ended up implementing an `Eic` trait on each `PinId` to describe the mapping between each pin and it's corresponding bits in the EIC registers.

Separately from the introdcution of `PinId` types, I also implemented the `Group` trait to handle pin groups in a way that removes the need to supply the `PORT` to every pin operation. Now, each pin can configure itself, which enables the implementation of `core::convert::From` between pin modes.

The new approach also makes it easy to add a type-erased API. I never looked into implementing it for the current API, but it sounds like @sympatron struggled a bit with the macros in #290.

In summary, I think you have to weigh type/trait complexity against macro complexity. In my opinion, the former is both more readable and more powerful than the latter.

### Conclusion

All of that being said, if you don't see the same value or you don't want to make breaking changes, that's fine. I can keep this stuff for my own use.

bradleyharden:
@blm768 and @Sympatron, here is what I propose to do.

I will split out the type-erased API into its own module. I will name it `gpio::erased`, and I will name the pin type `ErasedPin` instead of `PinV`. I will also remove the `V` suffix from other types, because they won't be necessary.

After thinking about it more and looking at the use cases in my current code, I think the concrete `PinT` type is more common and useful than the current `Pin` trait. I could go more in depth, but I think that's something for a different thread. I've been meaning to create an issue dedicated to some of the type-level design patterns I've run into.

I would like to rename the `PinT` type to just `Pin`. Then, I will rename the `Pin` trait to either `APin`, `IsPin` or something like it. I think `APin` is my current favorite. Do you two have any ideas or opinions?

I'll wait until I hear from @sajattack before I work on it any more.

sajattack:
Alright, good arguments.

blm768:
> I will split out the type-erased API into its own module. 

I think I like this idea. That API probably won't get used as frequently as the "main" API anyway. Perhaps `DynPin` is another potential name; it would sort of mirror the `dyn T` syntax for trait objects.

As far as the `Pin` trait goes, since it basically exists to tie a `PinId` and a `PinMode` together, maybe something like `ConfiguredPin` or `PinConfig` would be another option. Not sure I like those any better, though.

bradleyharden:
> Perhaps `DynPin` is another potential name; it would sort of mirror the `dyn T` syntax for trait objects.

I really like this idea. With only three letters, it manages to convey that type information is erased and that there is an additional run-time cost. Thanks for the suggestion.

I don't really like `ConfiguredPin`. I would take that to mean a pin configured for a particular task. `PinConfig` isn't bad, but it's a bit more verbose than `APin`. I like `APin`, because it gets at the idea that the trait is a meta-type representing all of the `Pin` types. That's what I went with for now, but I'm still open to suggestions.

bradleyharden:
I decided to name the pin trait `AnyPin`. I think that's the best I've come up with yet. I think `AnyPin`, `OptionalPin` and `SomePin` are names that fit really well together.

@sajattack, I think I'm ready to merge. Are you? Are there any more outstanding questions or concerns? Thanks @jacobrosenthal for pinging #214. @blm768 and @Sympatron, any final words? @jessebraham, did you ever find time to take a look?

jacobrosenthal:
Thanks for the good work. So far my only worry is on code size. I think it might need a look. 
```
jacob@j:~/Downloads/atsamd/boards/pygamer (gpio_pr)$ cargo size --example blinky_basic --release 
    Finished release [optimized] target(s) in 0.46s
   text    data     bss     dec     hex filename
   1896       0       4    1900     76c blinky_basic
jacob@j:~/Downloads/atsamd/boards/pygamer (gpio_pr)$ 


jacob@j:~/Downloads/atsamd/boards/pygamer (master)$ cargo size --example blinky_basic --release 
    Finished release [optimized] target(s) in 6.73s
   text    data     bss     dec     hex filename
   1884       0       4    1888     760 blinky_basic
jacob@j:~/Downloads/atsamd/boards/pygamer (master)$ 
```
12 more bytes total for the new.. Not outrageous.

add a second blink gpio
```
jacob@j:~/Downloads/atsamd/boards/pygamer (master)$ cargo size --example blinky_basic --release
    Finished release [optimized] target(s) in 0.53s
   text    data     bss     dec     hex filename
   1904       0       4    1908     774 blinky_basic

jacob@j:~/Downloads/atsamd/boards/pygamer (gpio_pr)$ cargo size --example blinky_basic --release 
    Finished release [optimized] target(s) in 3.58s
   text    data     bss     dec     hex filename
   1932       0       4    1936     790 blinky_basic
jacob@j:~/Downloads/atsamd/boards/pygamer (gpio_pr)$ 
```

cost 8bytes to add a gpio to old version
but cost 48 for the new?

jacobrosenthal:
new gpio prefers s more than the old gpio
opt 3 new is 32 vs 20, so 12 more
opt s new is 24 vs 16, so 8 more

How do other hals do?

cargo size --release
People use different release profiles so I just used this one for all
```
codegen-units = 1
lto = true
opt-level = 's' # neopixel prefers 3 over s
``` 

Per gpio cost of hals
nrf hal 20 bytes
stm32f1 24 bytes
stm32f4 48 bytes
old sam 16 bytes
new sam 24 bytes

---

I used everyones blink example, and just doubled the led ex:
```
    let mut led = pins.d13.into_open_drain_output(&mut pins.port);
    let mut led2 = pins.d12.into_open_drain_output(&mut pins.port);
    loop {
        delay.delay_ms(200u8);
        led.set_high().unwrap();
        led2.set_high().unwrap();
        delay.delay_ms(200u8);
        led.set_low().unwrap();
        led2.set_low().unwrap();
    }
```

bradleyharden:
@jacobrosenthal, I found the issue. I was preserving the drive strength when switching modes. After more thought, I decided against that. Now, every time you change pin mode, the drive strength will be reset to normal. I added functions to read and change the drive strength.

With `opt-level = 3`, here are the resutls.

```
master
$ cargo size --release --example blinky_basic
   text    data     bss     dec     hex filename
   1804       0       4    1808     710 blinky_basic
$ cargo size --release --example blinky_extra
   text    data     bss     dec     hex filename
   1828       0       4    1832     728 blinky_extra

gpio_pr
$ cargo size --release --example blinky_basic
   text    data     bss     dec     hex filename
   1720       0       4    1724     6bc blinky_basic
$ cargo size --release --example blinky_extra
   text    data     bss     dec     hex filename
   1740       0       4    1744     6d0 blinky_extra
```

The new code is smaller overall and uses only 20 bytes for a new pin, compared to 24 for the old code.

At `opt-level = "s"`:

```
master
$ cargo size --release --example blinky_basic
   text    data     bss     dec     hex filename
   1316       0       4    1320     528 blinky_basic
$ cargo size --release --example blinky_extra
   Compiling pygamer v0.5.1 (/mnt/media/Software/projects/rust/atsamd/boards/pygamer)
    Finished release [optimized] target(s) in 0.30s
   text    data     bss     dec     hex filename
   1332       0       4    1336     538 blinky_extra

gpio_pr
$ cargo size --release --example blinky_basic
   text    data     bss     dec     hex filename
   1320       0       4    1324     52c blinky_basic
$ cargo size --release --example blinky_extra
   text    data     bss     dec     hex filename
   1336       0       4    1340     53c blinky_extra
```

The old code is smaller by 4 bytes overall. The change in size for a second pin is the same.

I think we can consider the issue solved. I'll push the changes. I think we're ready to merge.

sajattack (changes requested):
Please add a doc comment here 
2020-11-18-105409_307x137_scrot

bradleyharden:
Ok. I restructured the documentation, and I added intra-doc links. Rust 1.48 lands today, which should support those links. If you build the documentation without intra-doc link support, I think it will work, but it will look bad.

TDHolmes:
🎉 excited to try this out!

PR #303, opened 2020-11-03
Pygamer Qspi

jacobrosenthal:
This seems to be working for pygamer.

Also I'm pretty sure my change to your comment  @twitchyliquid64  is correct, but would be happy to be called out on that. I cant test that high a clock speed with my non pro salae logic analyser :(

Its possibly my comments arent perfect for your chip, but im pretty sure they hold for the family? Though we should probably document somewhere that this works for the w25 family only currently.

Beyond that Im dying, I cant get the GD25Q16C on the edgebadge to set the QE bit to save my life.

### DIFF
--- a/boards/pygamer/examples/qspi.rs
+++ b/boards/pygamer/examples/qspi.rs
@@ -1,0 +1,134 @@
+//! Demonstrates reading and writing to the onboard GD25Q64C 8MiB flash.
+//! The entire chip is erased, some data written, and then read back.
+//! The Red LED blink incessantly if the data written + read back
+//! was not the same.
+//!
+//! | device  | block   | sector   | page    | unit    |
+//! |---------|---------|----------|---------|---------|
+//! | 8388608 | 64/32K  | 4096     | 256     | bytes   |
+//! | 32768   | 256/128 | 16       | -       | pages   |
+//! | 2048    | 16/8    | -        | -       | sectors |
+//! | 128/256 | -       | -        | -       | blocks  |
+//! http://www.gigadevice.com/datasheet/gd25q64c/
+//!
+//! Note:: Write Enable bit will return to reset by the following situation:
+//! -Power-Up
+//! -Write Disable (WRDI)
+//! -Write Status Register (WRSR)
+//! -Page Program (PP)
+//! -Sector Erase (SE)/ Block Erase (BE) / Chip Erase (CE)
+//! -Software reset (66H+99H)
+
+#![no_std]
+#![no_main]
+
+use panic_halt as _;
+use pygamer as hal;
+
+use hal::clock::GenericClockController;
+use hal::delay::Delay;
+use hal::entry;
+use hal::pac::{CorePeripherals, Peripherals};
+use hal::prelude::*;
+use hal::qspi::Command;
+
+#[entry]
+fn main() -> ! {
+    let mut peripherals = Peripherals::take().unwrap();
+    let core = CorePeripherals::take().unwrap();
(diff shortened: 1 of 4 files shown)


jacobrosenthal:
RE edgebadge, it should be same as your code I think because it doesnt have the writestatus2 register and should write past it
```
    flash
        .write_command(Command::WriteStatus, &[0x00, 0x02])
        .unwrap();
```

Adafruit hits this path I think
https://github.com/adafruit/Adafruit_SPIFlash/blob/12a69700ecaeed03d23ce5a64b4354495dae992a/src/Adafruit_SPIFlashBase.cpp#L166

Because of the definition
https://github.com/adafruit/Adafruit_SPIFlash/blob/3a7a0665ac1a52c7209cc4ff0fd047057c2f1109/src/flash_devices.h#L90-L100

I cant think of what to try to check..

jacobrosenthal:
Should we be rustfmt nightly?

twitchyliquid64:
> Its possibly my comments arent perfect for your chip, but im pretty sure they hold for the family?

Maybe move these more specific comments into the example code rather than the qspi crate? Given qspi can be used for anything and not just flash operating with standard commands.

> Beyond that Im dying, I cant get the GD25Q16C on the edgebadge to set the QE bit to save my life.

I cant see the command code for `WriteStatus2` in the datasheet - are you sure youre not meant to use `WriteStatus`? Aside from that, your implementation looks fine.

Did you also play around with 7.5: _7.5. Write Enable for Volatile Status Register (50H)_ ?

twitchyliquid64:
> Also I'm pretty sure my change to your comment @twitchyliquid64 is correct, but would be happy to be called out on that.

Nope, youre spot on xD

jacobrosenthal:
> Maybe move these more specific comments into the example code rather than the qspi crate? Given qspi can be used for anything and not just flash operating with standard commands.

Yeah. I think this impl is odd in that way. Its clearly hardcoded to a certain style of chip. Dummy cycles are hardcoded for instance. The impl should probably be more or less expressive in the future. 

Speaking of dummy cycles my god flash sucks. Both of the datasheets for the gigadevice chips say 06bh read take 1 dummy cycle, but adafruit uses 8 and 8 works for my GD25Q64, and 1 doesnt.. so Just version 1 million of the datasheet is just blatantly wrong i guess... sigh

> I cant see the command code for WriteStatus2 in the datasheet - are you sure youre not meant to use WriteStatus? Aside from that, your implementation looks fine.

No I'm only using WriteStatus2 for the GD25Q64. Note it did seem like WriteStatus worked fine on the 64 as well but adafruit uses writestatus2 directly. I suppose its always better to not wipe out those lower level status bits if you dont need to.

Im using WriteStatus for the GD25Q16C same as your implementation and thats what's not working. 
```
    flash
        .write_command(Command::WriteStatus, &[0x00, 0x02])
        .unwrap();
```

> Did you also play around with 7.5: 7.5. Write Enable for Volatile Status Register (50H) ?

No I hadnt, Adafruit doesnt use it.. Didnt seem to help sadly.

Either way I think ive proven out the second implementation so.. whatever. Ready for merge whenever

twitchyliquid64:
> Yeah. I think this impl is odd in that way.

Note that we are using this peripheral in memory mode, it also has a direct SPI mode.

> Either way I think ive proven out the second implementation so.. whatever. Ready for merge whenever

Awesome,thanks! Should be ready to go after a re-base to master.

twitchyliquid64 (approved):

jessebraham (approved):

Issue #78, opened 2019-07-27
Arduino/Genuino Zero Support


commonbreed:
Are there any plans for [Zero](https://store.arduino.cc/genuino-zero) support?

I see there's MKRZERO support, but I'm unsure regarding how similar the pin assignments are on these boards.

djmcgill:
The answer is unfortunately that they are not especially similar, you can compare the two:
- https://github.com/arduino/ArduinoCore-samd/blob/master/variants/mkrzero/variant.cpp
- https://github.com/arduino/ArduinoCore-samd/blob/master/variants/arduino_zero/variant.cpp

However adding the mapping between board pins and chip pins shouldn't be too bad, it would look extremely similar to the [mkrzero one](https://github.com/atsamd-rs/atsamd/blob/master/boards/arduino_mkrzero/src/lib.rs) but with the values from arduino_zero sheet above.

commonbreed:
Nice, this doesn't look too difficult to implement. When I have a minute, I'll do it - cheers. I'll leave this open until then.

CrafterSvK:
Hello, I was porting Nano 33 Iot and after one upload the USB communication stopped working. There is a space reserved so the bootloader won't be rewritten but somehow it stopped working. I will update you here just in case there is a same problem with Zero.

PR #305, opened 2020-11-03
Arduino Zero support

Sympatron:
Added support for Arduino Zero. Mostly copied from Arduino MKRZERO and Arduino Nano 33 IoT and adapted.

I don't have an Arduino Zero and therefore cannot test this. Maybe @craben20 who requested support in #78 can test this.


sajattack:
Slapped a help wanted label on here, hoping we can find someone to test this.

PR #307, opened 2020-11-06
wio_terminal: Derive Debug/Eq/PartialEq for button enums

twitchyliquid64:
Been using this functionality from another crate, and its super annoying without Debug/Eq.


jessebraham (approved):
Thanks!

PR #313, opened 2020-11-12
ADC fixes + experimental thumbv7 interrupt-driven conversions

twitchyliquid64:


jessebraham (approved):
LGTM, thank you!

hal/src/common/thumbv6m/adc.rs
```diff
@@ -33,18 +33,27 @@ impl Adc<ADC> {
 
         let mut newadc = Self { adc };
         newadc.samples(adc::avgctrl::SAMPLENUM_A::_1);
-        newadc.gain(adc::inputctrl::GAIN_A::DIV2);
+        newadc.gain(adc::inputctrl::GAIN_A::_1X);
```
TDHolmes, line 36: We had it like this because INTVCC1 is VDDANA / 2

twitchyliquid64: Ah! That explains such a strange invariant - none of the other chips have a 1/2 gain option.

I think this only exists in the first place because trying to measure the VDD voltage against a reference that isn't below the minimum operating voltage, will be non-linear when VDD approaches the reference. d51 solves this by having a configurable reference, rather than 2 references + a gain option.

Theres a lot of changes in this PR, and while I don't agree this is a nice default, I think a 1-liner to restore the default gain setting to `DIV2` would be preferable to rolling back everything.

twitchyliquid64: We talked about this in the chat. I'll PR a one-liner to restore the `DIV2` default on thumbv6.

TDHolmes:
Sounds good, thanks!!

PR #315, opened 2020-11-12
thumbv6 ADC: restore default 0.5x gain

twitchyliquid64:


TDHolmes:
Thanks @twitchyliquid64 !

PR #317, opened 2020-11-12
thumbv7em ADC: Load calibration values when initializing

twitchyliquid64:
Appears to make the reading more accurate (less diff compared to my multimeter), but I don't own any precision test equipment so who knows lol


sajattack:
Thanks again!

PR #319, opened 2020-11-14
Add feather features

ianrrees:
As discussed in atsamd-rs/community Matrix chat - this adds two features to the Feather M0 BSP to support the various RFM radio modules, and the "Express" board with an SPI flash and Neopixel.

While adding those pins, I noticed that there wasn't a pin definition for the AREF pin which is brought out on the Feather boards so added that too.


sajattack:
I thought the AREF pin was just a voltage reference, is it also a GPIO?

ianrrees:
Yep, I'm pretty sure it's usable as a GPIO.  Though, it's wired to a 1uF cap to ground so is a bit restricted in what it can be used for.

PR #320, opened 2020-11-15
wio_terminal: Enable overclocking of ILI9341 (works fine)

twitchyliquid64:
Looks like the datasheet only specs the lowest speed at the lowest voltage, but people have even managed to get it going at 80Mhz from a teensy.

This PR lets you specify the clock speed at configuration. We also conservatively overclock at 58Mhz for some of the examples where you notice the display refresh speed.

### DIFF
--- a/boards/wio_terminal/src/display.rs
+++ b/boards/wio_terminal/src/display.rs
@@ -5,6 +5,7 @@ use atsamd_hal::hal::spi;
 use atsamd_hal::prelude::*;
 use atsamd_hal::sercom::{PadPin, SPIMaster7, Sercom7Pad1, Sercom7Pad2, Sercom7Pad3};
 use atsamd_hal::target_device::{MCLK, SERCOM7};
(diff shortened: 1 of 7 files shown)


boards/wio_terminal/src/display.rs
```diff
@@ -48,20 +49,21 @@ impl Display {
...
         self,
         clocks: &mut GenericClockController,
         sercom7: SERCOM7,
         mclk: &mut MCLK,
         port: &mut Port,
+        clock_speed: F,
         delay: &mut Delay,
     ) -> Result<(LCD, Pc5<Output<PushPull>>), ()> {
         // Initialize the SPI peripherial on the configured pins, using SERCOM7 and
         // running at 20MHz.
```
sajattack, line 62: This comment should be updated or removed.

sajattack:
I run an ili9341 at 67MHz for my pi portable and I can confirm it's fine.

PR #322, opened 2020-11-18
Add enumerations for DMA trigger source to PACs

### LINKED ISSUE #321: SVDs are missing DMA trigger source enumerations
The DMAC has a control register (CHCTRLB for SAMD21 at least) where one field is the DMA trigger source, called TRIGSRC.  It seems the SVD files only have the DISABLED=0 enumeration, but there are actually several different sources.

I see there are some patch files to correct the SVD before the PAC is generated; how are the patch files constructed?

ianrrees:
This is intended to fix #321.  The SAMD11 and SAMD21 values are straight out of their datasheets.  SAMD5x changes required a little bit of interpretation of the datasheet, and I apologise for not making up some better docstrings than what are in the datasheet.

### DIFF
--- a/pac/atsamd11c/src/dmac/chctrlb.rs
+++ b/pac/atsamd11c/src/dmac/chctrlb.rs
@@ -288,6 +288,44 @@ impl<'a> LVL_W<'a> {
 pub enum TRIGSRC_A {
     #[doc = "0: Only software/event triggers"]
     DISABLE = 0,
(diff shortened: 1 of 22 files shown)


sajattack:
Looks like the D51 has some errors related to multiple names with the same value https://github.com/atsamd-rs/atsamd/pull/322/checks?check_run_id=1416440095#step:5:121

PR #323, opened 2020-11-19
GMAC support

Sympatron:
I started working on ethernet support for SAME53/54.
The basic initialization is copied from Atmel Start.

Maybe someone can have a look at the overall structure of the code. Since this is my first driver from scratch any pointers will be appreciated!

I particularly don't like [the IOSet bit](https://github.com/atsamd-rs/atsamd/compare/master...Sympatron:ethernet?expand=1#diff-679caa4423f9c36943e79cea66de8847c05c34c79996172b3c8d478edc84a5daR272-R307), because for some reason I could not get `From<(GMDC, GMDIO)>` to work for each possible combination, because of function overloading I think.

### DIFF
--- a/boards/atsame54_xpro/.cargo/config
+++ b/boards/atsame54_xpro/.cargo/config
@@ -1,0 +1,14 @@
+# vim:ft=toml:
+[target.thumbv7em-none-eabihf]
+runner = 'arm-none-eabi-gdb'
+
+[build]
+target = "thumbv7em-none-eabihf"
+rustflags = [
+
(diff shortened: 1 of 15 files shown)


bradleyharden:
@Sympatron, I will try to create a PR with my SERCOM pads module this weekend. I handle IOSET for the SERCOM pads, which is much more complicated than GMAC. You can take a look at how I did it.

This isn't up to date, but [this](https://github.com/bradleyharden/atsamd/blob/v2_modules/hal/src/common/thumbv7em/sercom/pads_v2.rs) has the core of what I did. There should only be minor modifications in my next PR.

PR #325, opened 2020-11-21
Automation: Workflows to automatically bump + release PAC crates

twitchyliquid64:
Implements generating the diff for bumping PAC versions.

If that works and you are happy with this approach, I'll add subsequent steps to commit the diff.


sajattack:
How are you planning to handle major minor patch differences?

sajattack:
Oh nevermind I'm blind.

PR #327, opened 2020-11-21
Refactor the sercom module and pads submodule

### LINKED ISSUE #214: Refactor Pin and Pad implementation

bradleyharden:
Here is a summary of the changes in this PR. I'm not sure if any of them will be controversial.

- Combine the SAMD11 and SAMD21 `sercom` modules within the `thumbv6m` module. This was straightforward, since they were almost identical. I just had to handle the feature attributes.

- Remove the `BUS_STATE_OWNED` constant in the `thumbv6m` `i2c` module, because it is never used and gives compile warnings.

- Restructure the public-facing `sercom` module to provide `v1` and `v2` sub-modules. Re-export `v1::*` by default. This is the same approach we took with the `gpio` module. Alternatively, we could instead use an inverted approach, i.e. `v1::gpio`, `v1::sercom`, `v2::gpio` and `v2::sercom`. I'm not sure which is better. What do you think? One seems to imply local refactoring, while the other hints at more global refactoring.

- All of the existing `sercom` sub-modules were placed in the `v1` module. The `v1::pads` module has been replaced by a shim that uses `v2::pads` under the hood.

- The `v2` module only contains the new `pads` module right now. I figured we could migrate other parts of the `sercom` module to `v2` as we desire.

- Right now, the `sercom` sub-modules (i.e. `uart`, `spi` and `i2c`) are private. Only their contents are publicly re-exported in the user-facing `sercom` module. In this update, I still re-export the contents of each sub-module, but I also change the sub-modules to `pub` visibility. I see this as useful in a few different ways. For instance, I think it would be helpful to view the documentation for the `spi` module without the clutter of the `uart` module. It also gives users more choice in how and what they import. Are there any objections to this?

If merged, I believe this would close #214. There is more refactoring to do, but #214 primarily addressed the GPIO pins and SERCOM pads.


sajattack:
Looks good. I'd like to hear from @jacobrosenthal about the i2c BUS_STATE_OWNED.

bradleyharden:
Pinging @blm768 and @Sympatron, because they both had feedback last time.

hal/src/common/sercom/v1/pads.rs
```diff
@@ -0,0 +1,85 @@
...
+define_pads!(Sercom2, Sercom3);
+#[cfg(any(feature = "min-samd21g", feature = "min-samd51g"))]
+define_pads!(Sercom4, Sercom5);
+#[cfg(feature = "min-samd51n")]
+define_pads!(Sercom6, Sercom7);
+
+/// The PadPin trait makes it more ergonomic to convert a pin into a Sercom pad.
+/// You should not implement this trait for yourself; only the implementations
+/// in the sercom module make sense.
+pub trait PadPin<T> {
```
blm768, line 56: Seems like it would make sense to seal this trait, although I don't know if it really matters since it's going away anyway.

blm768, line 56: Seems like it would make sense to seal this trait, although I don't know if it really matters since it's going away anyway.

hal/src/common/thumbv6m/sercom/v2/pad_map.rs
```diff
@@ -0,0 +1,300 @@
...
+//! [`PinId`]: crate::gpio::v2::PinId
+//! [`PinMode`]: crate::gpio::v2::PinMode
+use crate::gpio::v2::*;
+use crate::sercom::v2::pads::*;
+
+//==============================================================================
+//  Pad definitions
+//==============================================================================
+
+// TODO
```
blm768, line 25: Is there something left to do here?

hal/src/common/thumbv7em/sercom/v1/pad_map.rs
```diff
@@ -0,0 +1,366 @@
...
+//  Pad definitions
+//==============================================================================
+
+// This macro accepts three different patterns to handle configuration
+// attributes applied to both individual PadNums as well as entire IoSets
+macro_rules! map {
+    // Single instance, with optional attribute
+    (
+        $( #[$cfg:meta] )?
+        $Sercom:ident, $IoSet:ident, $PadNum:ident, $PinId:ident, $Cfg:ident
```
blm768, line 33: Is `IoSet` just ignored here?

blm768 (commented):
Just a few thoughts after skimming this. Looks pretty good at first glance.

hal/src/common/thumbv7em/sercom/v1/pad_map.rs
bradleyharden: Yes. If you were able to catch that, you must have been reading very closely. Thanks for the attention.

I needed to transfer the Map implementations from IOSET to PinId to support the v1 API, but I already had the entire table and macro written for IOSET. I got lazy and just copied it from the other file and updated the macro. I can clean this up a bit.

hal/src/common/thumbv6m/sercom/v2/pad_map.rs
bradleyharden: I always forget to check for TODOs.

I probably wanted to add some comments for the macro. I can do that

hal/src/common/sercom/v1/pads.rs
bradleyharden: Sure. That would be a nice addition.

hal/src/common/thumbv7em/sercom/v1/pad_map.rs
blm768: Probably not worth too much effort; I just wanted to confirm that the omission is intentional, which I figured was probably the case.

bradleyharden:
I implemented the requested changes, and I added a few new things based on recent GPIO discussions. I believe the commit messages should explain enough. If not, let me know.

I think this is just about ready to merge as well.

PR #328, opened 2020-11-21
Added support for SAM E54 Xplained Pro Evaluation Kit

Sympatron:


boards/atsame54_xpro/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+license = "MIT OR Apache-2.0"
+edition = "2018"
+repository = "https://github.com/atsamd-rs/atsamd"
+readme = "README.md"
+
+[dependencies]
+cortex-m = "0.6"
+embedded-hal = "0.2.3"
+smart-leds = "0.3"
+ws2812-spi = "0.2.0"
```
sajattack, line 18: Is there a ws2812 on this board? I don't see it used in the lib.rs.

sajattack:
I assume you have the hardware and tested this?

boards/atsame54_xpro/Cargo.toml
Sympatron: No, I will remove it.

Sympatron:
Yes, I tested it. I am unsure if `memory.x` should be different from the other BSPs because the Xplained boards have no bootloader by default.
To use this the way it is, you either have to set the FLASH origin to zero or flash a bootloader first.

boards/atsame54_xpro/memory.x
```diff
@@ -0,0 +1,33 @@
+MEMORY
+{
+  FLASH (rx) : ORIGIN = 0x00000000+0x4000, LENGTH = 0x100000-0x4000 /* First 16KB used by bootloader */
```
sajattack, line 3: To answer your question about no bootloader, please remove the +0x4000 and -0x4000 here.

Sympatron: Yeah, it's probably best not to expect a bootloader if there is none by default.

boards/atsame54_xpro/src/pins.rs
```diff
@@ -0,0 +1,444 @@
...
+
+    // Data Gateway Interface
+    pin dgi_ss = d1,
+);
+
+impl Pins {
+    /// Split the device pins into subsets
+    pub fn split(self) -> Sets {
+        let analog = Analog {
+        //     a0: self.a0,
```
sajattack, line 173: I forget if I already asked about these commented out pins. Is this still WIP?

Sympatron: The problem is that all of these pins are already there with other names. It's the same with the PCC pins. I didn't want to remove it, because I thought it might be useful to someone, but didn't know how to solve this better.
The current naming scheme is based on the board docs.

sajattack: I see. Then I guess this is good to go. Thanks again!

PR #330, opened 2020-11-24
Allow SAMD11 and SAMD21 to read DRVSTR bit in PORT

ianrrees:
Per suggestion in matrix from @bradleyharden .  The SAMD21 change agrees with newer versions of the datasheet, the newest SAMD11 datasheet is quite old and lists that bit as write-only, but a quick test on actual hardware indicates that the bit is indeed readable.


jessebraham (approved):
Great, thanks for this!

PR #331, opened 2020-11-24
Add Pin methods to AnyPin and convert between v1::Pin and v2::Pin

bradleyharden:
Although there is a one-to-one mapping between `Pin<I, M>` and `AnyPin<Id = I, Mode = M>`, the compiler doesn't know that. Provide a means to convert between the two, and implement all of `Pin`'s public methods in `AnyPin`.

@ianrrees, this is just a quick PR to show you what I mean. I still need to verify one safety concern, and I would like to flesh out the documentation more. I think the topic deserves more attention than it's getting right now.

### DIFF
--- a/hal/src/common/gpio/v2/pin.rs
+++ b/hal/src/common/gpio/v2/pin.rs
@@ -3,41 +3,42 @@
 //! This module provides a type-level API for GPIO pins. It uses the type system
 //! to track the state of pins at compile-time. To do so, it uses traits to
 //! represent meta-types and types as instances of those meta-types. For
-//! example, the trait [InputConfig] acts as a type-level `enum` of the
-//! available input configurations, and the types [Floating], [PullDown] and
-//! [PullUp] are the type-level variants of [InputConfig].
+//! example, the trait [`InputConfig`] acts as a type-level `enum` of the
+//! available input configurations, and the types [`Floating`], [`PullDown`] and
+//! [`PullUp`] are the type-level variants of [`InputConfig`].
 //!
 //! When applied as a trait bound, meta-types restrict type parameters to the
 //! corresponding variants. All of the traits in this module are closed, using
 //! the `Sealed` trait pattern, so the type-level instances found in this module
 //! are the only possible variants.
 //!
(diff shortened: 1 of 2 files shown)


bradleyharden:
Actually, I think I just figured out how to do it without any `unsafe` code.

hal/src/common/gpio/v2/pin.rs
```diff
@@ -703,18 +711,186 @@ macro_rules! pin_id {
...
+///         let mut pin: Pin<P::Id, PushPullOutput> = pin.into_push_pull_output();
+///         pin.set_high().unwrap();
+///         let pin: Pin<P::Id, FloatingInput> = pin.into_floating_input();
+///         let _bit = pin.is_low().unwrap();
+///         let pin: Pin<P::Id, P::Mode> = pin.into_mode::<P::Mode>();
+///         self.pin: P = pin.as_any::<P>();
+///         self
+///     }
+/// }
+/// ```
```
bradleyharden, line 797: @ianrrees or someone else. Could you please have a look at this documentation and make sure it all makes sense?

hal/src/common/gpio/v2/pin.rs
```diff
@@ -842,7 +1032,49 @@ where
...
+    #[inline]
+    pub fn as_any<P>(self) -> P
+    where
+        P: AnyPin<Id = I, Mode = M>,
+    {
+        // SAFETY:
+        // core::ptr::read makes a bitwise copy, regardless of whether the type
+        // implements Copy. Thus, the returned value is a copy, and we still
+        // need to drop self.
+        unsafe { core::ptr::read(&self as *const _ as *const P) }
```
bradleyharden, line 1050: Could someone please review this `unsafe` code and make sure you agree with my safety justification. For it to make sense, you'll probably need to understand the example in the `AnyPin` documentation above.

hal/src/common/gpio/v2/pin.rs
ianrrees: Looks good!  Just a few typos: inherenet, strcut, provies.

bradleyharden:
Unless anyone else has comments, I think this is ready to merge.

PR #338, opened 2020-11-25
Fix feather m4 usb

TDHolmes:
Got a feather M4 recently and wanted to use USB serial. It appears to have been neglected and wasn't working. This fixes that & adds an example so it shouldn't happen again


TDHolmes:
I stole the lib.rs changes from itsybitsy m4's implementation

sajattack:
Looks good, thanks.

PR #339, opened 2020-11-25
GH workflows: Implement HAL + BSP bump & release

twitchyliquid64:
There are two workflows:
  1. Release crates workflow - can run `cargo publish` on PACs/HAL/BSPs.
  2. Bump crate versions workflow - can patch/minor/major bump any combination of PACs/HAL/BSPs.

I also fixed an inconsistent BSP name (folder name and crate name need to match, safe to fix as that crate is not yet published).

Tested in my fork:

 - Workflow run: https://github.com/twitchyliquid64/atsamd/runs/1455301485?check_suite_focus=true
 - Generated PR: https://github.com/twitchyliquid64/atsamd/pull/7


jessebraham (approved):
Awesome, thanks!

PR #342, opened 2020-11-25
Fix publish of hal crate, correct bump workflow filename

twitchyliquid64:
- Run publish command for HAL crate using `--no-verify`, due to its dependence on chip-specific feature flags
 - Rename `bump-pac.yml` to `bump-crates.yml` to reflect that the workflow can bump any combination of the crate types


TDHolmes:
🎉

Issue #254, opened 2020-09-12
Add uart, i2c_master, spi_master and usb_allocator for all BSPs


sajattack:
Should be fairly simple to do, just time consuming, and would have prevented issues like #251

jessebraham:
If you spot any errors in the table below, or any changes have been made, please post in this issue and I will update the table.

| Board | `i2c_master` | `spi_master` | `usb_allocator` |
|:------|:-------------|:-------------|:----------------|
| arduino_mkrvidor4000 | ❌ | ❌ | ❌ |
| arduino_mkrzero | ❌ | ❌ | ✔ |
| arduino_nano33iot | ✔ | ✔ | ✔ |
| circuit_playground_express | ✔ | ❌ | ❌ |
| edgebadge | ❌ | ❌ | ❌ |
| feather_m0 | ✔ | ✔ | ✔ |
| feather_m4 | ✔ | ✔ | ✔ |
| gemma_m0 | ❌ | ❌ | ❌ |
| itsybitsy_m0 | ✔ | ✔ | ✔ |
| itsybitsy_m4 | ✔ | ✔ | ❌ |
| metro_m0 | ✔ | ✔ | ❌ |
| metro_m4 | ✔ | ✔ | ✔ |
| pfza_proto1 | ✔ | ✔ | ❌ |
| pygamer | ❌ | ❌ | ❌ |
| pyportal | ✔ | ✔ | ❌ |
| samd11_bare | ✔ | ❌ | ❌ |
| samd21_mini | ❌ | ❌ | ❌ |
| serpente | ❌ | ❌ | ❌ |
| sodaq_one | ✔ | ✔ | ❌ |
| sodaq_sara_aff | ❌ | ❌ |❌  |
| trellis_m4 | ✔ | ❌ | ❌ |
| trinket_m0 | ✔ | ❌ | ✔ |
| xiao_m0 | ✔ | ✔ | ✔ |

sajattack:
You missed uart

jessebraham:
Yes I just noticed that, thanks. Will update when I've gone through all the BSPs (again).

Sympatron:
Are you talking about the actual BSP implementation (i.e. convenience functions to initialize I2C, etc.) or are you talking about examples?

jessebraham:
This is referring to the convenience functions, you can see [an example here](https://github.com/atsamd-rs/atsamd/blob/master/boards/trinket_m0/src/lib.rs#L188-L212).

[...]

PR #344, opened 2020-11-27
Add additional BSP peripheral helper functions

jessebraham:
Just starting to pick away at #254 a bit.

- Added `spi_master` to the Trinket M0 ([reference](https://learn.adafruit.com/adafruit-trinket-m0-circuitpython-arduino/pinouts#unique-pad-capabilities-2910532-4))
- Renamed `usb_bus` to `usb_allocator` for Metro M0, added deprecation notice and compatibility shim for `usb_bus`
- Added `spi_master` and `uart` to the Wio Lite MG126 ([reference](https://wiki.seeedstudio.com/Wio-Lite-MG126/))

### DIFF
--- a/boards/metro_m0/src/lib.rs
+++ b/boards/metro_m0/src/lib.rs
@@ -256,7 +256,7 @@ pub fn usb_allocator(
 #[cfg(feature = "usb")]
 #[deprecated(
     since = "0.8.0",
-    note = "Please use the usb_allocator function instead",
+    note = "Please use the usb_allocator function instead"
 )]
 pub fn usb_bus(
     usb: pac::USB,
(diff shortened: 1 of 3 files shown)


boards/metro_m0/src/lib.rs
```diff
@@ -254,3 +252,19 @@ pub fn usb_bus(
         usb,
     ))
 }
+
+#[cfg(feature = "usb")]
+#[deprecated(
+    since = "0.8.0",
```
jessebraham, line 258: I had written this awhile back. I don't think the BSPs have been released yet, but if `0.9.0` gets published then I can update this.

boards/wio_lite_mg126/src/lib.rs
```diff
@@ -94,37 +95,19 @@ define_pins!(
     /// Digital 1: TX
     pin tx = a27,
 
+    /// USB Data MInus
```
sajattack, line 98: MInus -> Minus please

PR #346, opened 2020-11-28
make a unified RTC driver for all chip types with both timer & clock capabilities

### LINKED ISSUE #44: Implement relevant RTC modes for all build targets
I cannot figure out how to use the RTC. 

> 19.5.3 The RTC bus clock (CLK_RTC_APB) can be enabled and disabled in the Power Manager, and the default state of CLK_RTC_APB can be found in the Peripheral Clock Masking section. A generic clock (GCLK_RTC) is required to clock the RTC. This clock must be configured and enabled in

16.6.2.6 shows that CLK_RTC_APB is enabled by default, and the apbamask shows that is is on.
```rust
peripherals.PM.apbamask.read().rtc_().bit() // true
```

I tried writing RTC enable bit but it blocks forever.
```rust
let rtc = peripherals.RTC.mode0();
rtc.ctrl.write(|w| w.enable().clear_bit()); // <- this blocks forever
```

I know that I need to enable a clock generator, but I don't know how to connect a generator to the RTC peripheral.

---
Thanks for writing this crate, it helped me build [this calculator](
https://hackaday.com/2019/03/17/custom-calculator-rolls-d20-so-you-dont-have-to/?fbclid=IwAR1IYuRLkDdo2XcZw_CfjDKmrROX_7yHBsyun8kYLAFGvnyQC0oytxSEFR4). Now I want to use the RTC to seed the PRNG.

TDHolmes:
Refactors the existing M4 clock/calendar driver and M0 timer driver into a unified driver for both modes. Tested the timer mode changes on my feather M0 and feather M4 (updated examples) but haven't tested the wio_terminal clock example or the clock/calendar on M0 devices.

- [x] Test the wio_terminal clock example (can someone help with this? @jessebraham?)
- [x] Add a M0 clock/calendar example

should close #44


TDHolmes:
Is there a more concise way to write
```rs
#[cfg(any(
     feature = "samd51",
     feature = "same51",
     feature = "same53",
     feature = "same54"
 ))]
 #[cfg(any(feature = "samd11", feature = "samd21"))]
```

It makes it kind of hard to read. I wish there was a single feature I could use for either group of devices

bradleyharden:
> Is there a more concise way to write
> 
> ```rust
> #[cfg(any(
>      feature = "samd51",
>      feature = "same51",
>      feature = "same53",
>      feature = "same54"
>  ))]
>  #[cfg(any(feature = "samd11", feature = "samd21"))]
> ```
> 
> It makes it kind of hard to read. I wish there was a single feature I could use for either group of devices

@TDHolmes `min-samd51g`

TDHolmes:
thanks @bradleyharden , much nicer

jessebraham:
Looked over it quickly and things look pretty good on the surface, thanks for doing this! I have quite a bit I'd like to get done this weekend but I will do my best to do a more thorough review and test it out on the Wio Terminal.

(@twitchyliquid64 if you're able to test it before I get around to it that'd be awesome, no worries if not of course)

twitchyliquid64:
Hopefully I can test tonight. Thank you so much for writing this!

TDHolmes:
Got a feather M0 example up based on the wio_terminal one, however it's running twice as fast as it should be. I'm not sure what's causing this and I'll be curious to hear if it's happening on the other HW too.

twitchyliquid64:
Gonna try it out on the wio terminal now, one sec

twitchyliquid64:
Works fine and runs at the correct speed on the wio terminal.

TDHolmes:
thanks for checking @twitchyliquid64. Very odd. I'll look for SAMD11/21 specific reasons for this

twitchyliquid64:
Maybe theres a different base clock or a different divider default somewhere?

TDHolmes:
found it, the divider for GCLK2 is limited to 5 bits, causing overflow to 0 on the divide by 32 case. Going to add a runtime check for this so others don't hit this too!

TDHolmes:
works on feather M0 now!

TDHolmes:
with this tested on both platforms, this should be good to review

hal/src/common/rtc.rs
```diff
@@ -0,0 +1,305 @@
...
+            minutes: clock.minute().bits(),
+            hours: clock.hour().bits(),
+            day: clock.day().bits(),
+            month: clock.month().bits(),
+            year: clock.year().bits(),
+        }
+    }
+}
+
+/// RtcClock represents the RTC configured in clock/calendar mode.
```
twitchyliquid64, line 45: Update comment to reflect the new name: s/RtcClock/Rtc

hal/src/common/rtc.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// triggering the interrupt; it does not configure the interrupt
+    /// controller.
+    fn disable_interrupt(&mut self) {
+        self.mode0().intenclr.write(|w| w.cmp0().set_bit());
+    }
+}
+
+/// Helper type for computing cycles and divider given frequency
+#[derive(Debug, Clone, Copy)]
+pub struct TimerParams {
```
twitchyliquid64, line 262: Is this identical copypasta from `timer::TimerParams` ? If so, is there any way we can just use that copy instead of duplicating here?

twitchyliquid64 (commented):
LGTM!

hal/src/common/rtc.rs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    /// Common helper function that gets the best divider & calculates cycles
+    /// with that divider.
+    fn new_from_ticks(ticks: u32) -> Self {
+        let mut divider = ((ticks >> 16) + 1).next_power_of_two();
+        if divider > 1024 {
+            // lame. Can we catch this at compile time? Probably??
+            divider = 1024;
+        }
```
bradleyharden, line 296: @TDHolmes, here's one thought. In the vein of "parse don't validate", you calculate a valid divider here, but then you have to re-validate it again in the code for `start` above. I think it might make sense to define an `enum` to capture the "parsing" you do here.

hal/src/common/rtc.rs
TDHolmes: there's a few differences. The timer divider has some gaps that RTC doesn't, and timer is in 16 bit mode, so we need to check if the counter compare value fits in 16 bits whereas here we're 32 bit counter mode. I'll think about this though, there might be a good way to do this.

hal/src/common/rtc.rs
```diff
@@ -0,0 +1,305 @@
...
+        // 32 bit counter mode plz
+        self.mode0_ctrla().modify(|_, w| w.mode().count32());
+
+        // Now that we have a clock routed to the peripheral, we
+        // can ask it to perform a reset.
+        self.reset();
+
+        // the SVD erroneously marks swrst as write-only, so we
+        // need to manually read the bit here
+        while self.mode0_ctrla().read().bits() & 1 != 0 {}
```
bradleyharden, line 204: You might see if @ianrrees can help you push a change upstream to Microchip, to correct the SVD. It looks like there might be other areas in the RTC that could use changes or corrections as well.

TDHolmes: That'd be great! This seems to be an issue on the RTC & timer peripherals

hal/src/common/rtc.rs
TDHolmes: I couldn't think of a clean way to do this, however the timer version of TimerParams is copypasta'd 3 times for SAMD11, SAMD21, and thumbv7em, so that should be definitely cleaned up. I think I'll do that, but not in this PR.

hal/src/common/rtc.rs
TDHolmes: This makes sense. Trying to directly use the `PRESCALER_A` enum but it isn't quite working. Will pull out the debugger tomorrow to see what's up

hal/src/common/rtc.rs
ianrrees: Yep, I'm planning on assembling a list of fixes on Monday, will add this one.

hal/src/common/rtc.rs
TDHolmes: Figured it out. Was using the enum to divide and calculate ticks when I should have been using the divider value

TDHolmes:
All comments addressed. LMK if there are any other suggestions

hal/src/common/rtc.rs
TDHolmes: #352 unifies TimerParams for the timers

twitchyliquid64 (approved):
Thanks for this! I think this is g2g once its rebased onto master.

PR #349, opened 2020-11-28
Improve gpio::v2 in a few areas

bradleyharden:
Here are two small improvements to `gpio::v2`. They are purely internal to the module and don't affect the external API at all. The commit messages contain more details.


bradleyharden:
I just realized I was missing `repr(C)` in one case that absolutely required it. I just added it. That change is not an improvement, it is a bug fix.

bradleyharden:
I just realized I was missing `repr(C)` in one case that absolutely required it. I just added it. That change is not an improvement, it is a bug fix.

PR #351, opened 2020-11-28
Added missing TIDM::ENIDn fields to PACs

Sympatron:
The SVD files do not contain the TIDM::ENIDn fields for the GMAC peripheral.
See page 505 Chapter `24.9.24 GMAC Type ID Match n Register` in the data sheet.


jessebraham (approved):
Great, thanks!

PR #352, opened 2020-11-28
unify the identical copies of TimerParams

TDHolmes:
the SAMD11, SAMD21, and thumbv7em copies of `TimerParams` are identical. This unifies them into one common place.


twitchyliquid64 (approved):
Love it, thanks! will merge once green.

PR #353, opened 2020-11-29
Beginnings of wifi support for the Wio Terminal

twitchyliquid64:
- Methods on UART to control interrupts and read flags
 - Wifi controller implementing Seeed's eRPC protocol
 - Generic trait `erpc::RPC` to abstract implementation of specific RPCs
 - Implementations for reading the version, starting a scan, and enumerating results
 - Demo of the above implementations in the `wifi_scan` example
 - Fixing the `clock` example now that I bumped the version constraint on the `nom` crate beyond what a different dependency could handle


boards/wio_terminal/src/wifi/wifi_rpcs.rs
```diff
@@ -0,0 +1,396 @@
...
+#[derive(Debug, Copy, Clone)]
+#[repr(u32)]
+pub enum BssType {
+    Infra = 0,
+    Adhoc = 1,
+    Any = 2,
+    Unknown = core::u32::MAX,
+}
+
+const WEP_ENABLED: u32 = 1;
```
sajattack, line 91: You could use https://crates.io/crates/bitflags for this

boards/wio_terminal/src/wifi/wifi_rpcs.rs
```diff
@@ -0,0 +1,396 @@
...
+        }
+
+        let mut res = GenericArray::<ScanResult, N>::default();
+
+        //let (data, ssid) = SSID::parse(data)?;
+        for i in 0..N::to_usize() {
+            res[i] = unsafe { *((data.as_ptr() as *const ScanResult).offset(i as isize)) };
+        }
+
+        //let s: ScanResult = unsafe { core::mem::transmute::<[u8;
```
sajattack, line 318: There's some commented out code here. Did you forget or is it still a work in progress?

boards/wio_terminal/src/wifi/wifi_rpcs.rs
twitchyliquid64: Done (but in different crate now)

boards/wio_terminal/src/wifi/wifi_rpcs.rs
twitchyliquid64: Fixed (but in different crate now).

hal/src/common/thumbv7em/sercom/v1/uart.rs
```diff
@@ -171,7 +171,6 @@ macro_rules! uart {
                             w.rxpo().bits(rxpo); // Uses pad 3 for rx
                             w.txpo().bits(txpo); // Uses pad 2 for tx (and pad 3 for xck)
 
-                            w.form().bits(0x00); // No parity
```
jessebraham, line 174: Just out of curiosity, what's the reason for this line being removed?

twitchyliquid64: I think I removed it when getting the UART working. Its safe either way - the UART is reset, and `0x00` (no parity) is the default value. I can put it back if we want to minimise the diff.

jessebraham: It's fine to leave as-is if it's reset to this value anyways. As I said, just curious.

jessebraham (approved):
This looks pretty good to me, thank you so much! I'm just going to update my RTL8720 and verify the example works (which I'm sure it does), then I'll be happy to merge.

PR #356, opened 2020-12-02
Use vcell v0.1.2

david-sawatzke:
The `register.rs` file included in the output of svd2rust uses
`as_ptr()`, which was only added in vcell v0.1.1 (now yanked).

See https://github.com/rust-embedded/svd2rust/pull/484 (maybe wait until it's looked at/merged)

This was found while doing #355 with a very old existing clone & running into weird compile errors:
```
error[E0599]: no method named `as_ptr` found for struct `VolatileCell<U>` in the current scope
  --> /[..]/atsamd/pac/atsamd21e/src/generic.rs:39:23
   |
39 |         self.register.as_ptr()
   |                       ^^^^^^ method not found in `VolatileCell<U>`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: could not compile `atsamd21e`

To learn more, run the command again with --verbose.
```

### DIFF
--- a/pac/atsamd11c/Cargo.toml
+++ b/pac/atsamd11c/Cargo.toml
@@ -9,12 +9,12 @@ repository = "https://github.com/atsamd-rs/atsamd"
 readme = "README.md"
 
 [dependencies]
-bare-metal = "~0.2"
-cortex-m = "~0.6"
-vcell = "~0.1.2"
+bare-metal = "0.2"
+cortex-m = "0.6"
+vcell = "0.1.2"
 
 [dependencies.cortex-m-rt]
-version = "~0.6.12"
+version = "0.6.12"
 optional = true
 
 [features]
--- a/pac/atsamd11c/Cargo.toml
+++ b/pac/atsamd11c/Cargo.toml
@@ -11,7 +11,7 @@ readme = "README.md"
 [dependencies]
 bare-metal = "~0.2"
 cortex-m = "~0.6"
-vcell = "~0.1"
+vcell = "~0.1.2"
 
 [dependencies.cortex-m-rt]
 version = "~0.6.12"
--- a/pac/atsamd21e/Cargo.toml
+++ b/pac/atsamd21e/Cargo.toml
@@ -9,12 +9,12 @@ repository = "https://github.com/atsamd-rs/atsamd"
 readme = "README.md"
 
 [dependencies]
-bare-metal = "~0.2"
-cortex-m = "~0.6"
-vcell = "~0.1.2"
+bare-metal = "0.2"
+cortex-m = "0.6"
+vcell = "0.1.2"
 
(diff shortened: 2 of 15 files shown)


twitchyliquid64:
Did you try `cargo update` on your old clone?

david-sawatzke:
Yes & that worked. But code needing a `caro update` is a bit broken, so I'd like to fix it. Especially since it *could* lead to uncompilable code, if some dependency has  a hard dependency on v0.1.0.

twitchyliquid64:
I dont know enough about cargo / rust idioms to know if this is a bug or a feature, so ill leave this to another reviewer.

sajattack:
I'm happy with it, it just means we'll have to watch a little more closely for the next update because point releases won't go automatically.

sajattack:
Should we wait for svd2rust to update?

david-sawatzke:
Should we maybe switch to a caret requirement to avoid that? Is there any reason why they're avoided, otherwise I'd switch all the pac dependencies to a caret requirement.

I can't think of any way this is wrong, but I have been proven otherwise in the past and will be in the future. This isn't a particularly pressing issue (the offending code is over half a year old and I seem to be the first one with this issue), so I'd rather wait until the `svd2rust` folks give their verdict.

sajattack:
Sure caret seems good. I've always been a little confused by this too. What is the difference between caret, tilde, and just "0.1". Docs here but I'm still not certain https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

sajattack:
I guess the difference between caret and tilde is caret can only go upwards, but what is the unspecified one? Is it the same as tilde?

david-sawatzke:
>  Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

Unspecified is the same as caret

sajattack:
Ok, let's just do unspecified for everything then. I'm not sure why we have tilde, it's only ever caused problems in the past.

david-sawatzke:
Done, now let's wait for CI

sajattack:
Maybe should be "^0.1.2", since that one specifies all the digits and we want it to update on patch versions. Sorry I was unclear.

david-sawatzke:
Wait, I'm confused now. "0.1.2" and "^0.1.2" mean the same, there's no difference

sajattack:
Oh. I thought "^0.1.2" was the same as "0.1" but "0.1.2" meant exactly 0.1.2 with no updates.

sajattack:
Haha I should probably know this by now given how many crates I maintain.

sajattack (approved):

sajattack:
Ping me when svd2rust r+'es

david-sawatzke:
@sajattack The time has come

sajattack:
That was quick.

PR #357, opened 2020-12-03
Harmonize bsp api with an eye towards clean docs

jacobrosenthal:
Just to start a discussion around what our bsp api should look like.

Most of us are using a pattern inherited export that ends up looking like this
https://docs.rs/pygamer/0.6.0/pygamer/

Recently merged bsps clearly rethought that. Take a look at
https://docs.rs/wio_terminal/0.2.0/wio_terminal/

I presume wed like to be harmonized. And presumably wed all like our docs to look nice.

These were the (breaking) changes I made to the pygamer and edgebadge just to generally clean them up, and ended up converging to the wio terminal style. 

Though Id prefer to match whatever we all decide.


twitchyliquid64:
Sorry for the inconsistent style with the wio_terminal - it wasnt my intention to buck the trend, but I do think consistency is ideal.

I can also tackle updating some of the other BSPs if that would help? I have hardware for the itsybitsy's.

jessebraham:
Just regarding the `wio_terminal` crate, that was originally in a separate repository which I created which was later merged in here. The API differences are largely my doing, but I honestly don't have a strong preference on how to do things.

sajattack:
@jacobrosenthal @jbeaurivage Should we close this now that pygamer has been cleaned up?

jbeaurivage:
I think there is still a little bit of work to do in `edgebadge`, but it's relatively minor.

PR #359, opened 2020-12-15
Add free running ADC support to InterruptAdc

ciniml:
Hi. 

I'm currently implementing an application which uses ADC in free-run mode to acquire sensor outputs periodically.
And I've found that current `InterruptAdc` implementation does not support free-run mode at all.

So I want to add free-run mode support to `Adc` and  `InterruptAdc`.

I've added some functions to set `FREERUN` bit in the `CTRLB` register of the ADC peripheral to `Adc` struct.

I have a concern about the modification of `service_interrupt_ready` function.
I've removed `disable_interrupts`  and `power_down` calls from the `service_interrupt_ready` in order not to stop the ADC conversion after the first conversion and added `stop_conversion` function instead.
This changes the usage of `InterruptAdc` and may breaks applications which currently uses it.

Is API changes like this allowed? Or do I need to implement it as  another function?

### DIFF
--- a/boards/wio_terminal/Cargo.toml
+++ b/boards/wio_terminal/Cargo.toml
@@ -79,6 +79,9 @@ name = "clock"
 required-features = ["usb"]
 
 [[example]]
+name = "microphone"
+
+[[example]]
 name = "orientation"
 
 [[example]]
--- a/boards/wio_terminal/examples/microphone.rs
+++ b/boards/wio_terminal/examples/microphone.rs
@@ -5,14 +5,14 @@ use embedded_graphics as eg;
 use panic_halt as _;
 use wio_terminal as wio;
 
+use cortex_m::peripheral::NVIC;
+use wio::hal::adc::{FreeRunning, InterruptAdc, SingleConversion};
 use wio::hal::clock::GenericClockController;
 use wio::hal::delay::Delay;
-use wio::hal::adc::{InterruptAdc, SingleConversion, FreeRunning};
-use wio::pac::{CorePeripherals, Peripherals};
 use wio::pac::{adc0::refctrl::REFSEL_A, gclk::pchctrl::GEN_A::GCLK11, interrupt, ADC1};
+use wio::pac::{CorePeripherals, Peripherals};
 use wio::prelude::*;
 use wio::{entry, Pins, Sets};
-use cortex_m::peripheral::NVIC;
 
 use core::fmt::Write;
 use eg::fonts::{Font6x12, Text};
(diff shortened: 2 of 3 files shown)


sajattack:
The InterruptAdc is a fairly new feature, so I don't think there will be too much existing code out there to break. I'm not too familiar with the different modes myself so I've asked a couple other people to weigh in on whether freerunning mode is a good default.

sajattack:
It was suggested that another method be added for freerunning conversions to avoid breaking existing conversions.

twitchyliquid64:
In general I don't think its a good idea to change the behavior of existing types, unless the change also causes a compile error for existing code (hence forcing existing usage to consider the change).

One idea: implement a new type parameter on InterruptAdc that describes how to handle the completion of a conversion.

Something like:

```rust
/// An ADC where results are accessible via interrupt servicing.
pub struct InterruptAdc<ADC, C>
where
  C: OnCompletion,
{
    adc: Adc<ADC>,
    m: PhantomData<C>,
}

/// Describes how an interrupt-driven ADC should finalize the peripheral
/// upon the completion of a conversion.
pub trait OnCompletion {
  fn on_complete(ADC);
}

pub struct FreeRunning;

impl OnCompletion for FreeRunning {
  fn on_complete() {
    // etc
  }
}
```

The method of instantiating the `InterruptAdc` will need to change too

But this way, the calling code gets to 'pick' the behavior, and the compiler will use the code from the underlying trait with no overhead :)

ciniml:
Ok. Breaking existing API seems not to recommended according to the reply.

Now I've modified the code not to modifying existing `InterruptAdc` but adding `FreeRunAdc` , which does not stop ADC after first conversion.

The method suggested by @twitchyliquid64 seems elegant way.  But it requires changing existing `InterruptAdc` API, so it is difficult to do this.

twitchyliquid64:
Sorry for the confusion - I think breaking behavior is okay **if the API also breaks as well**. So feel free to change the API.

ciniml:
> Sorry for the confusion - I think breaking behavior is okay **if the API also breaks as well**. So feel free to change the API.

I see. I'll re-implement in the suggested way.

ciniml:
I've implemented both single conversion mode and free-running mode in the `InterruptAdc`.

I've added `ConversionMode` trait to switch the procedure on `start_conversion`, `service_interrupt_ready` and `stop_conversion`.

twitchyliquid64:
Looks great!

I'll get the CI fixed in #360, and then we should be good to make this go green :)

twitchyliquid64 (approved):
Just confirming before I merge that you've tested this?

ciniml:
I've checked both SingleConversion and FreeRunning are working on my Seeed Wio Terminal.
Should I add some example code for the Wio Terminal BSP?

twitchyliquid64:
If you have time that would be great :)

ciniml:
Ok. I'll add examples for SimpleConversion and FreeRunning to Wio Terminal BSP.

ciniml:
I've added `microphone` example to Wio Terminal BSP,  which continuously acquires sampled data of microphone via ADC1 and calculate and display min/max/avg.

twitchyliquid64:
I love it, thank you so much for this and happy holidays!

PR #362, opened 2020-12-25
swap USB pads to use gpio v2 objects

ryankurte:
Swaps USB pads to `hal::common::gpio::v2` objects﻿ to work where one is using the V2 types. This is probably going to set of a whoole chain of build failures in other modules I guess.

Not sure whether it's useful to merge this as it's breaking, I guess it could be hidden behind features if y'all wanted to support both?

### DIFF
--- a/hal/src/common/thumbv7em/usb/mod.rs
+++ b/hal/src/common/thumbv7em/usb/mod.rs
@@ -10,11 +10,14 @@ pub use self::bus::UsbBus;
 mod devicedesc;
 use self::devicedesc::Descriptors;
 
+
+use gpio::v2::*;
+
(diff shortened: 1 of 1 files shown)


twitchyliquid64:
What was the error you were getting?

IIRC theres a compatibility layer between the v2 and legacy pins somewhere. Maybe we can make `UsbBus::new` generic over the input pin types, as long as it implements `Into<Function>` or whatever?

ryankurte:
the pad type is defined directly rather than being abstract so, it's a classic `Expected A Found B` error. I couldn't find an `.into()` bound that would work between v1 and v2, it might easier to implement a UsbPadD{M,P} trait and genericise over this?

Issue #363, opened 2020-12-28
Instructions: Migration to v2 GPIO API


jaredwolff:
I've been working on converting to the new V2 GPIO API. It was confusing at first how to migrate so this may help some folks:

1. Remove `define_pins!` macro from your `lib.rs` file
2. Include this at the top of your example/main `use hal::gpio::v2::Pins;` (where `hal`is `extern crate <your board> as hal;`
3. Then access the pins as shown in the [documentation](https://docs.rs/atsamd-hal/0.11.0/atsamd_hal/common/gpio/v2/pin/index.html)
    ```rust
    let pins = Pins::new(peripherals.PORT);
    let mut red_led = pins.pa23.into_push_pull_output();
    ```

I could be wrong here so any additions would be great.

jaredwolff:
Another thing that I needed to get working is USB. I'm using an SAMD21 so `/hal/src/samd21/usb/mod.rs` needs to be tweaked:

```
//! USB Device support

#[cfg(not(feature = "unproven"))]
use crate::gpio;
#[cfg(feature = "unproven")]
use crate::gpio::v2::{AlternateG, Pin, PA23, PA24, PA25};

pub use usb_device;

mod bus;
pub use self::bus::UsbBus;

mod devicedesc;
use self::devicedesc::Descriptors;

/// Emit SOF at 1Khz on this pin when configured as function G
#[cfg(feature = "unproven")]
pub type SofPad = Pin<PA23, AlternateG>;
#[cfg(not(feature = "unproven"))]
pub type SofPad = gpio::Pa23<gpio::PfG>;

/// USB D- is connected here
#[cfg(feature = "unproven")]
pub type DmPad = Pin<PA24, AlternateG>;
#[cfg(not(feature = "unproven"))]
pub type DmPad = gpio::Pa24<gpio::PfG>;

/// USB D+ is connected here
#[cfg(feature = "unproven")]
pub type DpPad = Pin<PA25, AlternateG>;
#[cfg(not(feature = "unproven"))]
pub type DpPad = gpio::Pa25<gpio::PfG>;
```

This is merely a bandaid until all projects/examples can move to the v2 API. Similar to what's propose in #362 

In my BSP `lib.rs` I've changed the `usb_allocator` function to:

```
use hal::gpio::v2::{Floating, Input, Pin, PA24, PA25};

...

#[cfg(feature = "usb")]
pub fn usb_allocator(
    usb: pac::USB,
[...]

PR #365, opened 2020-12-30
Swap to USB pad trait for thumv7em targets

ryankurte:
Supersedes https://github.com/atsamd-rs/atsamd/pull/362, related to https://github.com/atsamd-rs/atsamd/issues/363

An alternative approach that:
- Introduces `UsbPadD{m,p}` marker traits for implementation over USB pads
- Implements `UsbPadD{m, p}` marker trait over GPIO v1 and v2 types
- Adds `<Dm, Dp>` generic arguments to `UsbBus` and other types that default to existing (v1) pad types

This allows USB pad types to be overridden with something like: 
``` rust
// Override default USB GPIO types to v2
pub type UsbBus = atsamd_hal::usb::UsbBus<Pin<gpio::PA24, Alternate<H>>, Pin<gpio::PA25, Alternate<H>>>;
```

While (hopefully) maintaining compatibility with previous users of `UsbBus` and v1 GPIOs.


twitchyliquid64 (approved):
Awesomesauce!

twitchyliquid64:
I thought you looked familiar! I was at Kawaiicon last year + distantly work with Grace's team! :grinning: 

Small world ^^

ryankurte:
hah, wild how small it is sometimes ^_^

twitchyliquid64:
(WRT the PR, LGTM but as mentioned in https://github.com/atsamd-rs/atsamd/issues/363#issuecomment-752292525 I'll wait till @bradleyharden has had a chance to look at it)

bradleyharden:
@ryankurte,

From #362:

> I couldn't find an `.into()` bound that would work between v1 and v2, i might have just missed it though?

I added an implementation of `From` in both directions between `v1` and `v2`, but it missed the cut for the v0.11 release. @twitchyliquid64 just released v0.12, which now includes it. Instead of your current approach, you could make the `new` function generic with bounds like `Dm: Into<Pin<PA24, AlternateH>>`. That should be backwards compatible and accept both versions.

If you want to stick with a new trait for other reasons, you might want to make `crate::typelevel::Sealed` a super trait. That way, users won't be able to implement your new trait themselves, even if they try.

twitchyliquid64:
The GPIO trait changes are merged in - is this ready to go @ryankurte / @bradleyharden or should we track one of the newer types?

ryankurte:
I'd be happy either way, just haven't yet had a moment to revisit / rebase to use the `.into()` approach, whatever y'all prefer?

ryankurte:
Went with the sealed approach which is a bit more complex but... also kindof nice in that you can click the trait and see the types it's implemented over more directly in the docs than looking via `.into()`.

twitchyliquid64:
Thanks! Looks good to merge once we get CI green:

```
error[E0432]: unresolved import `cortex_m_0_6::peripheral::itm`
  --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-0.5.10/src/peripheral/mod.rs:91:46
   |
91 | pub use cortex_m_0_6::peripheral::{cbp, fpb, itm, tpiu};
   |                                              ^^^ no `itm` in `peripheral`

```

twitchyliquid64:
Hmmm i may have to fix this on master

twitchyliquid64:
According to the embedded rust chat something mad derped in cortex-m, that might have been the issue

PR #371, opened 2021-01-03
Fix gpio::v2 PORT access functions and add a bsp_pins macro

### LINKED ISSUE #364: Aliasing GPIO names at the board level still available for after v2 Refactor?
One thing I did notice during my migration here: #363  is whether or not now there's a way to alias the names of the pins for specific boards. Is that still possible or with the refactor that won't work anymore?

bradleyharden:
See discussion in #363.

@jaredwolff, could you give this a try and see what you like/don't like? I'd like to get some feedback on it before merging.

Edit: closes #364

### DIFF
--- a/hal/src/common/gpio/v2/pin.rs
+++ b/hal/src/common/gpio/v2/pin.rs
@@ -1660,7 +1660,7 @@ define_pins!(
 /// The following example macro call
 ///
 /// ```rust
-/// bsp_pins!(
+/// atsamd_hal::bsp_pins!(
 ///     #[cfg(feature = "unproven")]
 ///     PA24 {
 ///         name: led_pass,
@@ -1742,7 +1742,7 @@ macro_rules! bsp_pins {
             }
         )+
     ) => {
-        paste! {
+        atsamd_hal::paste::paste! {
             pub mod bsp_pins_mod {
                 use atsamd_hal::target_device::PORT;
                 use atsamd_hal::gpio::v2::{
--- a/hal/src/common/gpio/v2/pin.rs
+++ b/hal/src/common/gpio/v2/pin.rs
@@ -1643,3 +1643,194 @@ define_pins!(
         (PD21, 21),
     }
 );
+
+/// Helper macro to give meaningful names to GPIO pins
+///
+/// The normal [`Pins`] struct names each [`Pin`] according to its [`PinId`].
+/// However, BSP authors would prefer to name each [`Pin`] according to its
+/// function. This macro defines a new `Pins` struct with custom field names
+/// for each [`Pin`], and it defines type aliases and constants to make it
+/// easier to work with the [`Pin`]s and [`DynPin`](super::DynPin)s.
+///
+/// The types and constants are defined within a public module named
+/// `bsp_pins_mod` and are then re-exported with `pub use bsp_pins_mod::*;`.
+///
+/// # Example
+///
+/// The following example macro call
+///
+/// ```rust
+/// bsp_pins!(
+///     #[cfg(feature = "unproven")]
+///     PA24 {
+///         name: led_pass,
+///         aliases: {
+///             AlternateH: LedPass,
+///             #[cfg(feature = "usb")]
+///             AlternateM: UsbPin
+///         }
+///     }
+/// );
+/// ```
+///
+/// would expand to this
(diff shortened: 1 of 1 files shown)


jaredwolff:
@bradleyharden 

Maybe I'm doing something wrong here. For some reason its complaining about the `paste` function:

```
error: cannot find macro `paste` in this scope
  --> src/lib.rs:28:1
   |
28 | / bsp_pins!(
29 | |     #[cfg(feature = "unproven")]
30 | |     PA23 { name: led_pass }
31 | | );
   | |__^
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

I'm importing the macro as: `use hal::bsp_pins;`

jaredwolff:
One thing that would be immensely helpful is the inclusion of what you're importing in your documentation code. For example:

```
use paste::paste;
use hal::bsp_pins;

bsp_pins!(
    #[cfg(feature = "unproven")]
    PA24 {
        name: led_pass,
        aliases: {
            AlternateH: LedPass,
            #[cfg(feature = "usb")]
            AlternateM: UsbPin
        }
    }
);
```

jaredwolff:
Got it working on rtic:

```rust
#![deny(unsafe_code)]
#![no_main]
#![no_std]

extern crate circuitdojo_tester as hal;
use panic_halt as _;

use hal::delay::Delay;
use hal::prelude::*;
use hal::Pins;

#[rtic::app(device = hal::pac, peripherals = true)]
const APP: () = {
    struct Resources {
        led_pass: hal::LedPass,
        delay: Delay,
    }
    #[init()]
    fn init(cx: init::Context) -> init::LateResources {
        let mut peripherals = cx.device;
            peripherals.GCLK,
            &mut peripherals.PM,
            &mut peripherals.SYSCTRL,
            &mut peripherals.NVMCTRL,
        );
        let pins = Pins::new(peripherals.PORT);
        let led_pass = pins.led_pass.into_push_pull_output();

        let delay = Delay::new(cx.core.SYST, &mut clocks);

        init::LateResources { led_pass, delay }
    }
    #[idle(resources=[led_pass, delay])]
    fn idle(cx: idle::Context) -> ! {
        loop {
            let _ = cx.resources.led_pass.toggle();
            cx.resources.delay.delay_ms(500u32);
        }
    }
};
```

The pin alias worked like a charm.

bradleyharden:
> One thing that would be immensely helpful is the inclusion of what you're importing in your documentation code.

Paste shouldn't have been required. That's a mistake on my part. I'll find a way to use the re-exported version in the HAL. I can add a line to the example to show how to import the macro.

Also, this isn't really directed at you but more at everyone. I don't like how we use "hal" as the short name for Embedded HAL, ATSAMD-HAL and some BSPs. It's *really* confusing. I think we ought to start trying to change that.

jaredwolff:
>Also, this isn't really directed at you but more at everyone. I don't like how we use "hal" as the short name for Embedded HAL, ATSAMD-HAL and some BSPs. It's really confusing. I think we ought to start trying to change that.

I'm just using it the way it was done in the examples. If you have suggestions that make it more sane I'm all ears.

bradleyharden:
I made the requested changes. If @jaredwolff is satisfied and nobody else has any feedback, I think we can merge this.

jaredwolff:
Nice. I'll play with it to make sure there aren't any new problems.

bradleyharden:
Thanks.

This should close #364.

jaredwolff:
So far so good. I had to add some compatibility to the ADC and USB drivers Otherwise no issues with importing/using v2. Thanks again @bradleyharden!

bradleyharden:
Cool. Thanks for testing.

I actually just discovered `$crate` for use in macros today. It solves a problem I had when writing the `bsp_pins` macro, so I'll probably modify it. But that should he completely transparent to users.

jaredwolff:
Oh nice! Never heard of that either. Does it affect how you were importing the crates in the macro? Curious what it will look like afterwards.

bradleyharden:
@jaredwolff, see the latest implementation of `bsp_pins` in #410 for the new macro using `$crate`. It's not much of a difference. It just means I don't have to refer to `atsamd_hal` by name.

PR #375, opened 2021-01-07
Patch SVD for SWRST register in RTC

Gentoli:
Add SVD patch to remove `access` field on SWRST for all RTC modes.


twitchyliquid64:
This affects the RTC in all series?

Gentoli:
This shouldn't change any behaviour. It just allows `bit_is_set` to be called in samd11 and samd21. samd51 has the correct SVD already.

twitchyliquid64 (approved):

PR #378, opened 2021-01-10
Implement workflow for generating docs as github pages

twitchyliquid64:
Example workflow: https://github.com/twitchyliquid64/atsamd/runs/1647258893?check_suite_focus=true
Example render: https://atsamd-bot.github.io/hal-docs/


TDHolmes:
This is awesome! There should be links to these docs in the README!

PR #380, opened 2021-01-10
Reduce CI run time

TDHolmes:
install toolchain during setup only

### DIFF
--- a/.github/workflows/build-bsp.yml
+++ b/.github/workflows/build-bsp.yml
@@ -11,6 +11,34 @@ jobs:
       uses: actions/checkout@v2
     - id: set-matrix
       uses: ./.github/actions/list-BSPs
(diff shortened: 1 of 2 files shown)


twitchyliquid64:
Setup and build are different jobs so the environment isnt carried over.

The reason we have two different jobs is because one computes the matrix, which is used to shard the build jobs. I'm not aware of a way to dynamically generate the matrix without a separate job.

TDHolmes:
@twitchyliquid64 dang.. wasn't sure so I wanted to give it a shot. Yeah I get why the separate jobs are necessary... I wonder if there is any other way to cache environment/toolchain between jobs? Will poke around...

TDHolmes:
ugh and again, I meant to open this on my personal fork. Sorry for the noise

PR #381, opened 2021-01-11
run clippy on board support packages

TDHolmes:
Runs clippy on all board support packages using the nightly compiler and `--all-features`. Ignore clippy lints from auto-generated PAC codes because it has a lot of errors.

An example of what the clippy CI output looks like can be found [on this PR on my fork](https://github.com/TDHolmes/atsamd/pull/6/checks?check_run_id=1678974210). Unfortunately, the nicely formatted annotation will not work on PRs with branches coming from forked repos due to [this issue](https://github.com/actions-rs/clippy-check/issues/2), however the CI post-merge will have the annotation, and clippy errors will still block PRs, which is good.

**TODOs**:
- [x] push fixes to two clippy errors found, USB MaybeUninit & a drop in wifi.rs
- [x] show what a PR annotation looks like
- [x] test USB fix
- [x] change USB buffer code to `singleton!(: [u8; BUFFER_SIZE] = [0u8; BUFFER_SIZE]).unwrap()`
- [ ] test wifi fix (@twitchyliquid64 will give this a shot)


twitchyliquid64:
Nice!

I can test out the removal of `drop(b)` - I remember that was necessary when I was writing it ~3 months ago, but maybe the compilers gotten smarter since then.

As for the USB buffer change - LGTM.

TDHolmes:
Thanks @twitchyliquid64! USB change looks good, so once wifi change looks good we can merge

hal/src/samd21/usb/bus.rs
```diff
@@ -152,12 +152,13 @@ impl AllEndpoints {
...
-fn buffer() -> &'static mut [u8; BUFFER_SIZE] {
-    singleton!(: [u8; BUFFER_SIZE] = unsafe{MaybeUninit::uninit().assume_init()}).unwrap()
+fn buffer() -> &'static mut [MaybeUninit<u8>; BUFFER_SIZE] {
+    singleton!(: [MaybeUninit<u8>; BUFFER_SIZE] = unsafe{ MaybeUninit::uninit().assume_init() })
+        .unwrap()
 }
 
 struct BufferAllocator {
-    buffers: &'static mut [u8; BUFFER_SIZE],
+    buffers: &'static mut [MaybeUninit<u8>; BUFFER_SIZE],
```
Sympatron, line 161: What's the point of `MaybeUninit<u8>`? `u8` is always valid. If anything, I would have expected `MaybeUninit<[u8; BUFFER_SIZE]>`. This would prevent the buffer to be zero-initialized, but I think this would be UB, because the buffer is not actually initialized, before `assume_init()` is called.

TDHolmes: Apparently what we had was UB. This is the suggested fix as far as I can tell. I could try changing it to `MaybeUninit<[u8; BUFFER_SIZE]>`. I didn't try that.

Below is the clippy lint that motivated the changes to this buffer code:
```rust
error: this call for this type may be undefined behavior
1305
   --> /home/runner/work/atsamd/atsamd/hal/src/samd21/usb/bus.rs:156:45
1306
    |
1307
156 |     singleton!(: [u8; BUFFER_SIZE] = unsafe{MaybeUninit::uninit().assume_init()}).unwrap()
1308
    |                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
1309
    |
1310
    = note: `#[deny(clippy::uninit_assumed_init)]` on by default
1311
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninit_assumed_init
```

Sympatron: But the new thing is probably just as much UB. To be safe I would write `singleton!(: [u8; BUFFER_SIZE] = [0u8; BUFFER_SIZE]).unwrap()` since the original author did not add a comment to explain why this may be safe.
The downside would only be a few microseconds at reset to zero-initialize this.

TDHolmes: I don't know enough about `MaybeUninit` to know what is best to do here, so deferring to your suggestion!  Will add a TODO item.

twitchyliquid64: This buffer is a backing buffer for usb packets and endpoint descriptors. Data is never read from it that is not previously written with known values. The reads don't extend beyond the known values. Hence the `assume_init`

twitchyliquid64: Specifically, the buffer is used to store the inbound/outbound USB frames. On the inbound side, data is only read up to the size that the USB peripheral has written: https://github.com/atsamd-rs/atsamd/blob/0fc72b3ca4c363e158350a88ff7b85f63327abde/hal/src/samd21/usb/bus.rs#L447

On the outbound side, we write the data so its implicitly initialized: https://github.com/atsamd-rs/atsamd/blob/0fc72b3ca4c363e158350a88ff7b85f63327abde/hal/src/samd21/usb/bus.rs#L338

TDHolmes: Looking at the documentation for [`assume_init`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.assume_init), it looks like we must call it only after it has been initialized. This is not the case here. I think I agree with @Sympatron , the amount of time it takes to initialize the buffer once on boot is trivial and just avoid `MaybeUninit` and some `unsafe` code all together. Any objections?

twitchyliquid64: Good catch on the semantics for `assume_init` - SGTM on initializing the buffer.

TDHolmes: updated with this change

TDHolmes:
@twitchyliquid64 will you be able to test this change out soon? 😁

twitchyliquid64:
Sorry for the delay, just tested it and it looks spiffy!

twitchyliquid64:
Is this good to merge?

TDHolmes:
No worries, thanks for checking! Yeah let’s merge

PR #383, opened 2021-01-11
fix most gpio v2 clippy warnings

TDHolmes:
**NOTE**: Based off of #381. Only view most recent commit until it is merged.

fixes most of the gpio v2 clippy warnings. There are still ones that complain about:
```rust
warning: methods called `into_*` usually take self by value; consider choosing a less ambiguous name
   --> /Users/tdh/projects/atsamd/hal/src/common/gpio/v2/dynpin.rs:362:27
    |
362 |     pub fn into_alternate(&mut self, config: DynAlternate) {
    |                           ^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
```

and
```rust
warning: methods called `as_*` usually take self by reference or self by mutable reference; consider choosing a less ambiguous name
    --> /Users/tdh/projects/atsamd/hal/src/common/gpio/v2/pin.rs:1120:22
     |
1120 |     pub fn as_any<P>(self) -> P
     |                      ^^^^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
```

as fixing those would need some design thought. @bradleyharden any ideas on these APIs/namings? Not sure if this was already discussed.


bradleyharden:
Most of the changes are fine. But I'm not sure about a few.

Personally, I don't like removing the leading zeros in the macros. To me, the symmetry of the table helps more than leading zeros hurt. What do other people think?

With respect to the "into_*" methods, is it worth sacrificing symmetry between the pin and dynpin APIs to conform to naming conventions? I would lean towards no. What about others?

I'd like to note that they did name the tool Clippy for a reason. We don't have to accept everything.

bradleyharden:
I just realized that one of the naming issues was for the as_any and as_concrete functions. Those could be renamed into_any and into_concrete.

TDHolmes:
Definitely agree we don't need to fix everything necessarily, but at least on style/formatting, I like to follow those for uniformity in style and convention, if possible and isn't too onerous. 

Agreed on the API symmetry argument, but for the rest I'd lean towards fixing. We shouldn't be looking at that table much anyways and it doesn't make it extremely unreadable or anything.

hal/src/common/gpio/v2/dynpin.rs
```diff
@@ -412,11 +418,11 @@ impl DynPin {
     }
     #[inline]
     fn _is_low(&self) -> Result<bool, Error> {
-        Ok(self._read()? == false)
+        Ok(!(self._read()?))
```
Sympatron, line 421: This doesn't really feel like an improvement...
But to get rid of the warning I suppose a clippy attributes would have to be added?

bradleyharden: I agree that it's not really an improvement. But I also don't really care enough to fight it.

bradleyharden:
> at least on style/formatting, I like to follow those for uniformity in style and convention, if possible and isn't too onerous

I think that's probably where we disagree. I would rather choose the best representation for the specific site and only default to the convention when there is no compelling reason to do otherwise. But I also don't care enough to fight it. If everyone wants to satisfy Clippy, then go for it.

TDHolmes:
> > at least on style/formatting, I like to follow those for uniformity in style and convention, if possible and isn't too onerous
> 
> I think that's probably where we disagree. I would rather choose the best representation for the specific site and only default to the convention when there is no compelling reason to do otherwise. But I also don't care enough to fight it. If everyone wants to satisfy Clippy, then go for it.

Well sounds like this is trending towards two against haha. So certainly doesn't sound like "everyone" wants it. I'm down to just close this PR.

@bradleyharden do you want to do anything about the `into_x` / `as_y` lints?

bradleyharden:
I don't think you should close it. Clippy definitely has value. It already caught that naming issue. I think a bigger question is whether it's something we want to run manually look at the output, or something we want to run in CI and require attributes for exceptions to any rule.

I'd also like to hear what @sajattack and @jessebraham think. Stuff like this is ultimately super bikeshedy, so there's probably not much hope for good consensus. Someone probably just needs to make a decision.

TDHolmes:
At a minimum #381 will still go in that will catch clippy errors, which is great for catching definitely bad stuff. What we want to do with the warnings I'm flexible on. 

Unfortunately [clippy can't annotate PRs with new warnings when the PR is coming from a fork](https://github.com/actions-rs/clippy-check/issues/2), so new warnings will be very hard to notice unless we make warnings into errors, which yeah would require either caving to every clippy demand or putting in `#[allow(..)]` statements. I'm down either way

sajattack:
Yeah I don't really care about this level of linting/warning.

TDHolmes:
I'm fine then closing this PR, but still would like to merge #381 to catch clippy errors, and then @bradleyharden if you want to change the function naming as mentioned in the summary, I think that's the main thing worth doing out of these gpis v2 clippy warnings.

bradleyharden:
I resolved the naming issue with a different approach in #385. That was a fairly big change on its own, so I didn't fix any of the other clippy warnings. If we still want to follow some of clippy's suggestions, then let's re-run it on the new version first. If not, then we can probably close this.

TDHolmes:
Awesome. Let's at least close this and maybe revisit after those changes land.

PR #385, opened 2021-01-23
Improve the implementation of AnyPin and AnyPad

bradleyharden:
The naming problem in ~~#381~~ #383 got me thinking about the `AnyPin` trait. I realized that I could significantly improve the experience of using `AnyPin` by making sure the compiler understands how to convert between `P: AnyPin` and `ConcretePin<P>` using `Into`, `AsRef` and `AsMut`. I found a way to do that, and I think it turned out to be a really elegant solution. @blm768, I think you might be interested in this, given how closely you followed the development of the original GPIO refactor.

This does have some breaking changes, but I don't think that's a problem, given how new the API is. Rather than rename the `as_concrete` and related functions, I have instead made them obsolete, so they have been removed. I also took this as an opportunity to rename `ConcretePin` to `SpecificPin`. I think the latter name is a better fit with the name `AnyPin`.

All of these changes were made to the `AnyPad` trait as well.

Below is the commit message. It gives a little more background.

### Commit message

The AnyPin and AnyPad traits act as meta-types for the respective
Pin<I, M> and Pad<S, P, M> types, and they encapsulate the Pin and Pad
types with fewer type parameters.

Although useful, it was difficult to convert between a generic type
`P: AnyPin` and the corresponding `ConcretePin<P>`. Doing so required
using custom `as_concrete` and related functions. To ease this burden,
we repeated the entire `Pin` API on the `AnyPin` trait. However, that
places an additional burden on the maintainer, because any changes to
the API or documentation must be synchronized. Instead, it would be
preferrable to simplify conversion between the abstract and concrete
types, and then require users to perform that conversion manually.

Create an `Is` trait that represents type identity. The super traits of
`Is` enforce reflexive implementations of `From`, `AsRef` and `AsMut`
for `Is::Type`, meaning `T: From<T> + AsRef<T> + AsMut<T>` is always
true when `T: Is<Type = T>`. Implementations of `Is` must always have
`Is::Type == Self`.

Use the new `Is` trait to simplify converting between abstract and
generic types. Change the names `ConcretePin` and `ConcretePad` to
`SpecificPin` and `SpecificPad` to better match the names `AnyPin` and
`AnyPad`. Apply the trait bound `AnyPin: Is<Type = SpecificPin<Self>>`
and similarly for `AnyPad`. Doing so allows trivial conversion between
`P: AnyPin` and `SpecificPin<P>` using `Into`, `AsRef` and `AsMut`.

Update the documentation of the `pin` and `pads` modules to explain how
to use `Into`, `AsRef` and `AsMut` for the conversion. Also make minor
documentation updates in other areas as well.


hal/src/common/gpio/v2.rs
```diff
@@ -1,19 +1,19 @@
 //! # Version 2 of the GPIO module
 //!
-//! The new API provides two different submodules, [pin] and [dynpin],
-//! representing two different ways to handle GPIO pins. The default, [pin], is
-//! a type-level API that tracks the state of each pin at compile-time. The
-//! alternative, [dynpin] is a type-erased, value-level API that tracks the
+//! The new API provides two different submodules, [`pin`] and [`dynpin`],
```
twitchyliquid64, line 3: 'New ' is relative, maybe call this 'the GPIO API' ?

hal/src/common/gpio/v2/dynpin.rs
```diff
@@ -6,41 +6,40 @@
 //! value-level API solves this problem by erasing the type information and
 //! tracking the pin at run-time.
 //!
-//! Value-level pins are represented by the [DynPin] type. [DynPin] has two
-//! fields, `id` and `mode` with types [DynPinId] and [DynPinMode]
+//! Value-level pins are represented by the [`DynPin`] type. [`DynPin`] has two
```
twitchyliquid64, line 9: The term 'value-level' is unfamilar to me - is this the same as 'dynamically typed'?

bradleyharden: It's unfamiliar because I made it up. I thought it would help make things clear, but maybe I just made it worse? I was trying to contrast the `pin` module, which uses a "type-level" API, with the `dynpin` module, which uses a "value-level" API. I think maybe I explain the terms a bit more in the `v2` module docs.

At the value-level, a `DynPinMode` is a standard `enum`, just like you would normally write in Rust. Each different pin mode is represented as a different variant of the `DynPinMode` `enum`. But at the type-level, each different pin mode is its own type, and they are effectively variants of the `PinMode` meta-type (or type class), which is represented as a trait.

I couldn't find a way to summarize that distinction, so I made up the terms "type-level" and "value-level". But if it ends up adding rather than removing confusion, maybe I should find a different way to express it. What do you think?

I suppose you could call the "value-level" API a dynamically typed version of the "type-level" API. But that only makes sense if you expect pin modes to be represented as types to begin with. Outside of an embedded context, I don't think anyone would bother with the "type-level" approach. The "value-level" approach is much more idiomatic Rust.

hal/src/common/gpio/v2.rs
bradleyharden: How about

> The `v2` API provides two different submodules...

Does that work?

PR #390, opened 2021-01-25
Fixup more clippy suggestions

TDHolmes:
Did all of the ones that looked reasonable to me. Tested the `usb_echo` example on feather_m0 and feather_m4 for USB regressions as well as the `sleeping_timer_rtc` example for the time conversion to `From<>`.


TDHolmes:
Any love for this PR? 🤗

PR #391, opened 2021-01-25
Dmac

jbeaurivage:
# Direct Memory Access Controller

This module provides a type-safe API with compile-time guarantees that the peripheral and individual DMA channels are correctly configured before launching a DMA transfer. This is a pure Rust driver and is not derived from any existing C/C++ library.

This module currently supports most basic DMA functions, including memory-to-memory, memory-to-peripheral, peripheral-to-memory, and peripheral-to-peripheral transfers. One-shot and circular transfers are supported. More complex transfer configurations, including multi-buffer (linked-list descriptor), are not currently supported.

Transfers are supported for `i8`, `u8`, `i16`, `u16`, `i32`, `u32` and `f32` beat sizes.

# Enabling DMA support

The `dma` feature must be enabled in the board support crate or final executable.

`Cargo.toml`:

``` toml
[features]
dma = ["atsamd-hal/dma"]
```

# Channels and RAM

Using DMA channels require a certain amount of RAM - 32 bytes per channel, to be exact. RAM will be not allocated unless the `dma` feature is enabled for the HAL. By default, half the channels available on the chip are enabled. If you need all DMA channels enabled, enable the `max-channels` feature in your board support crate or final executable.

`Cargo.toml`

``` toml
[features]
dma = ["atsamd-hal/dma"]
max-channels = ["dma", "atsamd-hal/max-channels"]
```

RAM usage per chip family:
* `ATSAMD11` - 3 channels (default): 96 bytes
* `ATSAMD11` - 6 channels (max): 192 bytes
* `ATSAMD21` - 6 channels (default): 192 bytes
* `ATSAMD21`: - 12 channels (max): 384 bytes
* `ATSAMD51/ATSAME5x`: - 16 channels (default): 512 bytes
* `ATSAMD51/ATSAME5x`: - 32 channels (max): 1024 bytes

Individual channels are available through a split() method on the DMAC. They can be reused after a transfer is complete.

# Priority levels and Arbitration

The DMAC features 4 priority levels. Level 0 has the highest priority and level 3 has the lowest. Each channel can be assigned to one priority level. If two channels with the same priority level are requested to execute a transfer at the same time, the lowest channel number will have priority (in the default arbitration scheme, eg. not round-robin). By default, all priority levels are enabled when initializing the DMAC (see `DmaController::init`). Levels can be enabled or disabled through the various `level_x_enabled` methods.

Round-Robin Arbitration can be enabled for individual priority levels by using the various `level_x_round_robin` methods. By default, all priority levels are initialized with a fixed arbitration scheme. See ATSAMD21 datasheet section 19.6.2.4 for more information.

# Interrupts

This driver does not use or manage interrupts issued by the DMAC. Individual channels can be configured to generate interrupts when the transfer is complete, an error is detected or the channel is suspended. However, these interrupts will not be triggered unless the DMAC interrupt is unmasked in the NVIC. You will be responsible for clearing the interrupt flags in the ISR.

# About static lifetimes

The safe API this driver offers requires all buffers (source and destination) to have `'static` lifetimes. This is because `mem::forget` is a safe API, and therefore relying on `mem::drop` to terminate or abort a transfer does not guarantee the transfer will be terminated (specifically if `mem::forget` is called on a `DmaTransfer` containaing a `Channel<Busy, ID>`). This could cause the compiler to reclaim stack-allocated buffers for reuse while the DMAC is still writing to/reading from them! Needless to say that is very unsafe. Refer [here](https://docs.rust-embedded.org/embedonomicon/dma.html#memforget) or [here](https://blog.japaric.io/safe-dma/) for more information. You may choose to forego the `'static` lifetimes by using the unsafe API and the various `x_src_x_dest_unchecked` methods.

# Unsafe API

This driver also offers an `unsafe` API through the `x_src_x_dest_unchecked` methods. These do not enforce `'static` lifetimes, and allow using `*mut T` pointers as buffers. If you choose to use these methods, you MUST prove that a `DmaTransfer` containing a `Channel<Busy, ID>` will NEVER be dropped. You *must* call `wait()` or `stop()` manually on every `UnsafeTransfer` to convert their underlying channels back into `Channel<Ready, ID>`. No destructor or `Drop` implementation is offered for `DmaTransfers` or `UnsafeTransfer`s.

# Example
``` rust
let mut dmac = DmaController::init(peripherals.DMAC, &mut peripherals.PM);
// Get individual handles to DMA channels
let channels = dmac.split();
// Initialize DMA Channel 0
let chan0 = channels.0.init(PriorityLevel::LVL0, false, &mut dmac);
// Setup a DMA transfer (memory-to-memory -> incrementing source, incrementing destination)
// NOTE: buf_src and buf_dest should be either:
// &'static mut T, &'static mut [T], or &'static mut [T; N] where T: BeatSize
let xfer = DmaTransfer::inc_src_inc_dest(chan0, buf_src, buf_dest, false, ());
// Begin transfer with a software trigger
let xfer = xfer.begin(&mut dmac, TriggerSource::DISABLE, TriggerAction::BLOCK);
// Wait for transfer to complete and grab resulting buffers
let (buf_src, buf_dest, chan0, _) = xfer.wait(&mut dmac);
```

# Const generics and rustc version

This driver uses const generics pretty heavily, and therefore compiles only using rustc 1.51 and up. **I think this PR should only be merged into master when rustc 1.51 gets stabilized in March**.

# Improvements

* There is more testing to be done, so any help would be appreciated, especially on the Cortex-M4 chips as I don't own any SAMD51/SAMEx hardware. Also untested on SAMD11, but the DMAC is practically identical to the SAMD21 so I'm fairly confident there should be no problems there.
* No efforts have currently been made to support error handling on DMA transfers.
* Transfer complete detection: currently the method to detect if a transfer is complete is to check that both BUSY and PEND bits of the CHSTATUS are cleared. It does not check the TCMPL bit of CHINTFLAG. Doing so would be a little tricky, as it would interfere with this driver's intent of not touching interrupt handlers or flags. However it might be necessary (?) to ensure it works in all cases?


twitchyliquid64:
This is awesome!!!

The CI failure looks transient - I'm re-running now.

hal/src/common/mod.rs
```diff
@@ -9,6 +9,10 @@ pub mod time;
 pub mod timer_params;
 pub mod timer_traits;
 
+#[cfg(feature = "unproven")]
```
twitchyliquid64, line 12: Maybe an `#[cfg(all(feature = "dma", feature = "unproven"))]` instead of two cfg directives?

jbeaurivage:
@twitchyliquid64 thanks! I would love some additional testing, especially for the Cortex-M4 chips as I don't own any. Maybe we could figure out a way to integrate DMA transfers with peripherals for some sweet DMA driven UART/I2C/SPI/ADC operation. I'm not sure if those are better suited to be written in their respective modules or directly in the `dmac` module?

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,864 @@
...
+        trig_src: TriggerSource,
+        trig_act: TriggerAction,
+    ) -> UnsafeTransfer<B, P, Src, Dst, Busy, ID> {
+        // Compiler fence to prevent the compiler from re-ordering read/write operations
+        // beyond this fence.
+        // (see https://docs.rust-embedded.org/embedonomicon/dma.html#compiler-misoptimizations)
+        // (see https://docs.rust-embedded.org/embedonomicon/dma.html#compiler-misoptimizations)
+        // TODO change this to an atomic::fence since SAMD5x/SAME5x feature a
+        // data/instruction cache?
+        atomic::compiler_fence(atomic::Ordering::Release); //  ▲
```
jbeaurivage, line 587: Is an `atomic::compiler_fence` sufficient here? [This](https://docs.rust-embedded.org/embedonomicon/dma.html#compiler-misoptimizations) seems to suggest so. Although since the SAMD51/SAME5x have a data/instruction cache, I'm not so sure if an `atomic::fence` would be more appropriate.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,864 @@
...
+    /// resources
+    pub fn wait(self, dmac: &mut DmaController) -> (Src, Dst, Channel<Ready, ID>, P) {
+        let chan = self.chan.release(dmac.as_mut());
+
+        // Compiler fence to prevent the compiler from re-ordering read/write operations
+        // in front this fence.
+        // (see https://docs.rust-embedded.org/embedonomicon/dma.html#compiler-misoptimizations)
+        // TODO change this to an atomic::fence since SAMD5x/SAME5x feature a
+        // data/instruction cache?
+        atomic::compiler_fence(atomic::Ordering::Acquire); // ▼
```
jbeaurivage, line 644: Same for here

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,864 @@
...
+    /// resources
+    pub fn stop(self, dmac: &mut DmaController) -> (Src, Dst, Channel<Ready, ID>, P) {
+        let chan = self.chan.stop(dmac.as_mut());
+
+        // Compiler fence to prevent the compiler from re-ordering read/write operations
+        // in front this fence.
+        // (see https://docs.rust-embedded.org/embedonomicon/dma.html#compiler-misoptimizations)
+        // TODO change this to an atomic::fence since SAMD5x/SAME5x feature a
+        // data/instruction cache?
+        atomic::compiler_fence(atomic::Ordering::Acquire); // ▼
```
jbeaurivage, line 659: And here

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// BUSYCH is set when the channel is ACTIVELY transferring;
+    /// PENDCH is set when a trigger request has been received
+    /// but the transfer hasn't been started yet.
+    /// Therefore, when a trigger request is issued, PENDCH will be set first,
+    /// then when the arbiter begins to service the channel, PENDCH is cleared
+    /// and BUSYCH is set. To make sure the transfer is actually complete, the
+    /// channel needs to be both NOT PENDING and NOT BUSY.
+    #[inline(always)]
+    pub(crate) fn xfer_complete(&self, dmac: &mut DMAC) -> bool {
+        dmac.busych.read().bits() & (1 << ID) == 0 && dmac.pendch.read().bits() & (1 << ID) == 0
```
jbeaurivage, line 299: Does checking that `BUSYCH` and `PENDCH` are cleared guarantee that the transfer is complete?

jbeaurivage (commented):
Some things I was unsure of during development and I'd like clarifying

TDHolmes:
I have a feather_m4. I could help out testing. Also, getting a peripheral-to-mem example would be great. I'm sure most people would be interested in that usage of DMA. Thanks for working on this!

hal/src/common/dmac/transfer.rs
TDHolmes: Considering the following comment:
> In the case of Cortex-M7 cores you'll need memory barriers (DMB/DSB) if you are using the data cache (DCache), unless you manually invalidate the buffer used by the DMA. Even with the data cache disabled, memory barriers might still be required to avoid reordering in the store buffer.

I feel like it would be appropriate to use `atomic::fence` for those cores. Probably fine to just change to `atomic::fence` for everyone, or you could use one of the "-min" features to choose

jbeaurivage:
> I have a feather_m4. I could help out testing. Also, getting a peripheral-to-mem example would be great. I'm sure most people would be interested in that usage of DMA. Thanks for working on this!

@TDHolmes getting a peripheral -> memory (or the inverse) is easy, the problem is the peripheral (eg. an `I2CMaster`) drivers take ownership of the register blocks so accessing them safely isn't possible oustide the driver. I think doing so safely will require adding DMA support to every peripheral driver we wish to support, or even making entirely new drivers (eg `I2CMasterDma`)? I'm not entirely too sure how to go about this.

adamgreig:
That CI failure due to cortex-m 0.6.6 should be fixed now.

twitchyliquid64:
@jbeaurivage if you rebase on master the CI should run (and hopefully pass). Thanks!

TDHolmes:
@jbeaurivage ah, yeah makes sense. Perhaps that is out of scope of this initial PR.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,857 @@
...
+/// Transfer configuration abstraction. This is also used
+/// for `UnsafeTransfer`s so no `'static` bound.
+struct TransferConfiguration<B, Src, Dst>
+where
+    B: DmaBeat,
+    Src: DmaBuffer<B>,
+    Dst: DmaBuffer<B>,
+{
+    source: Src,
+    dest: Dst,
```
jbeaurivage, line 136: Should `source` and `dest` be changed to `Pin<DmaBuffer<B>>`? [This](https://docs.rust-embedded.org/embedonomicon/dma.html#immovable-buffers) uses `Pin`ning but [this](https://blog.japaric.io/safe-dma/) doesn't. Can we just assume that a `DmaTransfer where B: 'static` is safe?

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,857 @@
...
+    length: u16,
+    _t: PhantomData<B>,
+}
```
jbeaurivage, line 139: Should `source` and `dest` be changed to `Pin<DmaBuffer<B>>`? [This](https://docs.rust-embedded.org/embedonomicon/dma.html#immovable-buffers) uses `Pin`ning but [this](https://blog.japaric.io/safe-dma/) doesn't. Can we just assume that a `DmaTransfer where B: 'static` is safe?

bradleyharden:
@jbeaurivage, how does this compare with the samd-dma crate? We had talked about porting that over and integrating it here. I'm using it successfully right now with a linked list of descriptors.

As a broader question, if we accept this PR, does that mean we should drop attempts to port samd-dma? Does this PR have a better structure that will give a better long term solution?

jbeaurivage:
@bradleyharden I think both libraries were developed with very different philosophies. To me, the samd-dma crate feels more like a thin wrapper around the DMAC registers. The intent behind this PR, from the ground up, was to use (and perhaps abuse) the compiler checks to make using the safe API as bulletproof as possible in terms of memory safety, since it is very easy to introduce UB by misusing DMA transfers. The choice was made to prioritize safety and correctness over trying to support every functionality offered by the DMAC.

A lot of that revolves around typestates and ensuring every moving part is correctly configured before being available for use (eg. you can't build a `DmaTransfer` with a `Channel<Uninitialized, ID>`. It needs to be initialized first). 

One gripe I had with the samd-dma crate is that a lot of the functions that are marked as safe, really aren't. Here's a quick example showing it's pretty easy to introduce undefined behaviour using safe functions.:
``` rust
let a: u8 = 0xff;
let b: u32 = 0x0;
let descriptor = TransferDescriptor::new();

// These next three lines aren't undefined behaviour, but could lead to unexpected results,
// especially on little-endian platforms:
descriptor.set_src_addr(&a as *const ());
descriptor.set_dst_addr(&b as *const ());
// Trigger transfer

// These next three lines are undefined behaviour. Writing an u32 into an u8 will definitely overflow:
descriptor.set_src_addr(&b as *const ());
descriptor.set_dst_addr(&a as *const ());
// Trigger transfer
```

In contrast, this PR makes it statically impossible to introduce UB by using mismatched types, or by trying to transfer from/to arrays with mismatched lengths:
``` rust
// Assume chan was initialized
// chan: Channel<Ready, 0>
// Assume a and be were statically initialized
// a: &'static [u8; 15]
// b: &'static[u8; 15]
// This compiles,  as a and b point to arrays of the same type and length.
let transfer = DmaTransfer::inc_src_inc_dest(chan, a, b, false, ());
// Now assume a and b are of different lengths
// a: &'static [u8; 12]
// b: &'static [u8; 15]
// This will NOT compile
let transfer = DmaTransfer::inc_src_inc_dest(chan, a, b, false, ());
// Now assume a and b are of same length but different type
// a: &'static [u32; 15]
// b: &'static [u8; 15]
// This won't  compile either
let transfer = DmaTransfer::inc_src_inc_dest(chan, a, b, false, ());
```
These restrictions could be relaxed in time, as const generics become more featureful. We could do things like transfer from a source array that is shorter than its destination array, as it is not undefined behaviour. I also think that linked transfers could eventually become possible, perhaps by wrapping all involved `DmaTransfer`s into a `LinkedTransfer` object or something, although I think much more auditing would be necessary to make sure everything is safe.

`DmaTransfer`s also have provisions to take ownership of a generic payload. This is to support DMA implementations for different peripherals. For instance, to prevent data races, the `DmaTransfer` could take ownership of an `I2CMaster` until the transfer is complete.

Ultimately, I think both solutions try to serve different purposes. In my opinion, samd-dma could be a viable option *if none of its functionalities are exposed to the user of atsamd-hal*, and it's only used as an internal implementation for other peripherals. It's just too easy to misuse and inadvertently introduce UB using safe APIs. This PR tries to offer both a safe API and an unsafe but more flexible one, which I think could be used for both the HAL to implement DMA support for peripherals, and the end users of the HAL to implement their own transfers.

bradleyharden:
@jbeaurivage, thanks for the response. I only skimmed it, but your first paragraph told me you were going exactly where I thought you were going.

I'd like to review your comment and the code in depth. I'm just short on time at the moment. Maybe I'll find some time this weekend. Looking forward to it.

bradleyharden:
Also, I like the idea of focusing on the transfer types that can be made safe with the compiler, even if you can't do everything. There's still value in that.

But maybe alsp think if there might be a way to handle ping/pong buffers with your approach. I feel like that's a pretty common use case. Maybe there's a way to special case it, without solving linked lists generally. Or maybe you've already thought about it. Just an idea

jbeaurivage:
@bradleyharden looking forward to your review. I'm definitely open to suggestions. I agree there's probably a way to figure out linked buffers. I just wanted to focus on the simpler use cases first, and get working examples out for people to start testing.

bradleyharden:
@jbeaurivage, first I want to say I really like where you're going with this. That makes me even more disappointed it can't handle my current, linked-list use case.

I have some general comments that I will give here, and I also have some specific comments that I'll markup in the code. But before I do that, I'd like to note one thing. For the sake of brevity, I'm going to state my conclusions and opinions as fact, but I only spent 30-60 minutes reviewing, so there's a good chance I may have missed something or misunderstood what you're doing. Please correct me if that's the case.

### General

- I think your priority levels are inverted. You say

> The DMAC features 4 priority levels. Level 0 has the highest priority and level 3 has the lowest.

But the datasheet says:

> As long as all priority levels are enabled, a channel with a higher priority level number will have priority over a channel with a lower priority level number.

- I think you could use the feature `min-samd51g` in places where you are currently using `any()` with four different `samx5x` features. That would be simpler and less cluttered.

- I'm not sure if `#[inline(always)]` is the best choice. I remember moving away from it, to the standard `#[inline]`, but I can't remember why. You might want to look into that. Also, you might want to go through and double check your inlining decisions. I think I saw a few cases where a one- or two- line function wasn't marked as inline.

- I think you might be confused about the usage of `Sealed`. Implementing `Sealed` does nothing on its own. It's only when you use `Sealed` as a trait bound that it has any real effect. I see you implemented `Sealed` in a few places, but I couldn't find anywhere that uses it as a trait bound.

The usual pattern is to use `Sealed` as a super trait. Because `Sealed` is in a private module, users can never implement `Sealed` themselves, so they can never satisfy the requirements for your trait.

- In several places, you use `release` as a function name to consume and deconstruct an object. I think I've usually seen such functions named `free`.  I don't really care either way, but I think we should decide on a consensus name and stick with that. Maybe others can chime in on this.

- In my work on the `gpio`, `sercom` and soon-to-be `spi` modules, I came up with what I call the `Any*` trait pattern. I think it would be useful for you here. It helps tremendously to cut down on generics.

The [`AnyPin`](https://atsamd-rs.github.io/docs/samd51n/thumbv7em-none-eabihf/doc/atsamd_hal/common/gpio/v2/pin/trait.AnyPin.html) documentation explains the concept well, but it doesn't make much use of it directly ([implementation here](https://github.com/atsamd-rs/atsamd/blob/4816bb13a12a604e51f929d17b286071a0082c82/hal/src/common/gpio/v2/pin.rs#L727)).

On the other hand, my new SPI module makes heavy use of the pattern. I haven't made a PR yet, but the `thumbv7em` version is essentially done. I'm only missing module-level docs. Here is a [link](https://github.com/bradleyharden/atsamd/blob/spi_pr/hal/src/common/thumbv7em/sercom/v2/spi.rs) for reference, if you feel like looking through it. The `AnyPads` and `AnyConfig` traits are the most relevant.

As a specific example, your `UnsafeTransfer` struct takes 6 generics. Instead of storing a `Channel` and requiring both `ChanStatus` and `ID` generics, you could instead take a single type that satisfies a new `AnyChannel` trait. The trait would have an associated type and constant to store the `ChanStatus` and `ID` respectively. You could also create an `AnyTransferConfig` trait to consolidate `B`, `Src` and `Dst`. That would reduce the number of generics on `UnsafeTransfer` from 6 to 3. I also think the pattern also helps to encapsulate behavior well.

### Transfers

- Right now, you have four different transfer types and individual functions for each. But each of those functions is just a thin wrapper. Instead, I would propose consolidating all four transfer types into a single, generic function. I'm pretty certain you could implement any difference in behavior between `DmaBuffer` types using associated constants or functions in the trait.

The [`convert`](https://github.com/atsamd-rs/atsamd/blob/4816bb13a12a604e51f929d17b286071a0082c82/hal/src/common/gpio/v2/pin.rs#L425) function in the `PinMode` trait is a really good example of this pattern. I was able to collapse all of the different pin mode conversion functions into a single implementation of a trait function.

- Instead of implementing `DmaBuffer` for `[T]` and `[T; N]` specifically, you might consider a blanket implementation for anything that implements either `AsMut<[T]>` or `DerefMut<[T]>`. That would allow using `heapless::Vec`, user-defined types, etc.

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// [`DmaTransfer`](super::transfer::DmaTransfer).
+    ///
+    /// # Return
+    ///
+    /// A `Channel` with a `Ready` status
+    pub fn init(
+        mut self,
+        lvl: PriorityLevel,
+        enable_interrupts: bool,
+        controller: &mut DmaController,
```
bradleyharden, line 123: Somewhere in the past year, I came across an article that gave a good rationale for the position of various arguments in a function call. I forget where I found it. Maybe it was on lobste.rs? Anyway, one of the takeaways was that arguments that provide context should go first. I think the `controller` argument is exactly that. It's the context you need to make use of the `lvl` and `enable_interrupts` arguments. I would move it up to the first argument (other than `self`).

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,313 @@
...
+                feature = "same51",
+                feature = "same53",
+                feature = "same54"
+            ))]
+            d.chprilvl.modify(|_, w| w.prilvl().bits(lvl as u8));
+
+            if enable_interrupts {
+                // Enable all interrupt sources
+                d.chintenset
+                    .modify(|_, w| w.susp().set_bit().tcmpl().set_bit().terr().set_bit());
```
bradleyharden, line 150: I would avoid chaining writes like this. `rustfmt` does terrible things, and this is really unreadable. I would stick with multiple calls of `w.field().change();`. You'll just need to leave the semi-colon off the last one.

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,313 @@
...
+        controller: &mut DmaController,
+    ) -> Channel<Ready, ID> {
+        let dmac = controller.as_mut();
+
+        // Software reset the channel for good measure
+        self._reset_private(dmac);
+
+        self.with_chid(dmac, |d| {
+            // Reset the channel to its startup state and wait for reset to complete
+            d.chctrla.modify(|_, w| w.swrst().set_bit());
```
bradleyharden, line 132: You reset the channel a few lines above, but then you do it again here. Am I missing something? Or is that just redundant?

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,313 @@
...
+            fun(dmac);
+        });
+    }
+
+    /// Set channel ID and run the closure. A closure is needed to ensure
+    /// the registers are accessed in an interrupt-safe way, as the SAMD21
+    /// DMAC is a little funky. For the SAMD51/SAMEx, we simply take a reference
+    /// to the correct channel number and run the closure on that.
+    #[cfg(any(
+        feature = "samd51",
```
bradleyharden, line 103: I think you could use `min-samd51g` here.

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,313 @@
...
+pub struct Uninitialized;
+/// Initialized and ready to transfer channel
+pub struct Ready;
+/// Busy channel
+pub struct Busy;
+
+/// DMA channel, capable of executing
+/// [`DmaTransfer`](super::transfer::DmaTransfer)s.
+pub struct Channel<Status, const ID: u8> {
+    _status: Status,
```
bradleyharden, line 66: This isn't strictly necessary, but I think it might help clarify intent and help prevent mistakes when refactoring. I would create a `Status` trait (with `Sealed` as a super trait). Then, I would implement `Status` for `Uninitialized`, `Ready` and `Busy`. Finally, I would have `Channel` accept a type parameter `S`, `where S: Status`.

Effectively, the status trait is like a meta-enum, and the `Uninitialized`, `Ready` and `Busy` types are variants of that type-level enum. To me, that makes the intent really clear.

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Start transfer on channel using the specified trigger source.
+    ///
+    /// # Return
+    ///
+    /// A `Channel` with a `Busy` status.
+    pub(crate) fn start(
+        mut self,
+        trig_src: TriggerSource,
+        trig_act: TriggerAction,
+        dmac: &mut DMAC,
```
bradleyharden, line 227: You might consider separating the configuration of the trigger source and action from the act of starting the transfer. You've already done that for the most part, but the trigger source and action are an exception. I would change that. I think it would also help to cut down on the latency from the point of calling `start` (or whatever calls `start`) to the point of the transfer actually starting, because you don't need to change the trigger settings before starting.

Also, again, I would move up the `dmac` argument.

hal/src/common/dmac/dma_controller.rs
```diff
@@ -0,0 +1,420 @@
...
+    DESCRIPTOR_SECTION, WRITEBACK,
+};
+use crate::target_device::{DMAC, PM};
+
+/// Initialized DMA Controller
+pub struct DmaController {
+    dmac: DMAC,
+}
+
+impl AsMut<DMAC> for DmaController {
```
bradleyharden, line 51: If I'm not mistaken, the implementation would allow users to get a reference to the `DMAC` struct in safe code. That would allow them to make arbitrary changes to the DMA configuration. I would propose adding `unsafe` functions called `dmac` and `dmac_mut` to serve the same purpose. That way users must use `unsafe` to access the `DMAC` directly.

hal/src/common/dmac/dma_controller.rs
```diff
@@ -0,0 +1,420 @@
...
+        // right now we blindly enable all priority levels
+        dmac.ctrl.modify(|_, w| {
+            w.lvlen3()
+                .set_bit()
+                .lvlen2()
+                .set_bit()
+                .lvlen1()
+                .set_bit()
+                .lvlen0()
+                .set_bit()
```
bradleyharden, line 105: This is another example of `rustfmt` murdering the code. I would separate these into individual field changes.

hal/src/common/dmac/dma_controller.rs
```diff
@@ -0,0 +1,420 @@
...
+
+        // Release the DMAC
+        self.dmac
+    }
+
+    /// Issue a software reset to the DMAC and wait for reset to complete
+    #[inline(always)]
+    fn swreset(dmac: &mut DMAC) {
+        dmac.ctrl.modify(|_, w| w.swrst().set_bit());
+        while !dmac.ctrl.read().swrst().bit_is_clear() {}
```
bradleyharden, line 183: This is a nit, but I would prefer `bit_is_set` here. I tend to avoid `!` when possible. It's really narrow and hard to see. It's also a prefix operator, which then applies to the condition at the end of the line (`bit_is_clear`). I think using the `bit_is_set` function is much more readable.

hal/src/common/dmac/dma_controller.rs
```diff
@@ -0,0 +1,420 @@
...
+        all(feature = "samd21", not(feature = "max-channels"))
+    ))]
+    pub fn split(&mut self) -> DmacChannels {
+        DmacChannels(
+            new_chan(),
+            new_chan(),
+            new_chan(),
+            new_chan(),
+            new_chan(),
+            new_chan(),
```
bradleyharden, line 205: Consider using the `seq_macro` crate to generate the `split` function and `DmacChannels` struct. It's basically a for loop in a macro. To do so, you might need to define your `NUM_CHANNELS` constant in a macro, but that's an easy change. Check the closed issues on the `seq_macro` GitHub for a clearer explanation of that last point.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,857 @@
...
+struct TransferConfiguration<B, Src, Dst>
+where
+    B: DmaBeat,
+    Src: DmaBuffer<B>,
+    Dst: DmaBuffer<B>,
+{
+    source: Src,
+    dest: Dst,
+    length: u16,
+    _t: PhantomData<B>,
```
bradleyharden, line 138: Do you need this `PhantomData`? Isn't it already constrained by `Src` and `Dst`? I might be wrong about that.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,857 @@
...
+    /// Check if the DMA address should be incrementing or not
+    fn incrementing(&self) -> bool;
+}
+
+impl<T: DmaBeat, const N: usize> DmaBuffer<T> for &mut [T; N] {
+    #[inline(always)]
+    fn to_dma_ptr(&mut self) -> *mut T {
+        let ptr = self.as_mut_ptr();
+        if self.incrementing() {
+            unsafe { ptr.add(N) }
```
bradleyharden, line 715: There's an easier way to do this. Use the `slice` method `as_mut_ptr_range`. It returns a `Range` struct with pointers to the first bye and the byte after the last. You can just ignore the `start` field and use the `end` field. It's easier and clearer than calculating it yourself.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,857 @@
...
+    /// Half word = [`u16`](core::u16)
+    HalfWord = 0x01,
+    /// Word = [`u32`](core::u32)
+    Word = 0x02,
+}
+
+// Trait enabling taking `*mut` references to references to a single `T`,
+/// references to arrays of `T` or slices of `T`, where `T: DmaBeat`,
+/// as well as required transfer length
+pub trait DmaBuffer<T: DmaBeat> {
```
bradleyharden, line 701: Normally, I would use `Sealed` as a super trait for something like this. But ff you use the blanket implementation approach for types that implement `AsMut<[T]>`, then you won't be able to do that. I'd have to think a little bit more about how you can prevent people from adding their own implementations.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,857 @@
...
+        let ptr = self.as_mut_ptr();
+        if self.incrementing() {
+            unsafe { ptr.add(N) }
+        } else {
+            ptr
+        }
+    }
+
+    #[inline(always)]
+    fn incrementing(&self) -> bool {
```
bradleyharden, line 722: Make this an associated constant. I don't see any advantage to the function approach.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,857 @@
...
+    }
+
+    fn incrementing(&self) -> bool {
+        false
+    }
+}
+
+/// Convert 8, 16 and 32 bit types
+/// into [`BeatSize`](BeatSize)
+pub trait DmaBeat {
```
bradleyharden, line 769: I would add `Sealed` as a super trait here. You don't want users to implement their own beat sizes.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,857 @@
...
+    /// Convert to BeatSize
+    fn to_beatsize() -> BeatSize;
+    /// Number of bytes in type
+    fn num_bytes() -> u8;
```
bradleyharden, line 773: Both `to_beatsize` and `num_bytes` would be better as associated constants.

hal/src/common/dmac/transfer.rs
bradleyharden: I'm not sure if it would help, but you could also use `num` traits as super traits as well, e.g. `PrimInt`, `AsPrimitive`, etc. That would allow you to work more easily with generic `DmaBeat`s, if that's a problem at all.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,857 @@
...
+        let length = 1;
+        let config = TransferConfiguration {
+            source,
+            dest,
+            length,
+            _t: PhantomData,
+        };
+        DmaTransfer {
+            xfer: UnsafeTransfer::setup(chan, config, circular_xfer, payload),
+        }
```
bradleyharden, line 458: In every wrapper function, you calculate a length, create a `TransferConfiguration` and then a `Transfer` of some type. I think it should be possible to do that all generically. This may not be the best idea, but it's one that comes to mind immediately. You could create a trait that is implemented on `(Src, Dst)` pairs. It would have a `length` function that would calculate the transfer length based on the tuple types.

hal/src/common/dmac/channel.rs
jbeaurivage: You're right, that was an oversight on my part. No need to reset the channel twice.

hal/src/common/dmac/transfer.rs
jbeaurivage: Yes, looks like it won't compile without the `PhantomData`

hal/src/common/dmac/transfer.rs
jbeaurivage: The function approach is necessary if we want to support `&mut [T]`'s of length unknown at compile time. We could just make `incrementing == true` for all `&mut [T]`, but I think that would be incorrect for slices of length 1.

hal/src/common/dmac/transfer.rs
bradleyharden: Hmm. I wonder why. It seems like it should be able to. I must be missing something.

hal/src/common/dmac/transfer.rs
bradleyharden: Hmm, yeah, looks like I didn't think that through all the way.

bradleyharden:
@jbeaurivage are you on the Matrix channel? Your name doesn't show up when I search. I had some spit-ball thoughts, but I'd rather not pollute the discussion here with ill-formed ideas.

ianrrees:
@jbeaurivage, I've just had a quick look over this, it's looking great!

This probably won't be very helpful, but for a work project I've been using the DMAC (via [samd-dma](https://github.com/proman21/samd-dma) copied in to a branch of the HAL) to and from the I2S peripheral with [these simplistic accessors](https://github.com/ianrrees/samd-i2s/blob/c9bd26b833eab2b5263fedea4269b7233638a18d/src/lib.rs#L425-L464) .

I'm also tinkering with accessing the USB via DMA, using [embedded-dma](https://github.com/rust-embedded/embedded-dma) traits - have you considered working them in to this?  I feel like those traits are slightly awkward when interfacing with peripherals like I2S, because the embedded-dma buffers have a fixed size but the peripheral can accept an infinite number of reads/writes.  It would also be nice if they could associate a trigger source with the "buffer".

jbeaurivage:
@ianrrees I think as far as this PR is concerned, it is ready to go as-is. I definitely do want to submit follow up PRs that add DMA support to a selection of peripherals. Maybe I could do one or two peripherals as examples and let the people here add support for other peripherals they're using (for example, I'm pretty unfamiliar with I2S and don't have I2S devices to test against). 

I think embedded-dma could also eventually be worked into this driver. Maybe "regular" transfers could implement these traits, and let more exotic transfer types (like circular) not implement them.

twitchyliquid64:
If everyones happy with this PR then im happy to merge. @bradleyharden u good with this?

bradleyharden:
Can I take one last look at it? I haven't looked at it in weeks, and @jbeaurivage has made a lot of changes.

twitchyliquid64:
Certainly!

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,357 @@
...
+#[cfg(feature = "min-samd51g")]
+use crate::target_device::dmac::CHANNEL;
+
+//==============================================================================
+// Channel Status
+//==============================================================================
+pub trait Status: Sealed {}
+
+/// Uninitialized channel
+pub struct Uninitialized;
```
bradleyharden, line 53: This is 100% a personal preference thing, but you can also implement these types as `enum`s. There are two main differences, as I see it. First, as `struct`s, you can instantiate them. But when they are blank `enum`s, they can't actually be instantiated at all. Rather, they only exist at the type level. To instantiate them, you need `PhantomData`. And second, they are grouped with the other `enum`s in the documentation. I prefer both of those options, so I typically opt for declaring them as `enum`s. I also view these as type-level variants of a type-level `enum` (the `Status` trait), so I prefer to group them with `enum`s.

Just thought I would add some food for thought.

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,357 @@
...
+    ///
+    /// A `Channel` with a `Ready` status
+    pub fn init(
+        mut self,
+        controller: &mut DmaController,
+        lvl: PriorityLevel,
+        enable_interrupts: bool,
+    ) -> Channel<Id, Ready> {
+        // SAFETY: this is safe because we only mutably borrow dmac once.
+        let dmac = unsafe { controller.dmac_mut() };
```
bradleyharden, line 173: Instead of using the `unsafe` function `dmac_mut` everywhere, you might consider making the `dmac` field `pub(super)`. I think that would be cleaner and remove some of the `unsafe` noise.

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /// Stop transfer on channel whether or not the transfer has completed
+    ///
+    /// # Return
+    ///
+    /// A `Channel` with a `Ready` status, ready to be reused by a new
+    /// [`Transfer`](super::transfer::Transfer)
+    #[inline]
+    pub(crate) fn stop(mut self, dmac: &mut DMAC) -> Channel<Id, Ready> {
```
bradleyharden, line 323: Is this actually called anywhere? I can't seem to find it.

hal/src/common/dmac/channel.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /// Wait for the channel to clear its busy status, then release the channel.
+    ///
+    /// # Return
+    ///
+    /// A `Channel` with a `Ready` status, ready to be reused by a new
+    /// [`Transfer`](super::transfer::Transfer)
+    #[inline]
+    pub(crate) fn free(self, dmac: &mut DMAC) -> Channel<Id, Ready> {
```
bradleyharden, line 350: I was going to suggest a non-blocking version of this, but then I saw it was `pub(crate)`. I followed the callers but couldn't find a caller of `stop`.

hal/src/common/dmac/dma_controller.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    /// Return a mutable reference to the underlying DMAC object exposed by the
+    /// PAC.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because `DmaController` may expect certain
+    /// registers to retain a configuration. Messing with that configuration may
+    /// be unsafe.
+    pub unsafe fn dmac_mut(&mut self) -> &mut DMAC {
```
bradleyharden, line 151: I've started to question my own use of `&mut` references to PAC structs. There doesn't seem to be much value, since each PAC struct contains only `VolatileCell`s, which can be mutated through shared references. There doesn't seem to be any upside, but the downside is possible UB if you accidentally alias. Do you have any counter arguments? I haven't yet convinced myself one way or the other.

hal/src/common/dmac/dma_controller.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        Self::swreset(&mut dmac);
+
+        // SAFETY this is safe because we write a whole u32 to 32-bit registers,
+        // and the descriptor array addesses will never change since they are static.
+        // We just need to ensure the writeback and descriptor_section addresses
+        // are valid.
+        unsafe {
+            dmac.baseaddr
+                .write(|w| w.baseaddr().bits(DESCRIPTOR_SECTION.as_ptr() as u32));
```
bradleyharden, line 176: Does rustfmt do this? Wouldn't it be better to break at the beginning of the closure body? Maybe adding brackets would help?

hal/src/common/dmac/dma_controller.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        // Enable DMA controller
+        dmac.ctrl.modify(|_, w| w.dmaenable().set_bit());
+
+        Self { dmac }
+    }
+
+    /// Enable multiple priority levels simultaneously
+    #[inline]
+    pub fn enable_levels(&mut self, mask: PriorityLevelMask) {
```
bradleyharden, line 197: I really like how these functions turned out with the mask types.

hal/src/common/dmac/dma_controller.rs
```diff
@@ -0,0 +1,293 @@
...
+    /// Issue a software reset to the DMAC and wait for reset to complete
+    #[inline]
+    fn swreset(dmac: &mut DMAC) {
+        dmac.ctrl.modify(|_, w| w.swrst().set_bit());
+        while dmac.ctrl.read().swrst().bit_is_set() {}
+    }
+
+    /// Split the DMAC into individual channels
+    #[cfg(all(feature = "samd11", not(feature = "max-channels")))]
+    pub fn split(&mut self) -> Channels {
```
bradleyharden, line 271: Won't this conflict with the definition below? It seems like this should just be another `with_num_channels` option added to `mod.rs`.

hal/src/common/dmac/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+#[macro_export]
+macro_rules! with_num_channels {
+    ($some_macro:ident) => {
+        $some_macro!(32);
+    };
+}
+
+#[cfg(all(feature = "samd11", not(feature = "max-channels")))]
+#[macro_export]
+macro_rules! with_num_channels {
```
bradleyharden, line 190: Doesn't this already cover the case I mentioned in `dma_controller.rs`?

hal/src/common/dmac/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+/// Descriptor representing a SRAM register. Datasheet section 19.8.2
+#[derive(Clone, Copy)]
+#[repr(C, align(16))]
+#[doc(hidden)]
+pub struct DmacDescriptor {
+    btctrl: BlockTransferControl,
+    btcnt: u16,
+    srcaddr: *mut (),
+    dstaddr: *mut (),
+    descaddr: *mut DmacDescriptor,
```
bradleyharden, line 256: Will you ever mutate another descriptor through this pointer? Does it need to be `*mut`?

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,545 @@
...
+/// Buffer useable by the DMAC.
+pub unsafe trait Buffer {
+    type Beat: Beat;
+    /// Pointer to the buffer. If the buffer is incrementing, the address should
+    /// point to the last beat transfer in the block.
+    fn dma_ptr(&mut self) -> *mut Self::Beat;
+    /// Return whether the buffer pointer should be incrementing or not
+    fn incrementing(&self) -> bool;
+    /// Buffer length in beats
+    fn buffer_len(&self) -> usize;
```
bradleyharden, line 157: I had a realization about this a while back, but I forgot about it. I can't remember if I told you or not. If I did and you disregarded it, then please feel free to ignore it again.

I realized that inherent `impl` names never conflict with trait `impl` names. If there is an inherent `impl` name, it will always be used, even if there is a trait function with the same name. If you want to use the trait function in that case, you have to explicitly call it.

That works to your advantage here. I think you can use the name `len` here without worrying. If the type is a slice, then it has an inherent implementation of `len` which is already what you want. But if it's not a slice and doesn't already have a `len` function, you can add one through this trait. Personally, I like the symmetry of that approach, but there is maybe some possibility for confusion, so I can see arguments against it.

hal/src/common/dmac/transfer.rs
```diff
@@ -0,0 +1,545 @@
...
+            self.payload,
+        )
+    }
+
+    /// Non-blocking; Immediately stop the DMA transfer and release all owned
+    /// resources
+    pub fn stop(self, dmac: &mut DmaController) -> (Channel<ChannelId<C>, Ready>, S, D, P) {
+        // SAFETY: This is safe because we only borrow dmac once.
+        let dmac = unsafe { dmac.dmac_mut() };
+        let chan = self.chan.into().stop(dmac);
```
bradleyharden, line 531: Ah, here it is. It didn't show up with a CTRL+F, because the diff was too large and hidden by default.

bradleyharden:
@jbeaurivage, I think it turned out really well. I like how clean everything is with the `Any_` traits. I'm glad to see the pattern found use outside my own code.

Resolve the remaining comments however you want. I think it's good to merge.

Also, hurray for `min_const_generics` ​🎉​! It's not perfect, but it's something.

hal/src/common/dmac/channel.rs
bradleyharden: Found it. See below

hal/src/common/dmac/dma_controller.rs
bradleyharden: Actually, isn't it instant UB if you have a simultaneous `&` reference and `&mut` reference? Seems risky. I'm leaning toward only using `&` from now on.

jbeaurivage: I think you're actually right. I'm going with `pub(super) fn dmac(&mut self) -> &DMAC`.

hal/src/common/dmac/mod.rs
jbeaurivage: I think these can indeed safely be `*const`

hal/src/common/dmac/dma_controller.rs
jbeaurivage: Yeah, that's rustfmt's fault. Even when adding a break with brackets, rustfmt will revert the formatting to above. I'm leaving it as is, otherwise rustfmt CI won't pass.

hal/src/common/dmac/transfer.rs
jbeaurivage: I will keep `buffer_len` because this warning pops up when changing the trait method name to `len`:
``` rust
    |
201 |     fn len(&self) -> usize {
    |     ^^^^^^^^^^^^^^^^^^^^^^ cannot return without recursing
202 |         self.len()
    |         ---------- recursive call site
    |
    = note: `#[warn(unconditional_recursion)]` on by default
    = help: a `loop` may express intention better if this is on purpose
```

hal/src/common/dmac/dma_controller.rs
jbeaurivage: Yeah, this should have been removed when `with_num_channels` was introduced.

jbeaurivage:
Alright friends this is positively ready to go.

Huge thanks to @bradleyharden for all the help.
@twitchyliquid64 all yours now!

twitchyliquid64:
YAAAAAS Thank you everyone for getting this over the line!

hal/src/common/dmac/transfer.rs
bradleyharden: I was curious why this didn't work. I think I got it to work [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6200c0f4649fe6ca1484898a01812962).

That error must have come when implementing `Buffer` for arrays, right? I think I understand what's going on.

Arrays don't define their own `len` function. Instead, they use unsized coercion to access `len` defined for slices. When you call `self.len()` within the `Buffer` context, it ends up choosing the `Buffer` definition of `len` over the coerced definition from slices. The solution is to explicitly do the coercion with `<[T]>::len(self)`.

PR #393, opened 2021-01-28
samd21 device adding support in USB and gpio

jaredwolff:
Related to @bradleyharden's recent PR #371. Leaving the old but also adding the new for folks who want to transition to v2. I left the old stuff but wasn't sure if it was even necessary. Happy to make adjustments as need be!


bradleyharden:
@jaredwolff, I just saw this. You shouldn't need to modify existing code at all to be compatible with the new GPIO module. You only need to perform a conversion between v1 and v2 Pin types. The v1 Pin type is now implemented as a newtype wrapper of the corresponding v2 Pin, so the conversion is trivial, and I have From/Into implementations for it.

In this specific case, if you're using v2 Pins but want to use the ADC module, just do

```rust
let Pa8<_> = pins.pa08.into()
```
where `pins` is from v2 and `Pa8` is a v1 type.

I still owe you some better documentation for migration. I think that will help. I've been trying to finish up my SPI module. Now that it's in a PR, I'll drop back and look at the docs.

jaredwolff:
Ahh I totally missed that @bradleyharden. I'll give it a whirl and report back.

jaredwolff:
I'm going to close this PR and we can discuss more in #403 403

PR #308, opened 2020-11-07
hal: Switch from custom time traits to `embedded_time`

twitchyliquid64:
WIP

Anyone got a thumbv7-based board they can test the timers/pwm on to make sure the new divider calcuations are correct?

PR #394, opened 2021-01-28
Make atsamd-hal build with no features specified (for library)

henrikssn:
We had a discussion about this on Matrix earlier today.

The goal is to be able to use the device agnostic parts of hal/common in a library, which is shared by crates targeting different SAMD devices. Eventually, this should be replaced by embedded-time (see #308) so this a form of of a stop-gap solution until that lands.


twitchyliquid64:
Can you add something to `.github/workflows/build-bsp.yml` to test that the build works with no features specified? given theres no BSP crate backing it, it will probably need to be its own job rather than a step in the `build` job.

.github/workflows/build-bsp.yml
```diff
@@ -45,6 +45,12 @@ jobs:
         cd boards/${{ matrix.bsp }}
         $(${build_invocation})
 
+    - name: Test Build HAL
```
twitchyliquid64, line 48: Can you move this into its own job? Where it currently is, it will run for every BSP (~60 times), which is a lot of redundant overhead.

henrikssn: Yeah I just realized... I need to read the docs for github actions, nevver used them before

henrikssn: Yeah I just realized... I need to read the docs for github actions, never used them before

henrikssn: Yeah I just realized... I need to read the docs for github actions, never used them before

twitchyliquid64: If you just duplicate + rename the `build` stanza, remove the steps after `Install Rust (thumbv7em)`, remove the `needs` and `strategy` stanzas, and what your step at the end it should work

Oh and change `${{ matrix.toolchain }}` to `stable`

twitchyliquid64:
Looks good, I dont think you need `needs: setup` (thats just to build the matrix for the BSP jobs), but after that LGTM

twitchyliquid64 (approved):

PR #398, opened 2021-02-02
Support enable/disable SOF interrupt in USB peripheral

henrikssn:
@twitchyliquid64 I haven't tested it yet but is this inline with your suggestion?

### DIFF
--- a/hal/src/samd21/usb/bus.rs
+++ b/hal/src/samd21/usb/bus.rs
@@ -716,6 +716,14 @@ impl Inner {
         usb.ctrlb.modify(|_, w| w.detach().clear_bit());
     }
 
+    fn sof_interrupt(&self, enable: bool) {
+        if enable {
+            self.usb().intenset.write(|w| w.sof().set_bit());
+        } else {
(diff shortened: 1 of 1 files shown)


twitchyliquid64:
Yepp!

We should be careful about testing it extensively, depending on the pecking order of the interrupts we may need to clear it in other places of `poll()` as well.

ianrrees:
Thanks for this!  It's a really handy bit towards solving a problem I've been mulling over.  What's the use case that prompted this initially?

I'm toying with a usb class implementation that is meant to do all its work in the `poll()` method, with an eye to making the bus own the class (per 2 [here](https://github.com/mvirkkunen/usb-device/issues/9)).  One issue with that approach currently, is that the class doesn't have a way to notify when it wants attention from the bus.

Currently this change doesn't trigger `poll()` for each class, but it looks like a straightforward modification to usb-device could change that, or [this](https://github.com/mvirkkunen/usb-device/issues/32#issuecomment-612592582) could be a more permanent solution.

jbeaurivage:
Closing, as this was superseded by #456.

PR #400, opened 2021-02-04
Experimental ADC sampling rate parameter

kdsch:
I'm sharing this patch for the sake of [discussion on Matrix](https://matrix.to/#/!JbgTnIGpDxEHkjzSep:gitter.im/$2kDhFTB42WdbPqoIvhCJlgLXxhfyqhfWI0yB0FETCSg?via=matrix.org).

### DIFF
--- a/hal/src/common/thumbv7em/clock.rs
+++ b/hal/src/common/thumbv7em/clock.rs
@@ -205,9 +205,9 @@ impl GenericClockController {
             state.reset_gclk();
             state.set_gclk_divider_and_source(GCLK1, 1, XOSC32K, false);
         } else {
-            enable_internal_32kosc(osc32kctrl);
-            state.reset_gclk();
-            state.set_gclk_divider_and_source(GCLK1, 1, OSCULP32K, false);
+            //enable_internal_32kosc(osc32kctrl);
+            //state.reset_gclk();
+            //state.set_gclk_divider_and_source(GCLK1, 1, OSCULP32K, false);
         }
 
         while state.gclk.syncbusy.read().genctrl().is_gclk0() {}
@@ -247,7 +247,7 @@ impl GenericClockController {
             state,
             gclks: [
                 OSC120M_FREQ,
-                OSC32K_FREQ,
+                Hertz(0), //OSC32K_FREQ,
                 Hertz(0),
                 Hertz(0),
                 Hertz(0),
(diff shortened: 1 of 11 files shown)


hal/src/common/thumbv7em/clock.rs
```diff
@@ -205,9 +205,9 @@ impl GenericClockController {
             state.reset_gclk();
             state.set_gclk_divider_and_source(GCLK1, 1, XOSC32K, false);
         } else {
-            enable_internal_32kosc(osc32kctrl);
-            state.reset_gclk();
-            state.set_gclk_divider_and_source(GCLK1, 1, OSCULP32K, false);
+            //enable_internal_32kosc(osc32kctrl);
```
twitchyliquid64, line 208: Hmmm?

twitchyliquid64 (commented):
Overall LGTM - we definitely follow the pattern of initializing peripherals with a clock and a frequency, and doing our best to calculate the best divider values.

Would you be able to use [TimerParams](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/common/timer_params.rs#L6) to compute the divider instead of doing it inline?

kdsch:
@twitchyliquid64 TimerParams is specific to timers; the ADC divider is different. Might be a good idea to move that logic into a function, though.

hal/src/common/thumbv7em/clock.rs
kdsch: On the SAMD5x, GCLK1 is the only generic clock generator whose maximum division factor is 2^16. All the others are 2^9.  See [14.8.3](http://ww1.microchip.com/downloads/en/DeviceDoc/60001507E.pdf#unique_5042). If DFLL is used as the input to GCLK1, then we can have a clock frequency 48MHz/[1..2^16]. 48kHz happens to be a standard audio sampling rate, so dividing the 48MHz clock by 1000 (only possible with GCLK1) would achieve an exact 48kHz ADC clock with minimal fuss. Unfortunately this doesn't account for the clock division that the ADC performs, but in certain cases it might be possible to achieve a 48kHz sample rate. Since GCLK1 is hardcoded by the HAL as the 32kHz oscillator, I disabled it to investigate this clocking approach.

bradleyharden:
I think this problem points to the current lack of flexibility in setting up the clocks. I remember running into that for my project. I ended up doing a lot of manual configuration with the PAC, and I created prototype `gclk` and `mclk` modules. Now that I managed to push a PR for the `spi` module, I'll look into what it would take to move the clock modules beyond the prototype stage. That would let users set it up however they want.

hal/src/common/thumbv7em/clock.rs
twitchyliquid64: Ah okies, just remember to uncomment when youre ready!

PR #402, opened 2021-02-09
New spi modules

bradleyharden:
This is a pretty big addition. I'm marking it as WIP mostly because I need #401 to be merged first. But I also think it might take some time in review.

The best way to get started is probably the documentation. Is there some way for me to render it for reviewers? Or do reviewers need to download it and render it locally?

@twitchyliquid64, you have talked through some of the SPI stuff with me before. @blm768, this is an extension of the GPIO and SERCOM pad work from before. @jbeaurivage, I think there's a lot of overlap in approach between this and your DMAC stuff. Although safety is less prominent of an issue for this.

I also added a few dependencies. I'm not sure how averse you are to that. If you have any specific objections, there's probably a workaround.

### Commit message

Move declarations of the `Sercom` types from the `pads` module to the base `sercom` module. Make them type aliases of the PAC SERCOM types. Add a `Sercom` trait method to enable the APB clock.

Create new SPI modules for both `thumbv6m` and `thumbv7em`. The new modules are more complete, more configurable and safer than the existing modules. They make use of the `v2` `gpio` and `sercom::pads` modules.

Also add an `SpiFuture` type that can execute SPI transactions autonomously, using interrupts.

### DIFF
--- a/hal/Cargo.toml
+++ b/hal/Cargo.toml
@@ -20,16 +20,23 @@ features = ["samd21g", "samd21g-rt", "unproven", "usb"]
 
 [dependencies]
 bitfield = "0.13"
+bitflags = "1.2.1"
 cortex-m = "0.6"
 embedded-hal = "0.2"
+generic-array = "0.14.4"
 nb = "0.1"
 paste = "1.0"
 rand_core = "0.5"
+seq-macro = "0.2.1"
 vcell = "0.1"
 
 [dependencies.jlink_rtt]
 optional = true
 version = "0.2"
+
+[dependencies.num-traits]
+default-features = false
+version = "0.2.14"
 
 [dependencies.usb-device]
 version = "0.2"
--- a/hal/src/common/sercom/v2/spi_future.rs
+++ b/hal/src/common/sercom/v2/spi_future.rs
@@ -1,0 +1,549 @@
+#![allow(broken_intra_doc_links)]
+//! A [`Future`]-like interface for SPI transactions
+//!
+//! An [`SpiFuture`] takes ownership of an [`Spi`] `struct` and a `[u8]`-like
+//! buffer. It then executes a full-duplex SPI transaction using iterrupts. On
+//! each `RXC` or `DRE` interrupt, the [`SpiFuture`] reads or sends `STEP` bytes
+//! of the buffer, where `STEP` is a value that depends on [`CharSize`], for
+//! SAMD11 & SAMD21 chips, or [`Length`], for SAMD51 & SAME5x chips.
+//!
+//! The provided buffer must implement [`AsRef`] and [`AsMut`] for `[u8]`, it
+//! must have an appropriate length (see below), and it must have a `'static`
+//! lifetime, i.e. it must be owned or a `&'static mut` reference.
+//!
+//! [`SpiFuture`] has extra, optional capabilities as well. It can accept a
+//! function or closure that will be called on completion of the transaction,
+//! acting like a [`Waker`]. And it can take a GPIO [`Pin`] to use as the SS
+//! line. If provided, the [`Pin`] will be set low at the beginnging of the
+//! transfer and brought high at completion.
+//!
+//! Calling [`start`] will enable the `DRE` and `RXC` interrupts and begin the
+//! transaction.
+//!
+//! ```
+//! use atsamd_hal::gpio::v2::{Pin, PA10, PushPullOutput};
+//! use atsamd_hal::sercom::v2::spi::AnySpi;
+//! use atsamd_hal::sercom::v2::spi_future::SpiFuture;
+//!
+//! fn wake_up() {
+//!     //...
+//! }
+//!
+//! fn use_future(spi: impl AnySpi, ss: Pin<PA10, PushPullOutput>) {
+//!     let buf = [0_u8; 12];
+//!     let future = SpiFuture::new(spi, buf)
+//!         .with_waker(wake_up)
+//!         .with_ss(ss);
+//!     future.start();
+//! }
+//! ```
+//!
+//! When sending and receiving finish, the [`SpiFuture`] will automatically
+//! disable the `DRE` and `RXC` interrupts. To test whether an [`SpiFuture`] is
+//! complete, use the [`poll`] method. While the transaction is in progress, it
+//! will return [`Poll::Pending`]. When the transaction is complete, it will
+//! return [`Poll::Ready`]. Once complete, you can consume the [`SpiFuture`] and
+//! [`free`] the constituent pieces. Doing so before the transfer has completed
+//! is `unsafe`.
+//!
+//! The actual transfer is performed by the [`send`] and [`recv`] methods, which
+//! should be called from the `DRE` and `RXC` interrupt handlers, respectively.
+//!
+//! ## `STEP` size and buffer length
+//!
+//! For SAMD11 & SAMD21 chips, `STEP` is equal to the number of bytes in the
+//! corresponding the [`CharSize::Word`] type, i.e. 1 for [`EightBit`] and 2 for
+//! [`NineBit`]. For SAMD51 & SAME5x chips, `STEP` is equal to the [`Length`] or
+//! 4, whichever is less.
+//!
+//! The provided buffer must have an appropriate length. For SAMD11 & SAMD21
+//! chips, as well as SAMDx5x chips with [`Length`]` <= 4`, a single write of
+//! `STEP` bytes represents an entire SPI transaction. In these cases, the
+//! provided buffer must represent an integer number of transactions. For
+//! example, a SAMD51 [`Spi`] struct with a [`Length`] of 3 could use buffers of
+//! length 3, 6, 9, etc. For longer [`Length`] values, the provided buffer must
+//! represent exactly one SPI transaction, so the buffer length must be equal to
+//! [`Length`]. For example, a SAMD51 [`Spi`] struct with a [`Length`] of 17
+//! could only use a buffer with exactly 17 bytes.
+//!
+//! Keep in mind that [`SpiFuture`] works only with `[u8]`-like things, which
+//! can introduce some limitations.
+//!
+//! Suppose you plan to execute [`NineBit`] transactions with a SAMD21 chip.
+//! Your data may come in the form of a `[u16]` slice. However, to use it with
+//! [`SpiFuture`], you will need reformulate it as a `[u8]` slice. The easiest
+//! way to do so is probably to transmute the slice to `[u8]` or cast the
+//! reference to `&'static mut [u8]`. Both of these operations are sound,
+//! because [`u8`] has no alignment requirements.
+//!
+//! In another scenario, suppose you wanted to use a SAMx5x chip with a
+//! transaction [`Length`] of 3 bytes. Your data might come in the form of a
+//! `[u32]` slice. In this situation, it would **not** be appropriate to
+//! transmute or cast the data to a `[u8]` slice. [`SpiFuture`] expects the data
+//! to be a *byte-packed* `[u8]` slice, so the extra byte in each `u32` makes it
+//! incompatible.
+//!
+//! ## [RTIC] example
+//!
+//! The [RTIC] framework provides a convenient way to store a `static`ally
+//! allocated [`SpiFuture`], so that it can be accessed by both the interrupt
+//! handlers and user code. The following example shows how [`SpiFuture`]s might
+//! be used for a series of transactions. It was written for a SAMx5x chip, and
+//! it uses features from the latest release of [RTIC], `v0.6-alpha.0`.
+//!
+//! ```
+//! use core::task::Poll;
+//! use atsamd_hal::pads_alias;
+//! use atsamd_hal::sercom::v2::Sercom0;
+//! use atsamd_hal::sercom::v2::pads::{IoSet1, Pad0, Pad3};
+//! use atsamd_hal::sercom::v2::spi::{Master, lengths::U12};
+//! use atsamd_hal::sercom::v2::spi_future::SpiFuture;
+//! use atsamd_hal::gpio::v2::pins::{Pin, PA10, PushPullOutput};
+//!
+//! type SS = Pin<PA10, PushPullOutput>;
+//! pads_alias!(type Pads = Pads<Sercom0, IoSet1, DI = Pad0, DO = Pad3>);
+//! type Spi = spi::Spi<spi::Config<Pads, Master, U12>>;
+//! type Future = SpiFuture<Spi, [u8; 12], SS, fn()>;
+//!
+//! //...
+//!
+//! #[resources]
+//! struct Resources {
+//!     #[task_local]
+//!     #[init(None)]
+//!     opt_spi_ss: Option<(Spi, SS)>,
+//!
+//!     #[lock_free]
+//!     #[init(None)]
+//!     opt_future: Option<Future>,
+//! }
+//!
+//! #[task(resources = [opt_spi_ss, opt_future])]
+//! fn task(ctx: task::Context) {
+//!     let task::Context { opt_spi_ss, opt_future } = ctx;
+//!     match opt_future {
+//!         Some(future) => {
+//!             if let Poll::Ready(_) = future.poll() {
+//!                 let (spi, buf, ss) = opt_future.take().unwrap().free();
+//!                 *opt_spi_ss = Some((spi, ss));
+//!                 use_data(buf);
+//!             }
+//!         }
+//!         None => {
+//!             if let Some((spi, ss)) = opt_spi_ss.take() {
+//!                 let buf: [u8; 12] = get_data();
+//!                 let future = opt_future.get_or_insert(
+//!                     SpiFuture::new(spi, buf)
+//!                         .with_waker(|| { task::spawn().ok(); })
+//!                         .with_ss(ss)
+//!                 );
+//!                 future.start();
+//!             }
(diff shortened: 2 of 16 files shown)


hal/Cargo.toml
```diff
@@ -20,17 +20,25 @@ features = ["samd21g", "samd21g-rt", "unproven", "usb"]
...
 bitfield = "0.13"
+bitflags = "1.2.1"
 cortex-m = "0.6"
 embedded-hal = "0.2"
+generic-array = "0.14.4"
 nb = "0.1"
 paste = "1.0"
 rand_core = "0.5"
+seq-macro = "0.2.1"
+typenum = "1.12"
```
Sympatron, line 31: `generic-array` re-exports `typenum`. Maybe it would be better to use that to avoid accidental version mismatches in the future? Not a major problem in my opinion, just a thought.

hal/src/common/sercom/v2/spi_future.rs
```diff
@@ -0,0 +1,469 @@
...
+
+    /// Poll the [`SpiFuture`]
+    ///
+    /// This function will return [`Poll::Pending`] until all bytes have been
+    /// received. When the transaction is complete, it will return
+    /// [`Poll::Ready`] with a reference to the completed buffer.
+    #[inline]
+    pub fn poll(&self) -> Poll<&[u8]> {
+        let buf = self.buf.as_ref();
+        if self.rcvd >= buf.len() {
```
Sympatron, line 434: Shouldn't this read and all writes to `self.rcvd` be volatile?

hal/Cargo.toml
bradleyharden: Yeah, that's a good idea. I don't really need a specific version of either. I'll do that.

hal/src/common/sercom/v2/spi_future.rs
bradleyharden: Hmm. I don't think so. But maybe there's something subtle I'm missing.

I think you're suggesting that multiple reads of `self.rcvd` by repeated calls to `poll` would be optimized away, so you would never see updates to `self.rcvd`. Is that right?

If you're using accessing `SpiFuture` from interrupts at different priority levels, then it would need to be wrapped in a `Mutex` of some type. In RTIC, you would `poll` from a closure passed to the `Mutex`. There's no `Mutex` type in `core`, so I'm not sure what you would do there. But any solution would probably look like `std::sync::Mutex`, where locking would create a new `MutexGuard`. In either case, it wouldn't make sense to `poll` more than once, since the `Mutex` ensures your access can't be preempted.

If you're accessing `SpiFuture` from interrupts at the same priority level, then they can't preempt each other to begin with. So once again, it doesn't make sense to `poll` more than once.

Is there a scenario I'm missing? Or something I'm misunderstanding?

bradleyharden:
I just realized something I would like to add. Right now, the SPI transaction `Length` (1-255 bytes) for SAMD51 & SAME5x chips **must** be known at compile time. I think that works well for most cases, but there are probably some cases where you would want to set the length dynamically.

I think I will add an extra, marker type for the `Length` trait that indicates a run-time dynamic length. I think that should be easy to add.

hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -0,0 +1,1768 @@
...
+            spi::Phase::CaptureOnFirstTransition => false,
+            spi::Phase::CaptureOnSecondTransition => true,
+        };
+        self.sercom.spi().ctrla.modify(|_, w| w.cpha().bit(cpha));
+        self
+    }
+
+    /// Change the SPI mode (clock polarity & phase)
+    #[inline]
+    pub fn spi_mode(self, mode: spi::Mode) -> Self {
```
twitchyliquid64, line 1127: `Mode`, `Phase`, `Polarity` are part of the API, could we re-export so peeps dont have to throw `use atsamd_hal::hal::spi::{Mode, Phase, Polarity};` everywhere?

hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -0,0 +1,1768 @@
...
+//impl_dipo_dopo!((NoneT, Pad3, Pad1, NoneT): (3, 2));
+
+//impl_dipo_dopo!((NoneT, Pad0, Pad3, Pad1): (3, 3));
+//impl_dipo_dopo!((NoneT, Pad0, Pad3, NoneT): (3, 3));
+
+//=============================================================================
+// Pads
+//=============================================================================
+
+/// Encapsulate the set of pads for an SPI peripheral
```
twitchyliquid64, line 341: Encapsulates

hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -0,0 +1,1768 @@
...
+/// ```no_run
+/// # use atsamd_hal::target_device::Peripherals;
+/// # use atsamd_hal::gpio::v2::Pins;
+/// # use atsamd_hal::sercom::v2::Sercom0;
+/// # use atsamd_hal::sercom::v2::pads::{Pad0, Pad1, Pad2, Pad3};
+/// # use atsamd_hal::sercom::v2::spi;
+/// let mut peripherals = Peripherals::take().unwrap();
+/// let pins = Pins::new(peripherals.PORT);
+/// let pads = spi::Pads::<Sercom0>::new()
+///     .data_in::<Pad0, _>(pins.pb08)
```
twitchyliquid64, line 378: Maybe add a comment that `data_in` corresponds to miso when SPI is in master mode? or on the `data_in` docs?

hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -0,0 +1,1768 @@
...
+///     .data_out::<Pad3, _>(pins.pb11)
+///     .sclk::<Pad1, _>(pins.pb09);
+/// ```
```
twitchyliquid64, line 381: It might be useful as well to add a note that you need to provide the pins as `v2` GPIO types, to avoid needing verbose type declarations.

hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -0,0 +1,1768 @@
...
+    #[inline]
+    pub fn op_mode<M2: Mode>(self) -> Config<P, M2, C> {
+        M2::configure(&self.sercom);
+        self.change()
+    }
+
+    /// Change the [`CharSize`]
+    #[inline]
+    pub fn char_size<C2: CharSize>(self) -> Config<P, M, C2> {
+        C2::configure(&self.sercom);
```
twitchyliquid64, line 1099: It might be useful to specify the default here?

twitchyliquid64 (approved):
Looks great thanks! Just a few minor comments.

Heres a patch for the wio_terminal display system that switches it to the new API, feel free to throw it in too.

```
diff --git a/boards/wio_terminal/src/display.rs b/boards/wio_terminal/src/display.rs
index 1bdb32c21..5b576f9fa 100644
--- a/boards/wio_terminal/src/display.rs
+++ b/boards/wio_terminal/src/display.rs
@@ -1,14 +1,17 @@
 use atsamd_hal::clock::GenericClockController;
 use atsamd_hal::gpio::*;
 use atsamd_hal::hal::blocking::delay::DelayMs;
-use atsamd_hal::hal::spi;
 use atsamd_hal::prelude::*;
-use atsamd_hal::sercom::{PadPin, SPIMaster7, Sercom7Pad1, Sercom7Pad2, Sercom7Pad3};
 use atsamd_hal::target_device::{MCLK, SERCOM7};
 use atsamd_hal::time::Hertz;
 use display_interface_spi::SPIInterface;
 use ili9341::{Ili9341, Orientation};

+use atsamd_hal::hal::spi::{Mode, Phase, Polarity};
+use atsamd_hal::sercom::v2::pads::{IoSet4, Pad, Pad1, Pad3};
+use atsamd_hal::sercom::v2::{spi, Sercom7};
+use atsamd_hal::typelevel::NoneT;
+
 /// ILI9341 LCD display pins (uses `SERCOM7`)
 pub struct Display {
     /// LCD MISO pin
@@ -36,7 +39,17 @@ pub struct Display {
 /// Type alias for the ILI9341 LCD display.
 pub type LCD = Ili9341<
     SPIInterface<
-        SPIMaster7<Sercom7Pad2<Pb18<PfD>>, Sercom7Pad3<Pb19<PfD>>, Sercom7Pad1<Pb20<PfD>>>,
+        spi::Spi<
+            spi::Config<
+                spi::Pads<
+                    SERCOM7,
+                    IoSet4,
+                    NoneT,
+                    Pad<SERCOM7, Pad3, IoSet4>,
+                    Pad<SERCOM7, Pad1, IoSet4>,
+                >,
+            >,
+        >,
         Pc6<Output<PushPull>>,
         Pb21<Output<PushPull>>,
     >,
@@ -58,23 +71,27 @@ impl Display {
         clock_speed: F,
         delay: &mut D,
     ) -> Result<(LCD, Pc5<Output<PushPull>>), ()> {
-        // Initialize the SPI peripherial on the configured pins, using SERCOM7.
+        let sclk: atsamd_hal::gpio::v2::Pin<_, _> = self.sck.into();
+        let mosi: atsamd_hal::gpio::v2::Pin<_, _> = self.mosi.into();
+
+        let pads = spi::Pads::<Sercom7, IoSet4>::new()
+            .sclk(sclk)
+            .data_out::<Pad3, _>(mosi);
+
+        // Wire the clock for SERCOM7 to the system clock.
         let gclk0 = clocks.gclk0();
-        let spi = SPIMaster7::new(
-            &clocks.sercom7_core(&gclk0).ok_or(())?,
-            clock_speed,
-            spi::Mode {
-                phase: spi::Phase::CaptureOnFirstTransition,
-                polarity: spi::Polarity::IdleLow,
-            },
-            sercom7,
-            mclk,
-            (
-                self.miso.into_pad(port),
-                self.mosi.into_pad(port),
-                self.sck.into_pad(port),
-            ),
-        );
+        clocks.sercom7_core(&gclk0).ok_or(())?;
+
+        // Initialize the SPI peripherial on the configured pads, using SERCOM7.
+        let spi = spi::Config::new(mclk, sercom7, pads, gclk0)
+            .baud(clock_speed)
+            .length::<spi::lengths::U1>()
+            .msb_first(true)
+            .spi_mode(Mode {
+                phase: Phase::CaptureOnFirstTransition,
+                polarity: Polarity::IdleLow,
+            })
+            .enable();

         // Configure the chip select, data/command, and reset pins as push-pull outputs.
         let cs = self.cs.into_push_pull_output(port);
```

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: Good idea

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: I meant it in the imperative mood. But I may not have been consistent across  the docs. I'll check it and make sure

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: Good call. Will do

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: Good catch

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: I thought I designed it to work with either pin type. Is that not working for you? See my comment on Matrix

twitchyliquid64: Nop, due to ambiguity in type inference :/ More details in my comment on matrix

twitchyliquid64: With into:

```
cannot infer type for struct `Pad<SERCOM7, Pad1, IoSet4>`
```

Without:
```
the trait `From<atsamd_hal::gpio::Pin<PB20, atsamd_hal::gpio::v2::Input<atsamd_hal::gpio::v2::Floating>>>` is not implemented for `Pad<SERCOM7, Pad1, IoSet4>
```

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: Wait, what do you mean by default? Do you mean the default `Config` returned by `new`? Looks like I forgot to document the default `CharSize` there. But there's no default for the `char_size` function. You have to provide the desired type.

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: I should have a fix for this.

hal/src/common/sercom/v2/spi_future.rs
Sympatron: You obviously thought this through a lot more than I have. I am pretty sure that you would need volatile for similar code in C, which I am more used to. I forget that Rust's ownership/safety model prevents these kinds of mistakes.

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: Right now, the default `CharSize` is documented in the module-level docs and the docs for `Config`, but it is not documented in the `Config::new` function. I added it there as well. Is that sufficient?

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: One problem with this. I define my own `Mode` trait in this module. I think I will skip re-exporting the embedded HAL `Mode`. If you're using the embedded HAL `Mode`, you're probably using one of the constants, e.g. `MODE_0`.

hal/src/common/sercom/v2/spi_future.rs
bradleyharden: Just because I've thought about it doesn't mean it's correct =D

I'm relying on the assumption that using only safe code should do what I want/expect to happen. I think that's right in this case, but I'm not 100% sure.

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: Until you prompted me here, I hadn't actually thought that much about how I write documentation. It looks like I generally stick to two approaches. Either I lead with a descriptive noun phrase and then describe what it does, e.g.

> Type-level `enum` representing the SPI transaction length, in bytes

Or I lead with an imperative that describes what it does, e.g.

> Control the `DIPO` field as a function of the [`PadNum`] type

I think I use the latter when I don't have any meaningful additions to the noun. For example,

> Trait that controls the `DIPO` field as a function of the [`PadNum`] type

is worse, because "Trait that" adds no real value. We already know it's a trait. Instead, I opt for the imperative, because it's usually terse. I think the same reasoning applies for using the imperative in Git commit messages, which I follow.

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: Actually, I'll just export the `Mode` constants instead.

bradleyharden:
Just updated with the suggested changes. Please merge #406 before this.

If no one else has any comments or feedback, I guess this could be ready to merge.

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: I can confirm this works with `v1` pins with your Wio Terminal diff

bradleyharden:
Don't merge just yet. I realized I forgot some pieces related to the new run-time dynamic length feature.

hal/src/common/thumbv6m/sercom/v2/spi.rs
twitchyliquid64: I dont feel strongly about this :joy: I mentioned it as I'm used to the Go style guide, which suggests all type-level doc strings should be full sentences (ie: subject + verb + what) even if it reads verbose.

Your logic makes sense

hal/src/common/thumbv6m/sercom/v2/spi.rs
twitchyliquid64: SGTM

twitchyliquid64 (approved):
Thanks for this! the new trait-based API is a big step forward.

Will merge once green!

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: Interesting, I'll have to read up on it. It probably sounds crazy, but I really like to have some justification for every decision, no matter how ultimately arbitrary it is. Ultimate bikeshedding I guess :p 

I always like to read arguments for/against certain styles and that sort of thing. I find it interesting how people reason about stuff.

hal/src/common/sercom/v2/spi_future.rs
Sympatron: I still have to learn to trust the Rust compiler more than I can trust any C compiler. Years of writing C taught me to be very careful and it's hard to unlearn. :upside_down_face:

PR #406, opened 2021-02-14
Fix warnings

bradleyharden:
Could someone please make sure my proposed fix for the UART module is correct? I based the fix on other code in the module, but I don't know it that well.


TDHolmes (approved):
LGTM! Probably should wait for someone else who has merge abilities to also review though

twitchyliquid64 (approved):

PR #408, opened 2021-02-16
Make UsbBus more flexible and fix warnings throughout boards

### LINKED ISSUE #404: v1 & v2 Pin compatibility in the thumbv7em USB module
@ryankurte, I was looking through your USB code while helping with #403, and I had some potential suggestions. I probably should have followed your PR more closely and brought this up sooner. Sorry about that. Better late than never?

I also think talking about this specific example might be useful to other people, like @jaredwolff and @jbeaurivage.

First, is there a reason you define your own `private::UsbSealed` trait? If you're not aware, there is already an existing `Sealed` trait in the `crate::typelevel` module. Unless you have a specific reason for using a custom `Sealed` trait, I think it might make more sense to use the existing one.

Also, just FYI, the implementations of `UsbSealed` don't need to be in a private module. Only the trait itself needs to be hidden. As long as users can't see the `Sealed` trait, they can't implement it themselves, so they can never satisfy the requirements of any other traits based on it.

I also thought about about the advantages of various approaches to the problem of accepting both `v1::Pin`s and `v2::Pin`s. Once #401 lands, I think I have a way to improve compatibility even more.

### Solution 1

Right now, your `UsbPad*` traits are only implemented on specific pins in specific modes. Users can provide either a `v1::Pin` or a `v2::Pin`, but it must already be configured for peripheral function `H`.

### Solution 2

Instead, you could do

```rust
struct Inner {
    _dm_pad: v2::Pin<PA24, AlternateH>,
    _dp_pad: v2::Pin<PA25, AlternateH>,
    // ...
}

impl UsbBus {
    pub fn new<Dm, Dp>(
        _clock: &clock::UsbClock,
        mclk: &mut MCLK,
        dm_pad: Dm,
        dp_pad: Dp,
        _usb: USB,
    ) -> Self
    where
        Dm: Into<v2::Pin<PA24, AlternateH>>,
        Dp: Into<v2::Pin<PA25, AlternateH>>,
    {
        // ...
        let inner = Inner {
            _dm_pad: dm_pad.into(),
            _dp_pad: dp_pad.into(),
            // ...
        };
        // ...
    }
}
```

That would be *slightly* more flexible. It could still accept a `v1::Pin` in the correct mode, because `Into` is implemented between identical `v1` and `v2` pins. But this solution could also accept *any* `v2::Pin`, even if it's not in mode `AlternateH`, because `v2::Pin`s can change mode with `Into` as well (unlike `v1::Pin`s).

### Solution 3

With #401 and the `AnyPin` trait, we can actually go even further.

```rust
struct Inner {
    _dm_pad: v2::Pin<PA24, AlternateH>,
    _dp_pad: v2::Pin<PA25, AlternateH>,
    // ...
}

use crate::gpio::v2::pin::{AnyPin, SpecificPin};

impl UsbBus {
    pub fn new<Dm, Dp>(
        _clock: &clock::UsbClock,
        mclk: &mut MCLK,
        dm_pad: Dm,
        dp_pad: Dp,
        _usb: USB,
    ) -> Self
    where
        Dm: AnyPin<Id = PA24>,
        Dp: AnyPin<Id = PA25>
    {
        // ...
        let dm_pad: SpecificPin<Dm> = dm_pad.into();
        let dp_pad: SpecificPin<Dp> = dp_pad.into();
        let inner = Inner {
            _dm_pad: dm_pad.into(),
            _dp_pad: dp_pad.into(),
            // ...
        };
        // ...
    }
}
```

This option can accept *any* `v1` or `v2` pin, as long as it has the correct `PinId`. The first conversion takes the input from an "unkown" type `Dm` to `SpecificPin<Dm>`, where `SpecificPin<Dm> == v2::Pin<Dm::Id, Dm::Mode>`. After this conversion, the compiler knows that `dm_pad` is a `v2::Pin` and not some arbitrary type `Dm`. The second conversion takes it from `v2::Pin<PA24, Dm::Mode>` to `v2::Pin<PA24, AlternateH>`.

To me, this last option is the most optimal, because it can accept the most pins. I think this also serves as an example of how we can slightly modify existing modules to be compatible with both `v1` and `v2` pins. That should help ease the transition.

bradleyharden:
Change the USB modules to accept the corrects pin in any mode. v2::Pins
do not need access to the Port to change mode, so also remove the Port
argument from corresponding BSP functions.

Fix warnings throughout the board modules, so that everything builds
cleanly.

@ryankurte, this implements the proposed solution in #404. But it ended up being a little more complicated than I originally thought. I'll expand on that in #404.

Edit: closes #404

### DIFF
--- a/boards/arduino_mkrzero/examples/usb_logging.rs
+++ b/boards/arduino_mkrzero/examples/usb_logging.rs
@@ -45,7 +45,6 @@ fn main() -> ! {
             &mut peripherals.PM,
             pins.usb_n, // PA24, also usb_dm
             pins.usb_p, // PA24 also usb_dp
-            &mut pins.port,
         ));
         USB_ALLOCATOR.as_ref().unwrap()
     };
--- a/boards/arduino_mkrzero/src/lib.rs
+++ b/boards/arduino_mkrzero/src/lib.rs
@@ -11,6 +11,8 @@ pub use cortex_m_rt::entry;
 pub extern crate panic_halt;
 
 #[cfg(feature = "usb")]
+use gpio::v2::{AnyPin, PA24, PA25};
+#[cfg(feature = "usb")]
 use hal::clock::GenericClockController;
 #[cfg(feature = "usb")]
 use hal::usb::usb_device::bus::UsbBusAllocator;
@@ -123,20 +125,11 @@ pub fn usb_allocator(
     usb: pac::USB,
     clocks: &mut GenericClockController,
     pm: &mut pac::PM,
-    dm: gpio::Pa24<Input<Floating>>,
-    dp: gpio::Pa25<Input<Floating>>,
-    port: &mut Port,
+    dm: impl AnyPin<Id = PA24>,
+    dp: impl AnyPin<Id = PA25>,
 ) -> UsbBusAllocator<UsbBus> {
-    use gpio::IntoFunction;
-
     let gclk0 = clocks.gclk0();
     let usb_clock = &clocks.usb(&gclk0).unwrap();
 
-    UsbBusAllocator::new(UsbBus::new(
-        usb_clock,
-        pm,
-        dm.into_function(port),
-        dp.into_function(port),
-        usb,
-    ))
+    UsbBusAllocator::new(UsbBus::new(usb_clock, pm, dm, dp, usb))
 }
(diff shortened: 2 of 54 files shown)


bradleyharden:
Hold off on merging this. I just had a stroke of inspiration while brushing my teeth. I'll make more changes tomorrow

twitchyliquid64:
HODL'ing.

> I just had a stroke of inspiration while brushing my teeth

:joy:

bradleyharden:
Better than having a stroke while brushing your teeth.

I added `usb` to the features of 6 more boards in `crates.json`, which should improve coverage for `build-all.py`.

I also simplified pin mode conversion. I'll update #404 with the simpler way to do it.

PR #409, opened 2021-02-16
Adds RTIC Monotonic implementation based on the RTC

henrikssn:
I still need to add an example for this, but it would be good to know if this is in the right direction.

The semantics of an RTIC Monotonic is a bit weird compared to the HAL Rtc, so I made sure to take the RTC instance in initialize and drop it, to make sure nobody tries to use both at the same time. Suggestions on improving this are welcome.

See https://docs.rs/cortex-m-rtic/0.5.5/rtic/trait.Monotonic.html

Ping @TDHolmes @twitchyliquid64

### DIFF
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -79,6 +79,7 @@ max-channels = ["dma", "atsamd-hal/max-channels"]
 # Enable pins for the adalogger SD card reader
 adalogger = []
 sdmmc = ["embedded-sdmmc", "atsamd-hal/sdmmc"]
+rtic = ["atsamd-hal/rtic"]
 
 [profile.dev]
 incremental = false
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -65,8 +65,7 @@ nom = { version = "5.1", default-features= false }
 heapless = "0.5"
 
 [dev-dependencies.cortex-m-rtic]
-git = "https://github.com/rtic-rs/cortex-m-rtic"
-branch = "new_monotonic"
+version = "0.6.0-alpha.2"
 
 [features]
 # ask the HAL to enable atsamd21g support
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -49,6 +49,7 @@ optional = true
 
 [dependencies.embedded-sdmmc]
 version = "0.3"
+optional = true
 
 [dependencies.rtic-monotonic]
 version = "=0.1.0-alpha.0"
@@ -154,4 +155,4 @@ required-features = ["adalogger", "usb", "sdmmc", "unproven"]
 
 [[example]]
 name = "blinky_rtic"
-required-features = ["rtic"]
+required-features = ["rtic", "unproven"]
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -59,6 +59,7 @@ nb = "0.1"
 drogue-nom-utils = "0.1"
 nom = { version = "5.1", default-features= false }
 heapless = "0.5"
+cortex-m-rtic = "0.5.5"
 
 [features]
 # ask the HAL to enable atsamd21g support
@@ -145,3 +146,7 @@ required-features = ["usb", "unproven"]
 [[example]]
 name = "adalogger"
 required-features = ["adalogger", "usb", "sdmmc", "unproven"]
+
+[[example]]
+name = "blinky_rtic"
+required-features = ["rtic"]
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -51,8 +51,7 @@ optional = true
 version = "0.3"
 
 [dependencies.rtic-monotonic]
-git = "https://github.com/rtic-rs/rtic-monotonic"
-branch = "master"
+version = "=0.1.0-alpha.0"
 optional = true
 
 [dev-dependencies]
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -49,6 +49,10 @@ optional = true
 
 [dependencies.embedded-sdmmc]
 version = "0.3"
+
+[dependencies.rtic-monotonic]
+git = "https://github.com/rtic-rs/rtic-monotonic"
+branch = "master"
 optional = true
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


henrikssn:
Also note RTIC team is working on a new API for Monotonic, not sure if it is best to wait for that: https://github.com/rtic-rs/rtic-core/blob/new_monotonic/src/monotonic.rs

TDHolmes:
the new API seems nicer as we'd be able to internally use the existing RTC struct and not dupe code / not directly access the RTC peripheral unnecessarily. When is it expected to land?

henrikssn:
> the new API seems nicer as we'd be able to internally use the existing RTC struct and not dupe code / not directly access the RTC peripheral unnecessarily. When is it expected to land?

The design was agreed upon today and the implementation will be proposed for next alpha version of RTIC. I think it is unclear at the moment when it lands in RTIC stable.

That said, we still need a separate struct to prevent users from using the other methods of the `Rtc` struct after converting it into a Monotonic, for example resetting the counter or changing RTC mode will make the whole system behaviour undefined.

TDHolmes:
> That said, we still need a separate struct to prevent users from using the other methods of the Rtc struct after converting it into a Monotonic, for example resetting the counter or changing RTC mode will make the whole system behaviour undefined.

Agreed, but this other struct can take `Rtc` struct as value (or initialize it itself internally) and everything is good since it takes the `RTC` peripheral as value, so no other entity can mess with things. Right?

henrikssn:
> > That said, we still need a separate struct to prevent users from using the other methods of the Rtc struct after converting it into a Monotonic, for example resetting the counter or changing RTC mode will make the whole system behaviour undefined.
> 
> Couldn't this other struct can take `Rtc` struct as value (or initialize it itself internally) and everything is good since it takes the `RTC` peripheral as value, so no other entity can mess with things. Right?

This is a good idea, I pushed a new commit which implements this. I had to extend the `Rtc` struct with a `timer_mode(&mut self)` method otherwise it was pretty straight forward.

hal/src/common/rtc.rs
```diff
@@ -300,3 +335,219 @@ impl TimerParams {
...
+
+impl U32Ext for u32 {
+    fn cycles(self) -> Duration {
+        Duration { inner: self }
+    }
+}
+
+/// RtcMonotonic implements Monotonic based on the RTC peripheral for keeping
+/// track of time in RTIC
+pub struct RtcMonotonic {}
```
twitchyliquid64, line 521: Can we store `Rtc` as a non-public field here, and then provide a method to convert back from `RtcMonotonic` to `Rtc`?

henrikssn: I have now changed this so that Monotonic is implemented directly on `Rtc` (which makes this obsolete)

henrikssn:
Note that the current version of this PR requires new Monotonic from RTIC, which is only available in an experimental branch. I won't suggest merging this until it is available on crates.io

TDHolmes:
Looks like RTIC merged the new monotonic design!

henrikssn:
> Looks like RTIC merged the new monotonic design! It's not on crates.io yet, at least not an official release. Looks like v0.6.0-alpha.1 has it

Yes! It is on my todo list to update this PR with any changes needed from the released (alpha) version, we will see when I get to it... :)

sajattack:
Is this still in progress? If so please convert to draft.

hal/src/common/rtc.rs
```diff
@@ -217,7 +220,7 @@ impl Rtc<Count32Mode> {
 
     /// Returns the internal counter value.
     #[inline]
-    pub fn count32(&mut self) -> u32 {
+    pub fn count32(&self) -> u32 {
```
henrikssn, line 223: This change is needed because it is otherwise impossible to implement `try_now` in the `Clock` trait from `embedded-time`.

henrikssn:
This is ready for review (I have updated the RTIC dependency to 0.6.0-alpha.2 and rebased against master)

boards/feather_m0/Cargo.toml
```diff
@@ -47,6 +47,11 @@ optional = true
 version = "0.1"
 optional = true
 
+[dependencies.rtic-monotonic]
+git = "https://github.com/rtic-rs/rtic-monotonic"
```
sajattack, line 51: git dependencies can't be published on crates.io

Sympatron: It's probably a good idea to wait for `rtic-monotonic` and RTIC 0.6 to be out of alpha before merging this PR anyway.

LongLiveCHIEF: what about 

```
[dependencies.dwt-systick-monotonic]
version = "0.1.0-alpha.1"
```

For reference: https://github.com/rtic-rs/cortex-m-rtic/blob/master/examples/periodic.rs#L13-L17

henrikssn: Changed to pull from "v0.1.0-alpha.0"

https://crates.io/crates/rtic-monotonic

henrikssn:
Friendly ping @Sympatron @twitchyliquid64

LongLiveCHIEF:
yeah, it would _really_ help if this was pulled.  It's easy enough to update if something changes in the rtic implementation down the road, but this is a useful feature _now_.

sajattack:
I think the rt features we already have were intended for a similar purpose, but they look like a pain to use so I can't really blame you.

Sympatron:
I think @LongLiveCHIEF is right. I don't know how long RTIC will stay in alpha and it's probably useful to have now.

sajattack:
> I think @LongLiveCHIEF is right. I don't know how long RTIC will stay in alpha and it's probably useful to have now.

I wasn't disagreeing with that, but I can't merge until the branch is updated to resolve the conflict.

bradleyharden:
FYI, I asked in the RTIC channel.

> <@bradleyharden:matrix.org> korken89, how stable is the new Monotonic? I generally wouldn't want to merge something for an alpha feature in another crate. But clearly the Monotonic implementation is popular. Do you see it changing much before 0.6 release?

As I see it we won't change it's definitions, but I cannot guarantee this. For example, while we are in alpha I will continuously update embedded-time.

So monotonic impls might stop compiling if there is a large enough breaking change in embedded-time

(this will happen when I update to `0.12` in the next alpha)

However, on the `Monotonic` methods, these should not change. We have tested them a lot in different monotonic implementations and we are not seeing that there is something we are missing.

bradleyharden:
Is the implementation behind a feature gate? I think we should still make it opt-in, even if we merge

Sympatron:
> I wasn't disagreeing with that, but I can't merge until the branch is updated to resolve the conflict.

I wasn't disagreeing with you either. I was just disagreeing with my previous statement. :P

sajattack:
> Is the implementation behind a feature gate? I think we should still make it opt-in, even if we merge

Looks like it's behind the `rtic` feature

bradleyharden:
Is this ready to go @henrikssn? Seems popular enough to merge.

henrikssn:
@bradleyharden Yes, I have now rebased against master so this is ready to merge.

sajattack:
One more suggestion, can be handled before or after merge, but we should probably add the rtic feature on one of the CI builds.

PR #396, opened 2021-02-01
feather M0 Adalogger/SD card reader support

TDHolmes:
Implements support for the adalogger feather M0 variant & it's SD card reader, with an example. 

Adds pin names for the SD card chip select and card detect pins and implements the required `TimeSource` trait on the RTC for time stamping. Adds two new features, one for the pin aliases and another for enabling the [embedded-sdmmc](https://github.com/rust-embedded-community/embedded-sdmmc-rs) trait impl on RTC.

Example USB serial output of the example:
```
Card inserted!
OK!
Card size...
31914983424 bytes
volume Ok(Volume { idx: VolumeIdx(0), volume_type: Fat(FatVolume { lba_start: BlockIdx(8192), num_blocks: BlockCount(524288), name: "FEA_DAT    ", blocks_per_cluster: 8, first_data_block: BlockCount(545), fat_start: BlockCount(1), free_clusters_count: None, next_free_cluster: None, cluster_count: 65467, fat_specific_info: Fat16(Fat16Info { first_root_dir_block: BlockCount(513), root_entries_count: 512 }) }) })
Listing root directory:
        Found: DirEntry { name: ShortFileName("FEA_DAT"), mtime: Timestamp(2021-01-31 16:20:40), ctime: Timestamp(1980-01-01 00:00:00), attributes: FVA, cluster: Cluster(0), size: 0, entry_block: BlockIdx(8705), entry_offset: 0 }
        Found: DirEntry { name: ShortFileName("SPOTLI~1"), mtime: Timestamp(2021-01-31 16:20:20), ctime: Timestamp(2021-01-31 16:20:20), attributes: DH, cluster: Cluster(2), size: 0, entry_block: BlockIdx(8705), entry_offset: 96 }
        Found: DirEntry { name: ShortFileName("FSEVEN~1"), mtime: Timestamp(2021-01-31 16:20:20), ctime: Timestamp(2021-01-31 16:20:20), attributes: DH, cluster: Cluster(192), size: 0, entry_block: BlockIdx(8705), entry_offset: 160 }
        Found: DirEntry { name: ShortFileName("HELLO.TXT"), mtime: Timestamp(2021-01-31 16:20:46), ctime: Timestamp(2021-01-31 16:20:40), attributes: FA, cluster: Cluster(290), size: 12, entry_block: BlockIdx(8705), entry_offset: 192 }
volume Err(FormatError("Partition type not supported"))
volume Err(FormatError("Partition type not supported"))
[...]

PR #410, opened 2021-02-22
Fix v2 SPI bugs and improve bsp_pins macro

bradleyharden:
### SPI

The DATA register is only 16-bits for SAMD11 & SAMD21 chips, but it is
represented as a 32-bit register by the SVD. Change the SPI API to
accept u16 instead of u32.

Also correct a bug in SpiFuture, where the SS pin would never be
properly deasserted.

Thanks to @jbeaurivage for finding both of these.

### `bsp_pins`

Allow commas after each pin block, and improve the docs to point out
where to find a list of PinModes.

Thanks to @TDHolmes for trying out the `bsp_pins` macro in #396.

### DIFF
--- a/hal/src/common/gpio/v2/pin.rs
+++ b/hal/src/common/gpio/v2/pin.rs
@@ -1561,14 +1561,14 @@ define_pins!(
 /// for each [`Pin`], and it defines type aliases and constants to make it
 /// easier to work with the [`Pin`]s and [`DynPin`](super::DynPin)s.
 ///
-/// The types and constants are defined within a public module named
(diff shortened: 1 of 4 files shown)


Sympatron:
> The DATA register is only 16-bits for SAMD11 & SAMD21 chips, but it is
> represented as a 32-bit register by the SVD.

Does this mean we should patch the SVD?

bradleyharden:
@Sympatron, the DATA **field** is correctly 16 bits, but the register itself is given as 32. The datasheet says it's only 16, but they probably reserve the upper bytes of the address space for code compatibility with SAMx5x chips. At least that's the conclusion @jbeaurivage and I came to. I don't really care either way. I think their SVD choice does probably make sense here.

PR #412, opened 2021-02-22
Use type-level enums for RTC modes

henrikssn:
I tried to make this as backwards compatible as possible, therefore there are a few weird cases left:

- DisabledMode is the non-running state after calling Rtc::new
- CountDown / Periodic is only implemented for DisabledMode because they mess with the mode settings

Really, CountDown / Periodic should not change mode settings and should only be implemented for Count32 (/Count16), but that would be a breaking change.

@bradleyharden might want to take a look at the typelevel stuff

### DIFF
--- a/boards/feather_m0/.gdbinit
+++ b/boards/feather_m0/.gdbinit
@@ -1,0 +1,2 @@
+target remote :3333
+file target/thumbv6m-none-eabi/debug/examples/blinky_rtic
--- a/boards/feather_m0/examples/clock.rs
+++ b/boards/feather_m0/examples/clock.rs
@@ -46,7 +46,7 @@ fn main() -> ! {
     clocks.configure_standby(ClockGenId::GCLK3, true);
     let rtc_clock = clocks.rtc(&timer_clock).unwrap();
     let mut rtc = rtc::Rtc::new(peripherals.RTC, rtc_clock.freq(), &mut peripherals.PM);
-    rtc.clock_mode();
+    let mut rtc = rtc.clock_mode();
 
     unsafe {
         RTC = Some(rtc);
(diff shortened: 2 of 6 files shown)


bradleyharden:
I like the direction, but as you say, it's tough to maintain backwards compatibility. Maybe consider a `v2` module as well? That would allow you to implement it optimally.

From my perspective, the `v1` modules are all great, but they usually solve one particular problem. That problem is typically the primary use case, but configuration beyond that use case is limited. If you're interested, I would encourage you to write a new module that fully captures all of the possible RTC modes. I think people in the future would find it useful.

henrikssn:
> I like the direction, but as you say, it's tough to maintain backwards compatibility. Maybe consider a `v2` module as well? That would allow you to implement it optimally.
> 
> From my perspective, the `v1` modules are all great, but they usually solve one particular problem. That problem is typically the primary use case, but configuration beyond that use case is limited. If you're interested, I would encourage you to write a new module that fully captures all of the possible RTC modes. I think people in the future would find it useful.

Good point, I pushed a new version where the `DisabledMode` hacks and the likes are removed. Turned out it was almost a noop for the examples (e.g. change `Rtc::new(...)` to `Rtc::count32_mode(...)`). The current version is not complete but it can be extended without breaking backwards compatibility (i.e. just add more methods).

What is the policy here on almost backwards compatible changes?

bradleyharden:
They accepted #408, which changed a function signature.

If you're willing to fix up all the BSPs and examples, I think they'll let you change just about anything. I wanted to avoid all that work for GPIO and SPI, so I opted for the v2 module approach.

henrikssn:
Yeah there is no BSP specific code to speak of here and the 4 examples are fixed.

@twitchyliquid64 @ryankurte any opinions from your side?

twitchyliquid64 (approved):
Sorry for the delay. LGTM!

henrikssn:
> Sorry for the delay. LGTM!

Thanks! I have rebased against master, this is now ready to merge.

PR #418, opened 2021-03-20
WIP: Wio Lite W600

Gisleburt:
This is a work in progress for the [Wio Lite W600](https://www.seeedstudio.com/Wio-Lite-W600-p-4155.html)

I had a few newb questions while I work on it, so thought I'd open the PR now, hope thats ok.

1. How does my README look. I've tried to fill out the information a bit more as there were a few bits I struggled with when I first started using this library with the Arduino Nano IoT 33.
2. ~Any idea where I can learn how to fix a broken bootloader? (I've now killed two of these devices 😅 and my [questions on Seeeds forums](https://forum.seeedstudio.com/t/wio-lite-w600-no-longer-appearing-in-dev-macos/257148) have gone unanswered)~ Think I've got this covered
3. How do we feel about adding additional functionality? One of the things I wanted from this chip was the WiFi capability. I believe I should be abled to impl embedded_nal for it, but does that belong in this repository or in another one?
4. Is there an easy way to test ADC and PWM without an expensive oscilloscope?
5. I had a few questions specific to the pins:
  - Is SWDIO and SWCLK useful in software? I assume not so should I delete these?
  - The datasheet has PA27_TX_LED and PB03_RX_LED but theres no extra LEDs I can find and I can't work out what they're for (unless they mean the TX01 and RX01 pads under the charge and reset leds. Any ideas?

### DIFF
--- a/boards/wio_lite_w600/Cargo.toml
+++ b/boards/wio_lite_w600/Cargo.toml
@@ -1,0 +1,82 @@
+[package]
+name = "wio_lite_w600"
+version = "0.2.0"
+authors = ["Daniel Mason <daniel@danielmason.com"]
+description = "Board Support crate for the Wio Lite W600"
+keywords = ["no-std", "arm", "cortex-m", "embedded-hal", "arduino"]
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/atsamd-rs/atsamd"
+readme = "README.md"
+documentation = "https://atsamd-rs.github.io/atsamd/atsamd21g/wio_lite_w600/"
+edition = "2018"
+
+[dependencies]
+cortex-m = "0.6"
+embedded-hal = "0.2.3"
+nb = "0.1"
+
+[dependencies.cortex-m-rt]
+version = "0.6.12"
+optional = true
+
+[dependencies.panic-abort]
+version = "0.3"
+optional = true
+
+[dependencies.panic-halt]
+version = "0.2"
+optional = true
+
+[dependencies.panic-semihosting]
+version = "0.5"
+optional = true
+
+[dependencies.panic_rtt]
+version = "0.1"
+optional = true
+
+[dependencies.atsamd-hal]
+path = "../../hal"
+version = "0.12"
+default-features = false
+
+[dependencies.usb-device]
+version = "0.2"
+optional = true
+
+[dependencies.usbd-serial]
+version = "0.1"
+optional = true
+
+[dependencies.numtoa]
+version = "0.2"
+optional = true
+
+[dev-dependencies]
+cortex-m-semihosting = "0.3"
+
+[features]
+# ask the HAL to enable atsamd21g support
+default = ["rt", "panic_halt", "atsamd-hal/samd21g"]
+rt = ["cortex-m-rt", "atsamd-hal/samd21g-rt"]
+unproven = ["atsamd-hal/unproven"]
+use_rtt = ["atsamd-hal/use_rtt", "panic_rtt"]
+usb = ["atsamd-hal/usb", "usb-device", "usbd-serial", "numtoa"]
+panic_halt = ["panic-halt"]
+panic_abort = ["panic-abort"]
+panic_semihosting = ["panic-semihosting"]
+
+[[example]]
+name = "pwm"
+required-features = ["unproven"]
+
+[[example]]
+name = "blinky_basic"
+
+[[example]]
+name = "adc"
+required-features = ["unproven"]
+
+[[example]]
+name = "adc_usb"
+required-features = ["unproven", "usb"]
--- a/boards/wio_lite_w600/examples/blinky_basic.rs
+++ b/boards/wio_lite_w600/examples/blinky_basic.rs
@@ -19,8 +19,8 @@ fn main() -> ! {
         &mut peripherals.SYSCTRL,
         &mut peripherals.NVMCTRL,
     );
-    let mut pins = hal::Pins::new(peripherals.PORT);
-    let mut led = pins.d13.into_open_drain_output(&mut pins.port);
+    let pins = hal::Pins::new(peripherals.PORT);
+    let mut led = pins.d13.into_push_pull_output();
 
     let mut delay = Delay::new(core.SYST, &mut clocks);
 
--- a/boards/wio_lite_w600/examples/blinky_basic.rs
+++ b/boards/wio_lite_w600/examples/blinky_basic.rs
@@ -1,0 +1,56 @@
+#![no_std]
+#![no_main]
+
+extern crate wio_lite_w600 as hal;
+
+use hal::clock::GenericClockController;
+use hal::delay::Delay;
+use hal::entry;
+use hal::pac::{CorePeripherals, Peripherals};
+use hal::prelude::*;
+
+#[entry]
+fn main() -> ! {
+    let mut peripherals = Peripherals::take().unwrap();
+    let core = CorePeripherals::take().unwrap();
+    let mut clocks = GenericClockController::with_external_32kosc(
+        peripherals.GCLK,
+        &mut peripherals.PM,
+        &mut peripherals.SYSCTRL,
+        &mut peripherals.NVMCTRL,
+    );
+    let mut pins = hal::Pins::new(peripherals.PORT);
+    let mut d13 = pins.d13.into_open_drain_output(&mut pins.port);
+    let mut d12 = pins.d12.into_open_drain_output(&mut pins.port);
+    let mut d11 = pins.d11.into_open_drain_output(&mut pins.port);
+    let mut d10 = pins.d10.into_open_drain_output(&mut pins.port);
+    let mut d9 = pins.d9.into_open_drain_output(&mut pins.port);
+    let mut d6 = pins.d6.into_open_drain_output(&mut pins.port);
+    let mut delay = Delay::new(core.SYST, &mut clocks);
+
+    loop {
+        d13.set_high().unwrap();
+        delay.delay_ms(100u16);
+        d13.set_low().unwrap();
+
+        d12.set_high().unwrap();
+        delay.delay_ms(100u16);
+        d12.set_low().unwrap();
+
+        d11.set_high().unwrap();
+        delay.delay_ms(100u16);
+        d11.set_low().unwrap();
+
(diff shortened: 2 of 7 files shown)


Gisleburt:
I've swapped to the V2 GPIO API. Its much easier to use with embedded_hal

Gisleburt:
I give up. Not enough examples or documentation for a simpleton like me and I don't think I've ever get past being able to turn pins on and off. It was fun while it lasted.

bradleyharden:
@Gisleburt, if you have specific problems or questions, we can try to help.

Gisleburt:
Hi @bradleyharden, thanks for reaching out. You've already been a huge help over on discord, but I think I'm just too new to embedded programming. Over the weekend I spent 2.5hrs just trying to get I2C to work using the `i2c_master` helper function and couldn't work out where the final Port type comes from.

My newness means a lot of things that make perfect sense to anyone who knows what they're doing are lost to me (in this case Ports). I could put in a lot more time to get better but this is all taking time away from other projects. I'm used to projects taking longer than estimated but my unfamiliarity with common terminology is making that orders of magnitude worse.

The BSP itself _might_ be fine as it is (questions above not withstanding), but I don't feel confident without having tested it myself. I might come back around at some later date but for now I need a break.

bradleyharden:
Sounds good. If you decide to get back into it, I would encourage you to ask a lot of questions. Matrix is good for that kind of thing.

Gisleburt:
Oh yeah, sorry, I was on Gitter, not discord, which I imagine is plugged in to Matrix. Thanks again for that

bradleyharden:
I use both, and I've seen that there is a bridge between them for the atsamd-hal channel. I'm not sure how well it works though.

bradleyharden:
@Gisleburt, I just realized what may have been the problem. I think you said you are using `gpio::v2`, but you had trouble finding a `Port` type for the `i2c_master` function. I think you may have hit one of the pain points in the conversion from `gpio::v1` to `gpio::v2`.

`v1::Pin`s require access to the `v1::Port` type to change pin mode. But this requirement was removed in `gpio::v2`. Because the `v2::Pin`s do not require the `v1::Port`, I don't provide any way to access it in the `gpio::v2` API. However, I forgot that users would still need access to the `v1::Port` when mixing `v1::Pin`s and `v2::Pin`s.

You're now the third person to run into this problem, so I really need to do something to address it. Let me think about the best solution. `v2` was meant to simplify things, but simultaneously supporting both `v1` and `v2` ends up making things more complicated than they were before. But that should get better as we migrate.

Sympatron:
@bradleyharden Would it be possible to deprecate `v1` in the near future?
It took me some time to wrap my head around embedded rust and `atsamd-rs` when there was only v1. Partially conflicting implementations for too long might scare more people like Gisleburt away.

bradleyharden:
@Sympatron, I would love to, but there's a lot of code that uses it, both within the HAL and within the BSPs. Maybe we could update the HAL, push a release, and let existing BSPs stay on an old version? That just pushes the problem to BSP maintainers, but it's at least a bit more tractable than trying to update all the code at once.

Gisleburt:
If deprecating the api wouldn't put errors into consumers of BSP crates then I think this is a good option, however, marking all of v1 as deprecated probably would. You could possibly deprecate the macro though, as only BSP creators will use this. The instructions here would also need updating: https://github.com/atsamd-rs/atsamd/wiki/Adding-a-new-board

I've just been re-reading the docs on `bsp_pins!` while I was commenting on this and I missed that PORT is available from there, so I'll have a look to see if I can work out how to turn on I2C mode again (perhaps not with the helper) and if I can I'll re-open the PR.

bradleyharden:
@Gisleburt, careful, there is a PORT type from the corresponding PAC, and there is a Port type from the gpio::v1 module. I'm not sure which you need. If you post a snippet or point me to the specific code on GitHub, I can probably see what's going on.

Gisleburt:
I havent managed to do anything useful with this, I'm still out of my depth, but I think this works.

I'm able to write to and read back from an i2c convertor which seems like its probably right.

boards/wio_lite_w600/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+        aliases: {
+            AlternateC: Mosi
+        }
+    },
+
+    // D24_PB11_S4_SPI_SCK
+    PB11 {
+        name: d24
+        aliases: {
+            AlternateB: Sck
```
bradleyharden, line 120: I think this must be a typo. All SERCOM pins are AlternateC or AlternateD.

boards/wio_lite_w600/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+    )
+}
+
+#[cfg(feature = "usb")]
+pub fn usb_allocator(
+    usb: pac::USB,
+    clocks: &mut GenericClockController,
+    pm: &mut pac::PM,
+    dm: impl AnyPin<Id=PA24>,
+    dp: impl AnyPin<Id=PA25>,
```
bradleyharden, line 328: For consistency, you could add USB aliases for these pins in the macro above. Then this function would look like the others.

boards/wio_lite_w600/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+pub fn uart<F: Into<Hertz>>(
+    clocks: &mut GenericClockController,
+    baud: F,
+    sercom2: pac::SERCOM2,
+    pm: &mut pac::PM,
+    rx: Rx,
+    tx: Tx,
+) -> UART2<
+    hal::sercom::Sercom2Pad3<gpio::v2::PA11>,
+    hal::sercom::Sercom2Pad2<gpio::v2::PA10>,
```
bradleyharden, line 305: Unfortunately, these types and similar ones above will require minor modifications once my next PR is merged. The SERCOM pad types were more complicated than necessary, and I figured out a way to simplify I think.

boards/wio_lite_w600/src/lib.rs
Gisleburt: Ah yeah, good point. Also found these pins were labeled backwards 🤦🏻‍♂️

boards/wio_lite_w600/src/lib.rs
Gisleburt: Thats great news :) Any simplification will be great, we can wait until your changes are merged and then I can update mine.

Gisleburt: I've just seen it! So much better!

Gisleburt:
Waiting on #434

bradleyharden:
@Gisleburt, #434 was merged, but you should base this on #451. I think I managed to simplify some things based on feedback from various people and my own experiences porting feather_m0. I think #451 represents the final form of SERCOM pads, since I've now ported both feather_m0 and pygamer. I didn't encounter any major problems.

bradleyharden:
@Gisleburt, you should be good to go now that #451 has been merged. Let me know if you have any problems

Gisleburt:
I _think_ I'm done.

I really wanted to get blinky_basic to work on the attached User LED but I _think_ for some incredible reason, they've attached that to the W600 chip (same for the Config Button), which I haven't yet added code for.

I assume its ok to come back and add more examples as I learn more about the board? (Though i might take a little break from this board to play with Neo Trinkey, an ATSAMD21E based board with a lot less things going on, expect a PR for that soonish 🤞🏻 😀 )

Really love the changes to the types btw, it picked up some errors I'd missed with SPI being on the wrong alternate pin configuration.

bradleyharden:
@Gisleburt, where does this stand? Do you have a working example that you tested? If so, and you fix the formatting issues, I can probably merge.

Gisleburt:
@bradleyharden yes, this is ready. In addition to fixing rustfmt, I've also added an extra couple of lines to the readme to explain why blinky_basic doesn't use the onboard led and what to do instead.

I plan to have a bit of a break and come back and add more useful examples once I'm a _lot_ more compitent with embedded programming, if thats ok with the maintainers.

boards/wio_lite_w600/Cargo.toml
```diff
@@ -0,0 +1,82 @@
...
+[[example]]
+name = "blinky_basic"
+
+[[example]]
+name = "adc"
+required-features = ["unproven"]
+
+[[example]]
+name = "adc_usb"
+required-features = ["unproven", "usb"]
```
jbeaurivage, line 82: @Gisleburt, you might want to remove the `[[example]]` entries that don't actually exist and keeping only `blinky-basic` - otherwise it won't compile.

Gisleburt: Good spot, thanks

boards/wio_lite_w600/examples/blinky_basic.rs
```diff
@@ -0,0 +1,40 @@
...
+fn main() -> ! {
+    let mut peripherals = Peripherals::take().unwrap();
+    let core = CorePeripherals::take().unwrap();
+    let mut clocks = GenericClockController::with_external_32kosc(
+        peripherals.GCLK,
+        &mut peripherals.PM,
+        &mut peripherals.SYSCTRL,
+        &mut peripherals.NVMCTRL,
+    );
+    let pins = hal::Pins::new(peripherals.PORT);
```
jbeaurivage, line 27: I think this should be `bsp::Pins::new(peripherals.PORT)`

Gisleburt: Whoops! Fixed. Thanks for that. 😅

TDHolmes:
only little things here and there, looked good overall!

bradleyharden:
Good catches. Thanks for the thorough review

Gisleburt:
@TDHolmes @bradleyharden thanks for all the great feed, I've updated a fair bit.

A few questions remain, shall we remove the mappings for `swdio`, `swclk`, `vbat`, `tx_led`,  and `rx_led`. We can always put them back in later. I could even just comment them out and leave a comment as to why we removed them.

UART v1 is deprecated and the interface for v2 is very clean, is the convenience method worth keeping and updating or should I bin it?

Gisleburt:
Ah, README for usb_ack, one min

TDHolmes:
Ah it is an arduino board. Then just the dependency issues

PR #423, opened 2021-03-27
Arduino Nano 33 IOT WiFi NINA SPI Master

olback:


sajattack:
Might be worth adding an example too.

olback:
> Might be worth adding an example too.

I'd love to but I don't have a working example with code from crates.io.

I got [wifi-nina](https://github.com/dflemstr/wifi-nina/) to work with some not-so-nice [modifications](https://github.com/olback/wifi-nina/tree/assertless-nocheck). I have since also found [wifinina](https://github.com/fionawhim/wifinina-rs) and [ublox-short-range-rs](https://github.com/BlackbirdHQ/ublox-short-range-rs) but not tested either of them.

sajattack:
I think we can close this then if it's untested/unworking? Or draft until it works? Which would you prefer @olback?

olback:
@sajattack It _is_ working but there isn't a great wifi-nina library. I have a [working fork](https://github.com/olback/wifi-nina/tree/assertless-nocheck) of the wifi-nina library. The spi helper implemented here works just fine.

sajattack:
Yeah, if it's just a spi helper that doesn't actually pull in the rest of the wifi-nina features, I don't think it's very valuable. Someone can just set up a spi from the hal rather than the bsp.

PR #425, opened 2021-04-06
Flatten HAL and move `common` content to crate root

vccggorski:
~WiP: I published this PR in parallel to #424 for review purposes. Preferably #424 is going to be merged in first; thus, `samd{1,2}1` modules will be absent and HAL structure will become much cleaner.~
#424 is already in.

- Incorporate `common/mod.rs` into `lib.rs`
- Create a shim implementation of `common` module inside of `lib.rs` for backward compatibility
- Removed `common/mod.rs`
- Moved `common/*` to crate root

Long term goals note:
- Get rid of the shim as soon as API breakage is acceptable (BSCs + examples are reworked)
- Make other implementation-leaking modules (`thumbv6m`, `thumbv7em`) private


jessebraham (approved):
This looks good to me, thanks! I'd just like to have another set of eyes on it before merging due to the number of changes made, just to be safe.

bradleyharden:
@jessebraham, should we move forward with this?

PR #427, opened 2021-04-09
Allow updating v2 SPI settings after the peripheral has been enabled

jbeaurivage:
This PR simply adds an `update_baud` method to `sercom::v2::Spi` structs to allow updating the baudrate on the fly.

Calling the method will temporarily disable the SERCOM peripheral, as the BAUD register is enable-protected.

### DIFF
--- a/hal/src/common/thumbv6m/sercom/v2/spi.rs
+++ b/hal/src/common/thumbv6m/sercom/v2/spi.rs
@@ -1444,7 +1444,7 @@ impl<C: ValidConfig> Spi<C> {
     /// some registers are enable-protected. This may interrupt any ongoing
     /// transactions.
     #[inline]
-    pub fn update_config<F>(&mut self, update: F)
+    pub fn reconfigure<F>(&mut self, update: F)
     where
         F: FnOnce(C) -> C,
     {
--- a/hal/src/common/thumbv6m/sercom/v2/spi.rs
+++ b/hal/src/common/thumbv6m/sercom/v2/spi.rs
@@ -1170,7 +1170,14 @@ where
     /// half the GCLK frequency. The minimum baud rate is the GCLK frequency /
     /// 512. Values outside this range will saturate at the extremes.
     #[inline]
-    pub fn baud<B: Into<Hertz>>(self, baud: B) -> Self {
+    pub fn baud<B: Into<Hertz>>(mut self, baud: B) -> Self {
+        self.update_baud(baud);
+        self
+    }
+
+    /// Update the baudrate
+    #[inline]
+    fn update_baud<B: Into<Hertz>>(&mut self, baud: B) {
         let baud: Hertz = baud.into();
         let baud = (self.freq.0 / 2 / baud.0).saturating_sub(1);
         let baud = if baud <= u8::MAX as u32 {
@@ -1181,8 +1188,7 @@ where
         self.sercom
             .spi()
             .baud
-            .modify(|_, w| unsafe { w.baud().bits(baud) });
-        self
+            .modify(|_, w| unsafe { w.baud().bits(baud) })
     }
 
     /// Control the buffer overflow notification
@@ -1234,6 +1240,16 @@ where
         self.sercom.spi().ctrla.modify(|_, w| w.enable().set_bit());
         while self.sercom.spi().syncbusy.read().enable().bit_is_set() {}
         Spi { config: self }
+    }
+
+    /// Enable or disable the SERCOM peripheral, and wait for the ENABLE bit to
+    /// synchronize.
+    fn enable_peripheral(&mut self, enable: bool) {
+        self.sercom
+            .spi()
+            .ctrla
+            .modify(|_, w| w.enable().bit(enable));
+        while self.sercom.spi().syncbusy.read().enable().bit_is_set() {}
     }
 }
 
@@ -1426,6 +1442,22 @@ impl<C: ValidConfig> Spi<C> {
     #[inline]
     pub unsafe fn sercom(&self) -> &SpiSercom<C> {
         &self.config.as_ref().sercom()
+    }
+
+    /// Update the SPI bus baudrate.
+    ///
+    /// **WARNING** this method assumes that the GCLK clock hasn't been changed
+    /// since [`Spi`] has been initialized.
+    ///
+    /// Calling this method will temporarily disable the SERCOM peripheral, as
+    /// the BAUD register is enable-protected. This may interrupt any ongoing
+    /// transactions.
+    #[inline]
+    pub fn update_baud<B: Into<Hertz>>(&mut self, baud: B) {
+        let config = self.config.as_mut();
+        config.enable_peripheral(false);
+        config.update_baud(baud);
+        config.enable_peripheral(true);
     }
 
     /// Enable interrupts for the specified flags
--- a/hal/src/common/thumbv6m/sercom/v2/spi.rs
+++ b/hal/src/common/thumbv6m/sercom/v2/spi.rs
@@ -892,7 +892,7 @@ pub trait CharSize: Sealed {
 
     /// Configure the `LENGTH` register and enable the `LENGTH` counter
     #[inline]
-    fn configure(sercom: &RegisterBlock) -> () {
+    fn configure(sercom: &RegisterBlock) {
         sercom
             .spi()
             .ctrlb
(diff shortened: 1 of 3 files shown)


hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -1428,6 +1444,22 @@ impl<C: ValidConfig> Spi<C> {
...
     }
 
+    /// Update the SPI bus baudrate.
+    ///
+    /// **WARNING** this method assumes that the GCLK clock hasn't been changed
+    /// since [`Spi`] has been initialized.
+    ///
+    /// Calling this method will temporarily disable the SERCOM peripheral, as
+    /// the BAUD register is enable-protected. This may interrupt any ongoing
+    /// transactions.
```
Sympatron, line 1454: Maybe it would be a good idea to wait for `TXC` to not interrupt ongoing transactions?`
I don't think updating the baud rate will ever be that urgent that you can't at least wait for one byte/word to be sent to prevent data corruption.

hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -1428,6 +1444,22 @@ impl<C: ValidConfig> Spi<C> {
         &self.config.as_ref().sercom()
     }
 
+    /// Update the SPI bus baudrate.
+    ///
+    /// **WARNING** this method assumes that the GCLK clock hasn't been changed
+    /// since [`Spi`] has been initialized.
```
Sympatron, line 1450: Can you change the clock after it has been used by a peripheral? I thought the ownership system would prevent such a thing?

jbeaurivage: I think as it stands now, the `SercomCoreClock` isn't owned by the `Spi` struct, so it can be changed under an enabled `Spi`.

hal/src/common/thumbv6m/sercom/v2/spi.rs
jbeaurivage: That's reasonable

jbeaurivage:
After discussing with @bradleyharden, we figured it made more sense to provide a closure that runs with the Spi peripheral disabled, and allows the underlying `Config` to be updated on-the-fly.

This implementation wraps the `config` field of `Spi` into an `Option`, because all methods that modify a `Config` take `self`. Therefore we need to temporarily move `self.config` out of `Spi` to be able to modify it. This is a bit unwieldy and leads to a LOT of `unwrap`s.

Another proposition would be to provide a set of methods that take `self`, and another set that takes `&mut self` to preserve the ability to use the builder pattern on `Config` structs, and also be able to mutate a `Config` without needing to move it.

sajattack:
Does @bradleyharden want to review this or is it ready for merge?

bradleyharden:
@sajattack, I will try to look at it later today

jbeaurivage:
@sajattack I still have a couple commits to make on this. @bradleyharden found a way to get rid of the `Option` using the `replace_with` crate. Also I still haven't decided if we should wait for a `TXC` before disabling the SERCOM.

hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -1428,6 +1438,28 @@ impl<C: ValidConfig> Spi<C> {
         &self.config.as_ref().sercom()
     }
 
+    /// Update the SPI configuration.
+    ///
+    /// Calling this method will temporarily disable the SERCOM peripheral, as
+    /// some registers are enable-protected. This may interrupt any ongoing
+    /// transactions.
+    #[inline]
+    pub fn update_config<F>(&mut self, update: F)
```
bradleyharden, line 1447: What do you think about `reconfigure` as a name?

hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -1428,6 +1438,28 @@ impl<C: ValidConfig> Spi<C> {
...
+    where
+        F: FnOnce(C) -> C,
```
bradleyharden, line 1449: I think I would prefer `F: FnOnce(SpecificConfig<C>) -> SpecificConfig<C>,`. It probably doesn't matter that much, because you're probably only calling this function when you have a concrete type for `C`, i.e. in end-user code. But `SpecificConfig<C>` tells the compiler that it will definitely receive a `Config` struct, not some unknown, arbitrary type `C`.

hal/src/common/thumbv6m/sercom/v2/spi.rs
```diff
@@ -1235,6 +1235,16 @@ where
         while self.sercom.spi().syncbusy.read().enable().bit_is_set() {}
         Spi { config: self }
     }
+
+    /// Enable or disable the SERCOM peripheral, and wait for the ENABLE bit to
+    /// synchronize.
+    fn enable_peripheral(&mut self, enable: bool) {
```
bradleyharden, line 1241: I have plans to do some internal refactoring with the register-setting code in this module. I've settled on what I think is a slightly better approach. Fortunately, this function is private, so it doesn't affect my plans. I can keep the public API the same.

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: The chips allow it. I'm not sure if the current `clock` module would allow it, but I intentionally didn't integrate with the existing `clock` module. I have a new `clock::v2` that is much more flexible. It can handle pretty much any clock configuration you want. The code is essentially done, but I need to find time to write the documentation. Part of that PR will be an update to `sercom::v2::spi` to integrate with it.

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: This is a good question. I think my answer in this case is no, but while I have your attention, I'd like to ask your opinions on a related question.

In general, the `Spi` struct does not track the transaction status. Without tracking that status, I don't see how you could implement this suggestion. And personally, I don't think the use case is common enough to justify adding the overhead of tracking transaction status. I think it's something users should do for themselves, when necessary.

Note, however, that there is one case where I'm already tracking transaction status, the implementation of `serial::Read` for read-only SPI masters (i.e. MISO but no MOSI). There, I track the transaction status so I know when to start a transaction vs. wait for the response. It would be conceivable to expand that implementation and always track the state, but I don't particularly like it. Thoughts?

Separately, I would like to ask you about a philosophical question regarding the `TXC` flag and the `DATA` register. When I wrote this module, I purposefully made the `read_data` and `write_data` methods `unsafe`, because directly reading the `DATA` register could violate some assumptions of the `embedded-hal` trait implementations. However, I wasn't very principled, because I allowed the `clear_flags` method to be marked safe, even though clearing the `TXC` flag could potentially disrupt the implementation of `serial::Write::flush`.

I haven't really come to a coherent conclusion on this point. Strictly speaking, I think there is an argument to be made that the `read_data` and `write_data` methods are not memory unsafe. Calling them incorrectly could lead to a logic error, but I don't see any immediate way to produce memory unsafety. Do you agree? Should I mark these functions safe? Or, alternatively, do you think `clear_flags` should be marked `unsafe`, because it could disrupt an `embedded-hal` trait? Or is the status quo sufficient?

jbeaurivage: I agree that waiting for a TXC might be a little overkill considering the use cases of this API. I don't really see the point of drastically changing the implementation just to support this functionality.

On the topic of marking `read_data` and `write_data` as `unsafe`, I would say it's definitely a grey area. It's not exactly memory unsafe, and `unsafe` shouldn't be abused to cover logic errors, but it can be seen as a bit of a pseudo-thread-safety problem.

Manually clearing the flags (or manually reading/writing the DATA register, which manipulates the flags), is essentially mutating a `static mut` variable, which can be mutated asynchronously by different contexts (main context, interrupt handlers, or the hardware itself). For that reason, I think it's sound to leave these functions as `unsafe`.

hal/src/common/thumbv6m/sercom/v2/spi.rs
jbeaurivage: That doesn't seem to compile, because the `Spi` struct expects `self.config` to be of type `C`, and so `replace_with` fails to compile when trying to replace some type `C` with a specific `Config`.

hal/src/common/thumbv6m/sercom/v2/spi.rs
jbeaurivage: Sounds good. I'll leave it as is and let you refactor in a subsequent PR.

jbeaurivage:
@sajattack, I think this is ready for merge.

hal/src/common/thumbv6m/sercom/v2/spi.rs
bradleyharden: @jbeaurivage, yes this is a bit tricky. I mention a similar situation in the [docs](https://atsamd-rs.github.io/docs/samd51n/thumbv7em-none-eabihf/doc/atsamd_hal/common/gpio/v2/pin/trait.AnyPin.html#generic-anypins) for `AnyPin`.

The compiler doesn't understand that when `C: AnyConfig`, `C == SpecificConfig<C>>`. From the compiler's perspective, `C` is some unknown, generic type, while `SpecificConfig<C>` is a `Config` built from associated types. However, `C` implements `AnyConfig`, which implies the bounds `C: From<SpecificConfig<C>>` and `C: Into<SpecificConfig<C>>`, so you can trivially "convert" between them, which is always a NOP.

The solution is

```rust
replace_with::replace_with(
    &mut self.config,
    || old_config,
    |config| update(config.into()).into()
);
```

I just checked, and this compiles.

bradleyharden:
@jbeaurivage, see my latest comment above.

jbeaurivage:
Good to go now!

PR #431, opened 2021-04-22
Add support for the P1AM-100 board

quentinmit:
I've marked this as a draft because there's a number of things I'm unsure about. Can someone who is more familiar with the new v2 APIs please tell me if I'm making appropriate use of them?

I'm not sure if I'm properly using the new pin aliases, for instance. It seems poor that I need to use the pin's name in combination with the alias (e.g. `let ss: hal::BaseSlaveSelect = pins.d18.into();`). Also the `uart` function has an explicit `PB22` and `PB23` because I couldn't figure out how to turn an alias into a `PinId`.

TODO:
- [ ] Implement helper for base controller SPI
- [ ] Implement helper for SD card IO
- [ ] Write example for UART and test

### DIFF
--- a/boards/p1am_100/README.md
+++ b/boards/p1am_100/README.md
@@ -1,0 +1,29 @@
+# Adafruit Feather M0 Board Support Crate
+
+This crate provides a type-safe API for working with the [Adafruit Feather M0
+board](https://www.adafruit.com/product/2772).
+
+## Prerequisites
+* Install the cross compile toolchain `rustup target add thumbv6m-none-eabi`
+* Install [cargo-hf2 the hf2 bootloader flasher tool](https://crates.io/crates/cargo-hf2) however your platform requires
+
+## Uploading an example
+Check out the repository for examples:
+
+https://github.com/atsamd-rs/atsamd/tree/master/boards/feather_m0/examples
+
+* Be in this directory `cd boards/feather_m0`
+* Put your device in bootloader mode usually by hitting the reset button twice.
+* Build and upload in one step
+```
+$ cargo hf2 --release --example blinky_basic
+    Finished release [optimized + debuginfo] target(s) in 0.19s
+    Searching for a connected device with known vid/pid pair.
+    Trying  Ok(Some("Adafruit Industries")) Ok(Some("PyBadge"))
+    Flashing "/Users/User/atsamd/boards/feather_m0/target/thumbv7em-none-eabihf/release/examples/blinky_basic"
+    Finished in 0.079s
+$
+```
+
+Note that some older Feather M0 boards do not come with support for HF2. For these boards,
+you can upload using the `bossa` tool as described in the [atsamd crate documentation](https://github.com/atsamd-rs/atsamd#getting-code-onto-the-device-gemma-m0).
--- a/boards/p1am_100/examples/pwm.rs
+++ b/boards/p1am_100/examples/pwm.rs
@@ -1,0 +1,47 @@
+#![no_std]
+#![no_main]
+
+extern crate cortex_m_rt;
+extern crate p1am_100 as hal;
+extern crate panic_halt;
+
+use hal::clock::GenericClockController;
+use hal::delay::Delay;
+use hal::pac::{CorePeripherals, Peripherals};
+use hal::prelude::*;
+use hal::pwm::Pwm4;
+
+use cortex_m_rt::entry;
+
+#[entry]
+fn main() -> ! {
+    let mut peripherals = Peripherals::take().unwrap();
+    let core = CorePeripherals::take().unwrap();
+
+    let mut clocks = GenericClockController::with_external_32kosc(
+        peripherals.GCLK,
+        &mut peripherals.PM,
+        &mut peripherals.SYSCTRL,
+        &mut peripherals.NVMCTRL,
+    );
+    let mut delay = Delay::new(core.SYST, &mut clocks);
+    let mut pins = hal::Pins::new(peripherals.PORT);
+
+    let _led = pins.led.into_mode::<hal::gpio::v2::AlternateE>();
+
+    let gclk0 = clocks.gclk0();
+    let mut pwm4 = Pwm4::new(
+        &clocks.tc4_tc5(&gclk0).unwrap(),
+        1.khz(),
+        peripherals.TC4,
+        &mut peripherals.PM,
+    );
+    let max_duty = pwm4.get_max_duty();
+
+    loop {
+        pwm4.set_duty(max_duty / 2);
+        delay.delay_ms(1000u16);
+        pwm4.set_duty(max_duty / 8);
+        delay.delay_ms(1000u16);
+    }
+}
--- a/boards/p1am_100/memory.x
+++ b/boards/p1am_100/memory.x
@@ -1,0 +1,8 @@
+MEMORY
+{
+  /* Leave 8k for the default bootloader on the Feather M0 */
+  FLASH (rx) : ORIGIN = 0x00000000 + 8K, LENGTH = 256K - 8K
+  RAM (xrw)  : ORIGIN = 0x20000000, LENGTH = 32K
+}
+_stack_start = ORIGIN(RAM) + LENGTH(RAM);
+
--- a/boards/p1am_100/src/lib.rs
+++ b/boards/p1am_100/src/lib.rs
@@ -134,14 +134,14 @@ bsp_pins!(
     PA22 { name: d0 }
     PA23 { name: d1 }
     PA24 {
-        name: pa24,
+        name: usb_dm,
         aliases: {
             #[cfg(feature = "usb")]
             AlternateG: UsbDm
         }
     }
     PA25 {
-        name: pa25,
+        name: usb_dp,
         aliases: {
             #[cfg(feature = "usb")]
             AlternateG: UsbDp
--- a/boards/p1am_100/src/lib.rs
+++ b/boards/p1am_100/src/lib.rs
@@ -1,0 +1,298 @@
+#![no_std]
+
+extern crate atsamd_hal as hal;
+
+#[cfg(feature = "rt")]
+extern crate cortex_m_rt;
+#[cfg(feature = "rt")]
+pub use cortex_m_rt::entry;
+
+use hal::prelude::*;
+use hal::*;
+
+pub use hal::common::*;
+
+pub use hal::target_device as pac;
+
+use gpio::{Floating, Input, PfC, Port};
+use hal::clock::GenericClockController;
+use hal::sercom::{I2CMaster3, PadPin, SPIMaster4, UART0};
+use hal::time::Hertz;
+
+#[cfg(feature = "usb")]
+use gpio::v2::{AnyPin, PA24, PA25};
+#[cfg(feature = "usb")]
+use hal::usb::usb_device::bus::UsbBusAllocator;
+#[cfg(feature = "usb")]
+pub use hal::usb::UsbBus;
+
+bsp_pins!(
+    PA02 {
+        name: d15,
+        aliases: {
+            AlternateB: A0
+        }
+    }
+    PA04 {
+        name: d18,
+        aliases: {
+            AlternateB: A3,
+            PushPullOutput: BaseSlaveSelect
+        }
+    }
+    PA05 {
+        name: d19,
+        aliases: {
+            AlternateB: A4,
+            PullUpInput: BaseSlaveAck
+        }
+    }
+    PA06 {
+        name: d20,
+        aliases: {
+            AlternateB: A5
+        }
+    }
+    PA07 {
+        name: d21,
+        aliases: {
+            AlternateB: A6,
+            AlternateG: I2sSerialData
+        }
+    }
+    PA08 {
+        name: d11,
+        aliases: {
+            AlternateC: SdaPin
+        }
+    }
+    PA09 {
+        name: d12,
+        aliases: {
+            AlternateC: SclPin
+        }
+    }
+    PA10 {
+        name: d2,
+        aliases: {
+            AlternateG: I2sSerialClock
+        }
+    }
+    PA11 { name: d3 }
+    PA12 {
+        name: pa12,
+        aliases: {
+            AlternateC: SdSdo
+        }
+    }
+    PA13 {
+        name: pa13,
+        aliases: {
+            AlternateC: SdSck
+        }
+    }
+    PA14 {
+        name: pa14,
+        aliases: {
+            PushPullOutput: SdSlaveSelect
+        }
+    }
+    PA15 {
+        name: pa15,
+        aliases: {
+            AlternateC: SdSdi
+        }
+    }
+    PA16 {
+        name: d8,
+        aliases: {
+            AlternateC: Spi0Sdo
+        }
+    }
+    PA17 {
+        name: d9,
+        aliases: {
+            AlternateC: Spi0Sck
+        }
+    }
+    PA18 {
+        name: pa18,
+        aliases: {
+            /// Host Enable, drive high to switch into USB host mode
+            #[cfg(feature = "usb")]
+            PushPullOutput: UsbId
+        }
+    }
+    PA19 {
+        name: d10,
+        aliases: {
+            AlternateC: Spi0Sdi
+        }
+    }
+    PA20 { name: d6 }
+    PA21 { name: d7 }
+    PA22 { name: d0 }
+    PA23 { name: d1 }
+    PA24 {
+        name: pa24,
+        aliases: {
+            #[cfg(feature = "usb")]
+            AlternateG: UsbDm
+        }
+    }
+    PA25 {
+        name: pa25,
+        aliases: {
+            #[cfg(feature = "usb")]
+            AlternateG: UsbDp
+        }
+    }
+    PA27 {
+        name: pa27,
+        aliases: {
+            PullUpInput: SdCardDetect
+        }
+    }
+    PA28 {
+        name: switch,
+        aliases: {
+            PullUpInput: Switch
+        }
+    }
+    PB02 {
+        name: d16,
+        aliases: {
+            AlternateB: A1
+        }
+    }
+    PB03 {
+        name: d17,
+        aliases: {
+            AlternateB: A2
+        }
+    }
+    PB08 {
+        name: led,
+        aliases: {
+            PushPullOutput: Led
+        }
+    }
+    PB09 {
+        name: pb09,
+        aliases: {
+            PushPullOutput: BaseEnable,
+            AlternateB: AdcBattery
+        }
+    }
+    PB10 { name: d4 }
+    PB11 { name: d5 }
+    PB22 {
+        name: d14,
+        aliases: {
+            AlternateD: UartTx
+        }
+    }
+    PB23 {
+        name: d13,
+        aliases: {
+            AlternateD: UartRx
+        }
+    }
+);
+
+/// Convenience for setting up the labelled SPI peripheral.
+/// This powers up SERCOM4 and configures it for use as an
+/// SPI Master in SPI Mode 0.
+pub fn spi_master<F: Into<Hertz>>(
+    clocks: &mut GenericClockController,
+    bus_speed: F,
+    sercom4: pac::SERCOM4,
+    pm: &mut pac::PM,
+    sck: gpio::Pb11<Input<Floating>>,
+    mosi: gpio::Pb10<Input<Floating>>,
+    miso: gpio::Pa12<Input<Floating>>,
+    port: &mut Port,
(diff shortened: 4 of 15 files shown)


boards/p1am_100/README.md
```diff
@@ -0,0 +1,29 @@
...
+    Finished release [optimized + debuginfo] target(s) in 0.19s
+    Searching for a connected device with known vid/pid pair.
+    Trying  Ok(Some("Adafruit Industries")) Ok(Some("PyBadge"))
+    Flashing "/Users/User/atsamd/boards/feather_m0/target/thumbv7em-none-eabihf/release/examples/blinky_basic"
+    Finished in 0.079s
+$
+```
+
+Note that some older Feather M0 boards do not come with support for HF2. For these boards,
+you can upload using the `bossa` tool as described in the [atsamd crate documentation](https://github.com/atsamd-rs/atsamd#getting-code-onto-the-device-gemma-m0).
```
jbeaurivage, line 29: Consider changing the README to refer to your specific board

boards/p1am_100/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+        bus_speed.into(),
+        hal::hal::spi::Mode {
+            phase: hal::hal::spi::Phase::CaptureOnFirstTransition,
+            polarity: hal::hal::spi::Polarity::IdleLow,
+        },
+        sercom4,
+        pm,
+        (miso.into_pad(port), mosi.into_pad(port), sck.into_pad(port)),
+    )
+}
```
jbeaurivage, line 232: You may want to consider using the new `sercom::v2::spi` driver. Your function could look something like:
``` rust
use sercom::v2::spi;
pub type Spi = spi::Spi<spi::Config<SpiPads, spi::Master, spi::EightBit>>;

pub fn spi_master<F: Into<Hertz>>(
    clock: &mut GenericClockController,
    sercom1: pac::SERCOM4,
    pm: &pac::PM,
    baud: F,
    mosi: Pin<PB10, InputFloating>,
    miso: Pin<PA12, FloatingInput>,
    sclk: Pin<PB11, FloatingInput>,
) -> Spi {
    let core_clock = &clocks.sercom4_core(&clocks.gclk0()).unwrap();
    let pads = sercom::v2::spi::Pads::<Sercom4>::new()
        .sclk::<Pad3, _>(sclk)
        .data_in::<Pad0, _>(miso)
        .data_out::<Pad2, _>(mosi);
    spi::Config::new(pm, sercom1, pads, core_clock.freq())
        .baud(baud)
        .enable()
}
```

bradleyharden: I recently discovered a way to simplify `spi::Pads`. It's almost ready for a PR, but not quite yet. If you decide to use `v2::spi`, it should only require minor changes.

bradleyharden:
I think I've responded to some of these point on Matrix, but I'll do it again here, for completeness.

> It seems poor that I need to use the pin's name in combination with the alias (e.g. `let ss: hal::BaseSlaveSelect = pins.d18.into();`)

This is a limitation of the current `Pins` struct approach. Each pin is a field within the `Pins` struct, and fields can only have one name. There are alternatives to this approach, but they have run-time costs. An obvious one that comes to mind is this. Let the `Pins` struct contain `Option<Pin>` values. Then alias functions could try to take the corresponding pin. If has already been taken by a different alias, then it would panic.

> Also the uart function has an explicit PB22 and PB23 because I couldn't figure out how to turn an alias into a PinId.

We could probably improve this. I guess it would make sense to have aliases for both the `PinId` and the `Pin` itself. Maybe something like this?

```rust
pub type UartRx = Pin<PB23, AlternateD>;
pub type UartRxId = PB23;
```

By the way, it is possible to go from `UartRx` to `PB23`, but it involves a bit of type-level magic and the `AnyPin` trait. The type `<UartRx as AnyPin>::Id` is `PB23`. I don't think it exists right now, but I could add an alias to make that less magical.

```rust
pub type SpecificPinId<P> = <P as AnyPin>::Id;
```

Then `SpecificPinId<UartRx>` would return `PB23`.

bradleyharden:
Let me see if I can find time to go through this in more depth and make it as clean and simple as possible. I think we're currently missing a good example of how to use `gpio::v2`. It's still quite new.

boards/p1am_100/src/lib.rs
bradleyharden: With the latest (unpublished) update to `v2::spi`, you could do this:

```rust
use sercom::v2::spi;
pub type SpiPads = spi::Pads<Sercom4, (Pad0, PB12), (Pad2, PB10), (Pad3, PB11)>;
pub type Spi = spi::Spi<spi::Config<SpiPads>>; // Master & EightBit are the defaults

pub fn spi_master(
    clock: &mut GenericClockController,
    sercom4: pac::SERCOM4,
    pm: &pac::PM,
    baud: impl Into<Hertz>,
    mosi: impl AnyPin<Id = PB10>,
    miso: impl AnyPin<Id = PB12>,
    sclk: impl AnyPin<Id = PB11>,
) -> Spi {
    let core_clock = &clocks.sercom4_core(&clocks.gclk0()).unwrap();
    let pads = spi::Pads::default()
        .sclk(sclk)
        .data_in(miso)
        .data_out(mosi);
    spi::Config::new(pm, sercom4, pads, core_clock.freq())
        .baud(baud)
        .enable()
}
```

quentinmit:
> I think I've responded to some of these point on Matrix, but I'll do it again here, for completeness.
> 
> > It seems poor that I need to use the pin's name in combination with the alias (e.g. `let ss: hal::BaseSlaveSelect = pins.d18.into();`)
> 
> This is a limitation of the current `Pins` struct approach. Each pin is a field within the `Pins` struct, and fields can only have one name. There are alternatives to this approach, but they have run-time costs. An obvious one that comes to mind is this. Let the `Pins` struct contain `Option<Pin>` values. Then alias functions could try to take the corresponding pin. If has already been taken by a different alias, then it would panic.

That has the downside of being runtime instead of compile-time. I've already found out how hard it is to debug panics on real hardware without an OCD probe.

> > Also the uart function has an explicit PB22 and PB23 because I couldn't figure out how to turn an alias into a PinId.
> 
> We could probably improve this. I guess it would make sense to have aliases for both the `PinId` and the `Pin` itself. Maybe something like this?
> 
> ```rust
> pub type UartRx = Pin<PB23, AlternateD>;
> pub type UartRxId = PB23;
> ```

I prefer the below, I think, as long as it's documented.

> By the way, it is possible to go from `UartRx` to `PB23`, but it involves a bit of type-level magic and the `AnyPin` trait. The type `<UartRx as AnyPin>::Id` is `PB23`. I don't think it exists right now, but I could add an alias to make that less magical.
> 
> ```rust
> pub type SpecificPinId<P> = <P as AnyPin>::Id;
> ```
> 
> Then `SpecificPinId<UartRx>` would return `PB23`.

Ahh, that's great. I'll switch to that.

I don't understand why it wants a PinId and not a Pin, though. Doesn't it already know that it needs a `Pin<_, AlternateD>`? Wouldn't taking a `Pin` give you compile-time type safety to make sure that the pin is configured in the correct alternate mode (and that no one else is trying to use it)?

bradleyharden:
> I don't understand why it wants a `PinId` and not a `Pin`, though. Doesn't it already know that it needs a `Pin<_, AlternateD>`? Wouldn't taking a `Pin` give you compile-time type safety to make sure that the pin is configured in the correct alternate mode (and that no one else is trying to use it)?

The SERCOM pad types are more complicated than they need to be right now. Part of my update to `v2::spi` also addresses differences in the `Pad` types between `v1` and `v2`. Explaining the current situation is more effort than it is worth when I have a way to simplify it very soon. However, the existence of both `v1` and `v2` is still making things more complicated than they need to be.

To answer your other question, there is some type-level magic going on that ensures the pin is configured in the correct mode, even if `AlternateD` is not present in the type signature. This is currently handled by the `sercom::v2::pads::Map` trait, but this is precisely the trait I think is too complicated. It will be going away soon.

boards/p1am_100/src/lib.rs
quentinmit: I think I've cleaned this up significantly, and using the proper pin aliases to boot (instead of `AnyPin`). I was able to avoid duplicating any information in the convenience functions that was already present in the `bsp_pins!` invocation.

boards/p1am_100/examples/pwm.rs
```diff
@@ -0,0 +1,47 @@
...
+    let mut clocks = GenericClockController::with_external_32kosc(
+        peripherals.GCLK,
+        &mut peripherals.PM,
+        &mut peripherals.SYSCTRL,
+        &mut peripherals.NVMCTRL,
+    );
+    let mut delay = Delay::new(core.SYST, &mut clocks);
+    let mut pins = hal::Pins::new(peripherals.PORT);
+
+    let _led = pins.led.into_mode::<hal::gpio::v2::AlternateE>();
```
bradleyharden, line 30: Is this leftover from something else? I don't think it makes sense for an LED to use peripheral function E.

boards/p1am_100/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+use hal::sercom::v2::spi;
+use hal::sercom::{I2CMaster0, PadPin, UART5};
+use hal::time::Hertz;
+
+#[cfg(feature = "usb")]
+use hal::usb::usb_device::bus::UsbBusAllocator;
+#[cfg(feature = "usb")]
+pub use hal::usb::UsbBus;
+
+bsp_pins!(
```
bradleyharden, line 29: While porting the feather_m0, I realized that this macro needed some tweaking. I added the ability to provide doc strings on the names as well as the aliases.

boards/p1am_100/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    PB23 {
+        name: d13,
+        aliases: {
+            AlternateD: UartRx
+        }
+    }
+);
+
+const BASE_CONTROLLER_FREQ: Hertz = Hertz(1000000);
+// FIXME: const BASE_CONTROLLER_SPI_MODE: dyn spi::Mode = &spi::MODE_2;
```
bradleyharden, line 204: I see your confusion here. There are two different types of "mode". There's the SPI mode, which is a combination of CPOL and CPHA. Then there's the operating mode, of which there are three, `Master`, `MasterHWSS` and `Slave`. Check the documentation for the `Mode` trait for more details. If the documentation is lacking (which it probably is), let me know what you don't understand and I can expand.

boards/p1am_100/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+/// SPI0 has the P1AM base controller connected.
+/// This powers up SERCOM1 and configures it for talking to the
+/// base controller.
+pub fn base_controller_spi(
+    clocks: &mut GenericClockController,
+    sercom1: pac::SERCOM1,
+    pm: &mut pac::PM,
+    sck: Spi0Sck,
+    mosi: Spi0Mosi,
+    miso: Spi0Miso,
```
bradleyharden, line 216: This is a perfectly valid way to implement it. I just wanted to point out my logic for using `AnyPin`.

I once read a blog post or something that had advice about the types in a function signature. It argued that the input parameters should be typed as generically as possible, while the return values should be typed as specifically as possible. The idea is that you impose as few requirements as possible on the caller, and you provide as much information as possible about your result.

By using aliases here, you are requiring the pin to already be in the correct mode. That's fine, and it actually may produce easier to understand code. But it also means the caller is forced to convert the pin mode manually. With `impl AnyPin<Id = PXXX>`, the caller can provide a pin in any mode, so long as it is the right `PinId`.

boards/p1am_100/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+) -> sercom::v2::spi::Spi<impl spi::ValidConfig> {
```
bradleyharden, line 217: A similar argument applies here. By using `impl ValidConfig`, you hide the exact `Pads`, `Mode` and `CharSize` from the caller. That's probably fine for most cases, but if you ever wanted to store the SPI struct in a `static` variable, you would be **required** to fully name the type. Just FYI.

I understand why you did it this way though. The `Pads` type in the current version of `v2::spi` is overly complex. I provided the `spi_pads_alias` macro to help with that. But I later realized that I could refactor the code to make the `Pads` struct look almost exactly like the macro. In the latest (unpublished) version, I updated the `Pads` struct and got rid of the macro.

boards/p1am_100/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+pub fn i2c_master<F: Into<Hertz>>(
+    clocks: &mut GenericClockController,
+    bus_speed: F,
+    sercom0: pac::SERCOM0,
+    pm: &mut pac::PM,
+    sda: Sda,
+    scl: Scl,
+    port: &mut Port,
+) -> I2CMaster0<
+    impl hal::sercom::AnyPad<Sercom = pac::SERCOM0>,
```
bradleyharden, line 238: Again, I think I understand why you did it this way. The difference between the `v1` and `v2` `Pad` types was confusing, and the `Map` trait was overly complex. In the latest iteration, the two versions are totally separate types.

If you want to update to `v2` `Pad` types right now, you could use

```rust
use hal::sercom::v2::{Pad, Sercom0, Pad0, Pad1};
// ...
I2CMaster0<
    Pad<Sercom0, Pad0, <Sda as AnyPin>::Id>,
    Pad<Sercom0, Pad0, <Scl as AnyPin>::Id>,
>
```

In the latest iteration, I also added a type alias `SpecificPinId<P>` to use instead of the `<P as AnyPin>::Id` syntax.

boards/p1am_100/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+        sercom0,
+        pm,
+        <hal::gpio::v1::Pin<_, _>>::from(sda).into_pad(port),
+        <hal::gpio::v1::Pin<_, _>>::from(scl).into_pad(port),
+    )
+}
+
+pub type Uart = UART5<
+        hal::sercom::Sercom5Pad3<<UartRx as AnyPin>::Id>,
+    hal::sercom::Sercom5Pad2<<UartTx as AnyPin>::Id>,
```
bradleyharden, line 254: FYI, this only works right now because the `v1` and `v2` pad types are mixed. The `Sercom5Pad3`, etc. aliases are for the `v1` pad types. They take a complete pin in their type parameter, e.g. `Sercom5Pad3<UartRx>`. The `v2` pad types only need the `PinId`. They look like this: `Pad<Sercom5, Pad3, UartRxPinId>`.

In the future update, your current code won't work. But it should be an easy fix.

Once the `v1` pad types are removed, maybe I'll recast the `SercomXPadY` type aliases to be `v2` pads. I could also consider adding a macro to automatically create `Pad` aliases, just like `Pin` aliases with `bsp_pins!`.

boards/p1am_100/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    )
+}
+
+#[cfg(feature = "usb")]
+pub fn usb_allocator(
+    usb: pac::USB,
+    clocks: &mut GenericClockController,
+    pm: &mut pac::PM,
+    dm: impl AnyPin<Id = <UsbDm as AnyPin>::Id>,
+    dp: impl AnyPin<Id = <UsbDp as AnyPin>::Id>,
```
bradleyharden, line 286: Why not use `UsbDm` and `UsbDp` here, like above? The `impl AnyPin` approach is only there to accept the same pin in any `PinMode`. But in this case, it seems like consistency would be more important.

hal/src/common/thumbv6m/sercom/v1/uart.rs
```diff
@@ -286,6 +286,34 @@ macro_rules! uart {
...
+                            w
+                        });
+                    }
+                }
+
+                pub fn flags(&self) -> crate::target_device::sercom0::usart::status::R {
+                    unsafe {
+                        self.usart().status.read()
+                    }
+                }
```
bradleyharden, line 316: I'm not a huge fan of this approach. I'd rather keep the direct calls to PAC code hidden and use something like `biitflags` or some other crate to implement the flags. That's the approach I took in `v2::spi`.

But at the same time, this is a `v1` module, and someone may come along and create a more flexible `v2` module. We can probably be more discerning about a new API and more liberal with the changes we make to the existing API.

boards/p1am_100/examples/pwm.rs
quentinmit: My intent was to use PWM to throb the LED between different brightnesses. I believe that's function E?

boards/p1am_100/src/lib.rs
quentinmit: I greatly prefer being explicit with the caller (and it's just a `.into()` call for them). Especially since the pin names don't match their function, so the `.into()` is an explicit clue that something is being converted.

boards/p1am_100/src/lib.rs
quentinmit: I don't want to have to write the pad number in this type. I want the compiler to infer it from the sercom and the pin.

hal/src/common/thumbv6m/sercom/v1/uart.rs
quentinmit: This is the exact API that's already exposed by the `thumbv7em` version of the package. I don't disagree that it's non-ideal but I think consistency is more important.

boards/p1am_100/src/lib.rs
quentinmit: Ahh, I need to explicitly use `embedded_hal::spi::Mode` as the type instead of `hal::spi::Mode`, even though the constant is `hal::spi::MODE_2`. Boy, that's confusing.

boards/p1am_100/src/lib.rs
quentinmit: Again I don't want to have to specify these pad numbers here at all, but I couldn't figure out how to avoid it since UART5 wants a `v1::Pad`.

boards/p1am_100/src/lib.rs
quentinmit: Good point; I wrote this first before I figured out how to use a pin alias. Fixed.

quentinmit:
I think this BSP is complete enough now that it's ready for final review for merging. PTAL, thanks!

boards/p1am_100/memory.x
```diff
@@ -0,0 +1,8 @@
+MEMORY
+{
+  /* Leave 8k for the default bootloader on the Feather M0 */
```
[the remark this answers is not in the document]
quentinmit: I added it to `crates.json`; is there anything else I need to do to include it in the CI pipeline?

twitchyliquid64:
As long as your crates dot json entry has a build command you should be good to go :)

boards/p1am_100/examples/pwm.rs
bradleyharden: Ah, I think I remember being confused by that in the HAL as well. One of the two (thumbv6 or thumbv7) does not take ownership of the pin, even though it needs it for PWM. That API could probably be improved.

boards/p1am_100/src/lib.rs
bradleyharden: I know you say you prefer letting the compiler infer the return types, but I think you'll discover that approach doesn't work well in this particular instance.

The `Spi` type is made to be used with the various `embedded_hal` traits. But those traits are only implemented for specific `Config` and `Pads` types. The idea is to use the type system to prevent misconfiguration.

In this case, though, you've only told the caller that the `spi::Config` type satisfies `ValidConfig`, nothing more. So the caller will have no idea what the `Spi` struct is capable of, and won't be able to use any of the `embedded_hal` traits.

If you want to use an existential type for this, I would probably choose `impl FullDuplex` or something like that. As it stands right now, I don't think you can do anything useful at all with the return value of this function.

boards/p1am_100/src/lib.rs
bradleyharden: I'm still not sure I quite understand the motivation for inferring types in situations like this. The output type is directly linked to the `Scl`, etc. input types. There would be no way to change the output type without also changing the input types, so all changes are breaking. And by using existential types, you only serve to hide potentially useful information from the caller.

If that's how you want to handle it, then feel free. I'm just trying to point out that there may be potentially unintended consequences when you hide type information (see the comment on `Spi` above). The HAL uses types to ensure valid configurations, so hiding types can lead to problems.

bradleyharden: FYI, I also plan to add `PinId` aliases to the `bsp_pins!` macro. So you could use something like `SclId` in place of `SpecificPinId<Scl>` or `<Scl as AnyPin>::Id`.

PR #432, opened 2021-04-22
DMAC peripherals

jbeaurivage:
# DMAC peripherals

This is a tracking PR to add DMA support for some peripherals. It is also meant to be an example for other contributors showing how to integrate DMA transfers into other peripheral implementations.

## API changes

There are some breaking API changes in the PR. Recognizing the fact that individual DMA `Channel`s should be moveable, and should be able to control themselves in a standalone way, the API was modified such that each `Channel` can access the `DMAC` PAC peripheral via `Peripherals::steal().DMAC`. This avoids neesing to pass an `&mut DmaController` reference every time we want to mutate or read a `Channel`'s configuration. The contract is as follows: each individual `Channel` may only access registers pertaining to itself. It may not modify or read registers controlling other channels, or general DMAC configuration registers. This is mostly done through the `Channel::with_chid` method.

## To-Do

- [x] thumbv7em SPI DMA-enabled transfers
- [x] Document all the things! + Add RTIC waker example

## async?

I want to start off this section by mentioning I don't know how contributors of this HAL feel about supporting `async/await`, so any comments are appreciated. 

Given that the entire point of DMA transfers is to free up the CPU while the hardware does the transfer, I think it makes sense to support a standard asynchronous model. Right now, DMA `Transfer`s have the option of adding a `waker` closure, which will get called when the transfer is complete, à la `sercom::v2::spi_future`.

This works very well on the application side of things, but it does make developing device driver libraries really hard, without resorting to the blocking peripheral implementations. The defacto traits in [embedded-hal](https://github.com/rust-embedded/embedded-hal) only support a small subset of non-blocking operations. Notably, I2C only has blocking support. I'd like to propose that we start supporting async traits, maybe from [embassy](https://github.com/embassy-rs/embassy)? These could be feature-gated behind an `async` feature.

Asynchronous DMA/Peripheral transfers seem to fit perfectly within the scope of usefulness of `async/await`. This is a personal opinion, but I'd really like seeing more asynchronous device drivers within the Rust embedded ecosystem.

### DIFF
--- a/hal/src/dmac/channel.rs
+++ b/hal/src/dmac/channel.rs
@@ -36,7 +36,9 @@ use crate::{
     target_device::DMAC,
     typelevel::{Is, Sealed},
 };
-use core::marker::PhantomData;
+
+use core::{marker::PhantomData, mem};
+use modular_bitfield::prelude::*;
 
 #[cfg(feature = "min-samd51g")]
 use super::dma_controller::{BurstLength, FifoThreshold};
@@ -140,7 +142,7 @@ impl<Id: ChId, S: Status> Channel<Id, S> {
     /// If an interrupt were to change the CHID register, we would be faced
     /// with undefined behaviour.
     #[cfg(any(feature = "samd11", feature = "samd21"))]
-    fn with_chid<F: Fn(&DMAC)>(&mut self, dmac: &DMAC, fun: F) {
+    fn with_chid<F: FnMut(&DMAC)>(&mut self, dmac: &DMAC, mut fun: F) {
         cortex_m::interrupt::free(|_| {
             // SAFETY: this is actually safe as long as we write a correct channel number to
             // the CHID register
@@ -158,7 +160,7 @@ impl<Id: ChId, S: Status> Channel<Id, S> {
     /// to the correct channel number and run the closure on that.
     #[cfg(feature = "min-samd51g")]
     #[inline]
-    fn with_chid<F: Fn(&CHANNEL)>(&mut self, dmac: &DMAC, fun: F) {
+    fn with_chid<F: FnMut(&CHANNEL)>(&mut self, dmac: &DMAC, fun: F) {
         let mut ch = &dmac.channel[Id::USIZE];
         fun(&mut ch);
     }
@@ -173,7 +175,6 @@ impl<Id: ChId, S: Status> Channel<Id, S> {
         mut self,
         controller: &mut DmaController,
         lvl: PriorityLevel,
-        enable_interrupts: bool,
     ) -> Channel<Id, Ready> {
         let dmac = controller.dmac();
 
@@ -187,21 +188,24 @@ impl<Id: ChId, S: Status> Channel<Id, S> {
 
             #[cfg(feature = "min-samd51g")]
             d.chprilvl.modify(|_, w| w.prilvl().bits(lvl as u8));
-
-            if enable_interrupts {
-                // Enable all interrupt sources
-                d.chintenset.modify(|_, w| {
-                    w.susp().set_bit();
-                    w.tcmpl().set_bit();
-                    w.terr().set_bit()
-                });
-            }
         });
 
         Channel {
             _id: self._id,
             _status: PhantomData,
         }
+    }
+
+    pub fn enable_interrupts(&mut self, dmac: &mut DmaController, flags: InterruptFlags) {
+        self.with_chid(dmac.dmac(), |d| {
+            d.chintenset.write(|w| unsafe { w.bits(flags.into()) })
+        })
+    }
+
+    pub fn disable_interrupts(&mut self, dmac: &mut DmaController, flags: InterruptFlags) {
+        self.with_chid(dmac.dmac(), |d| {
+            d.chintenclr.write(|w| unsafe { w.bits(flags.into()) })
+        })
     }
 
     #[inline]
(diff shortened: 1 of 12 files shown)


hal/src/dmac/transfer.rs
```diff
@@ -286,21 +285,24 @@ impl<B: AnyBufferPair> AsMut<B> for SpecificBufferPair<B> {
...
+pub struct Transfer<Chan, Buf, Pld = (), W = fn()>
 where
     Buf: AnyBufferPair,
     Chan: AnyChannel,
+    W: FnMut() + 'static,
 {
     chan: Chan,
     buffers: Buf,
     payload: Pld,
+    waker: Option<W>,
```
bradleyharden, line 297: I just had an idea. With this approach, you always pay for the space of a function pointer in every transfer struct. Maybe you could remove the Fn() bound on the struct and set the default type to (). You can apply the bound to the with_waker function and any function where it would be used. I think that would work.

hal/src/dmac/channel.rs
```diff
@@ -134,15 +136,22 @@ impl<Id: ChId, S: Status> Channel<Id, S> {
...
-            unsafe {
+            // SAFETY: This is ONLY safe if the individual channels are GUARANTEED not to
+            // mess with either:
+            // - The global DMAC configuration
+            // - The configuration of other channels.
+            //
+            // In practice, this means that the DMAC registers should only be accessed
+            // through the `with_chid` method.
+            let dmac = unsafe {
+                let dmac = Peripherals::steal().DMAC;
```
bradleyharden, line 149: This is a somewhat ad-hoc approach. I used to do something like this in the gpio module. I came up with and alternative that I think is better though. You can create your own PAC-like, ZST, smart pointers that provide an API for each channel's registers. It's basically as if you split the PAC struct into segments according to channel. Then each channel can take ownership of that ZST smart pointer. I think that ends up being cleaner than stealing.

See the gpio module for an example. It's a little bit more complicated there, because I needed to implement it for both Pin and DynPin, but I think the documentation on the RegisterInterface trait should explain it well.

hal/src/dmac/channel.rs
```diff
@@ -198,21 +199,36 @@ impl<Id: ChId, S: Status> Channel<Id, S> {
         }
     }
 
+    pub fn enable_interrupts(&mut self, flags: InterruptFlags) {
```
bradleyharden, line 202: +1 to this change

hal/src/dmac/channel.rs
bradleyharden: Also check out some of the clock modules in my clock_rebase branch. It's linked in the clock RFC issue. I used this pattern in almost every module of clock::v2.

hal/src/dmac/channel.rs
```diff
@@ -344,11 +361,52 @@ impl<Id: ChId> Channel<Id, Busy> {
...
+        while !self.xfer_complete() {}
         Channel {
             _id: self._id,
             _status: PhantomData,
         }
     }
+
+    #[inline]
+    #[cfg(any(feature = "samd11", feature = "samd21"))]
+    pub fn callback(&mut self) {
```
bradleyharden, line 374: Is this just a placeholder? What does it do?

bradleyharden:
Isn't embassy nightly only? I believe it uses several unstable features, ones that may not be stable for a long time either. It's also very new. I think it just got pushed to crates.io. I don't know that it's a great fit for a dependency. I guess if its feature-gated it's not a big deal.

To be honest, I'm not totally sold on the idea of async. It seems interesting, but there also seem to be a lot of mixed opinions on it. For embedded work, I really like the RTIC approach of interrupt driven tasks. Most of my use cases for embedded devices have some sort of real-time component, which seems to fit well with RTIC. Maybe the async model is better for other situations? Or maybe I just don't see the benefits yet?

bradleyharden:
I say that having developed spi_future myself, though, so maybe there are good use cases for it.

hal/src/dmac/channel.rs
jbeaurivage: This is still unfinished, but it's intended to be called by `Transfer::callback`, which should be placed in the DMAC interrupt handler. It will take care of calling the `Transfer`'s `waker` if it exists.

jbeaurivage:
@bradleyharden To be honest, I kind of agree that RTIC is probably the superior model as far as maturity right now. One problem I'm trying to solve is taking advantage of concurrent transfers and communication protocols *in the context of device driver crates*. Unless I'm missing something, all of the concurrency logic from e.g. RTIC has to be inside the final executable crate, which leaves device drivers with little options for dealing with asynchronicity. Here's
``` rust
/// This is a function in a device driver crate
fn example_transaction(spi: embedded_hal::spi::FullDuplex){
    // Receive single byte
    nb::block!(spi.read()).unwrap();

    // Process data...

    // Send single byte
    nb::block!(spi.send(something)).unwrap();
}
```
Even though `embedded_hal` provides non-blocking transfers, the only useful thing that can be done is to `block!` on them. The problem is even more exacerbated if we look at the fact that `embedded_hal` only provide non-blocking traits for single-byte transfers, and only blocking traits for buffered transfers. I2C transfers for example, are exclusively blocking.

I guess another way for library crates to achieve something similar would be to provide a waker, like `spi_future` and this PR currently does, and return an `nb::Error`. This does add a whole lot of work for driver developers though, because now the state has to be tracked internally, which is a little like what `async/await` is already trying to accomplish.

hal/src/dmac/channel/mod.rs
```diff
@@ -332,23 +277,71 @@ impl<Id: ChId> Channel<Id, Busy> {
     /// and BUSYCH is set. To make sure the transfer is actually complete, the
     /// channel needs to be both NOT PENDING and NOT BUSY.
     #[inline]
-    pub(crate) fn xfer_complete(&self, dmac: &DMAC) -> bool {
-        let id = Id::U8;
-        dmac.busych.read().bits() & (1 << id) == 0 && dmac.pendch.read().bits() & (1 << id) == 0
+    pub(crate) fn xfer_complete(&self) -> bool {
+        !self.regs.busych.read_bit() && !self.regs.pendch.read_bit()
```
ianrrees, line 281: (have only checked the SAMD21 docs, perhaps other chips are different)

I think this actually reports whether a beat transfer is pending or ongoing; it'll return `true` if there are more beats to transfer but the channel is waiting on a trigger, for instance.  In that case at least, what's actually wanted is `!self.regs.chctrla.read().enable().bit_is_set()`.

hal/src/dmac/channel/reg.rs
```diff
@@ -40,25 +40,35 @@ pub(super) trait Register<Id: ChId> {
...
+        // In practice, this means that the channel-specific registers should only be
+        // accessed through the `with_chid` method.
+
+        // Get the CHID contents before changing channel
+        let old_id = dmac.chid.read().bits();
+
+        // Change channels
+        unsafe {
+            dmac.chid.write(|w| w.id().bits(Id::U8));
+        };
```
ianrrees, line 62: Could this instead be done with a modify, something like:
``` Rust
let old_id = dmac.chid.modify(|r, w| {
    let old = r.read().bits();
    w.id().bits(Id::U8);
    old
});
```

jbeaurivage:
Okay party people, I think this is ready to merge!

sajattack:
I see a few updates in the examples, but no example of constructing the spi dma.

jbeaurivage:
@sajattack 
> I see a few updates in the examples, but no example of constructing the spi dma.
True that, spi docs is now up!

PR #437, opened 2021-04-29
WIP: add xiao_m0 sck/miso/mosi pin aliases

LongLiveCHIEF:
When using `<Pin>` type, this now allows the `xiao_m0` to use `<Pin>.sck`, `<Pin>.miso`, and `<Pin>.mosi` in the same way as other BSP's.

corresponding datasheet pinout here: https://files.seeedstudio.com/wiki/Seeeduino-XIAO/res/Seeeduino-XIAO-v1.0-SCH-191112.pdf

### DIFF
--- a/boards/xiao_m0/src/lib.rs
+++ b/boards/xiao_m0/src/lib.rs
@@ -55,6 +55,13 @@ define_pins!(
     /// Pin A10/D10/MOSI
     pin a10 = a6,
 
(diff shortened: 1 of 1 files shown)


LongLiveCHIEF:
I'm working on adding an example that uses `spi_master` for the `xiao_m0` before marking this as ready to merge.

PR #438, opened 2021-05-06
Add chip atsamd11d

VictorKoenders:
First of all thank you for this crate!

While trying out a board that [a friend of mine](https://twitter.com/mifune/status/1382707685368471555) gave me, I noticed that some of the pins didn't exist. Turns out he added a samd11d to the board, and this crate only has samd11c.

I followed the instructions on the wiki and was able to add the chip without any problems.

So far I have only tested the GPIO pins. Are there any areas that you think need customizing between the 2 chips?


bradleyharden:
@VictorKoenders, I have to admit, I knew we didn't have `atsamd11d` when I wrote `gpio::v2`, but I didn't feel like adding it at the time, and I didn't think anyone would actually end up using it. Thanks for adding this.

I think the `#[cfg(not(feature = "samd11c"))]` approach is fine for now, but we may want to eventually move to a `min-samd11d` feature. I'm not sure. It probably depends on other use cases.

VictorKoenders:
From what I can tell from the [datasheet](http://ww1.microchip.com/downloads/en/devicedoc/atmel-42363-sam-d11_datasheet.pdf), there are 3 possible flavors for the samd11 chip:
afbeelding

Looking at the [samd21 datasheet](https://ww1.microchip.com/downloads/en/DeviceDoc/SAM-D21DA1-Family-Data-Sheet-DS40001882G.pdf), there are a _lot_ of variations, with 32, 35, 48 and 64 pins.

To me as a user it makes more sense to have a `samd11d` and `samd11c` feature, but for the samd21 that would probably be very difficult to maintain.

Alternatively the crate could just expose all `samd11` pins, and if some of those pins are not available on `samd11c` then accessing this won't do anything.

bradleyharden:
To be clear, the `samd11d` feature would still exist. Right now we have features for each board, e.g. `samd51j` and features that cover multiple boards, e.g. `min-samd51g`. I'm suggesting we might want to add a `min-samd11d` feature. That wouldn't mean replacing `samd11d`.

The `min-xxxx` features are there to make it easier to select which boards have which features in the code. It keeps the `#[cfg(any(...))]` spam to a minimum.

VictorKoenders:
Rebased onto the latest version of master.

Is this pull request blocked on anything?

sajattack:
LGTM!

PR #441, opened 2021-05-19
Allow wider range of USB maxPacketSize values

ianrrees:
This is important for isochronous endpoints, which reserve bandwidth on the USB as opposed to bulk endpoints which use whatever bandwidth is available.  Isochronous transfers on full-speed USB can have maximum packet sizes anywhere between 1B and 1023B (not just powers of 2), but for instance bulk ones need to be either 1B, 2B, ... 32B, or 64B.

I've opened https://github.com/mvirkkunen/usb-device/issues/71 in usb-device to propose checking for valid endpoint sizes there, since that code will be the same across all implementations.

(note the old `alloc_ep()` code made buffers at least 64B - I couldn't find a reason for that, and it seems to work fine for smaller ones)

### DIFF
--- a/hal/src/thumbv6m/usb/bus.rs
+++ b/hal/src/thumbv6m/usb/bus.rs
@@ -805,7 +805,20 @@ impl Inner {
         max_packet_size: u16,
         interval: u8,
     ) -> UsbResult<EndpointAddress> {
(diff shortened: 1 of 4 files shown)


bradleyharden:
I'm going to merge this. Let me know if you hear anything from the `usb-device` repo.

PR #434, opened 2021-04-26
Refactor SERCOM pads, add documentation and port feather_m0 to gpio::v2

bradleyharden:
This is a megalith of changes I've been lumping into the same branch (exactly what you're supposed to do with Git...).

This still needs a lot of work on the documentation side of things. And I should probably split it out into digestible pieces. But I wanted to post it now, because it contains a complete port of the feather_m0 to `gpio::v2`. Several people have had questions about porting, so I wanted to get an example posted ASAP.

🚧 **NOTE** 🚧 Parts of the feather_m0 port are dependent on changes made to the HAL in this branch. Those changes have not yet been merged, so you may not be able to copy everything as-is.

@TDHolmes & @jaredwolff, you're the two that come to mind, but I know there were others working with `gpio::v2`.

PR #428, opened 2021-04-10
Ccbuf dith4 mode svd patch

TDHolmes:
Closes #377

PR #443, opened 2021-05-19
UART v2

### LINKED ISSUE #329: Refactor UART implementation
- [ ] Make UART generic over some trait that manages configuration of the baud rate (arithmetic vs fractional)
- [ ] Have methods to enable/disable interrupts & check intflag properly
- [ ] Handle errors properly by reading `intflag.ERR` + `status.*`, clearing them, and discarding bytes as appropriate
- [ ] Make UART parameters (stop bits, parity, polarity) configurable
- [ ] Generated documentation needs work

jbeaurivage:
This PR aims to provide a new UART driver with an API similar to `sercom::v2::spi`.

In fact, most of the type-level code was shamelessly copied from @bradleyharden's SPI driver. Thanks for doing most of the legwork! 

# Supported
* Basic UART functionality including different frame configurations: parity, stop bits and character size are configurable.
* Fractional or Arithmetic baud mode
* Flow control (CTS/RTS pins). 
* Rx + Tx, Rx-only or Tx-only
* Splitting of Rx and Tx halves
* IrDA encoding

# Not supported
* LIN mode (thumbv7em devices)
* 32-bit extension mode (thumbv7em devices)
* `Future`-mode - Use DMA transfers instead
* Synchronous (USART) operation

# To Do
- [x] thumbv7em support
- [x] Add baudrate flexibility (fractional baud + oversampling)
- [x] Refine Rx and Tx splitting
- [x] Finish IrDA implementation
- [x] DMAC support
- [x] Add feather_m0 and feather_m4 examples

# Dependencies
This PR is dependent on the new `Pads` API (#434).

# Related issues
Closes #329


hal/src/thumbv6m/sercom/v2/uart.rs
```diff
@@ -0,0 +1,1937 @@
...
+    S: Sercom,
+    RX: GetOptionalPad<S>,
+    TX: GetOptionalPad<S>,
+    RTS: GetOptionalPad<S>,
+    CTS: GetOptionalPad<S>,
+{
+    rx: RX::Pad,
+    tx: TX::Pad,
+    ready_to_send: RTS::Pad,
+    clear_to_send: CTS::Pad,
```
ianrrees, line 415: Minor nit: to me it feels a little inconsistent to abbreviate rx and tx, but not rts and cts.  I've got a very weak preference for receive & transmit rather than rts & cts, just because they're easier to search for, but either would be fine.

jbeaurivage: Fair enough. Though I prefer keeping tx and rx, only because `receive` and `transmit` could be mistaken as the act of actually receiving and transmitting data, rather than just assigning pins to pads.

bradleyharden: I agree with @ianrrees. But also keep in mind that this isn't user-facing.

jbeaurivage: Ah I see. I'll change the struct fields to receive/transmit, but change the user-facing methods to rts and cts. I hope this isn't too confusing?

bradleyharden:
Some comments:

- I would define the `Registers` trait directly on instances of `Sercom`. That would get rid of the associated type. Both `UartRx` and `UartTx` store a `Sercom` already, so you have direct access to it. The `Registers` trait is just a layer of convenience functions for the `Sercom`.
- You might want to rethink how splitting works. Right now, you let users take out the `UartRx` and `UartTx` pieces. But if they do that and then drop the rest, they are dropping the `Config`, which contains the GPIO pins. To retrieve the pins, you would have to do partial moves out of the `Uart` and then back in, before calling the `disable` function.
  - One approach would be to copy the `Config` struct into each half. You would have `split` and `join` methods to handle the logic. The `Config` struct is zero-sized except for the `freq` field, so it's minimal memory overhead. Doing so would also save you from having to copy the `Sercom` directly.
  - However, with this approach, would wouldn't want to let people `free` one half directly, because that could let you duplicate GPIO pins. You would need to force a `join`, then `disable`, then `free`.

hal/src/thumbv6m/sercom/v2/uart.rs
```diff
@@ -0,0 +1,2024 @@
...
+    {
+        /// Transform an [`UartTx`] into a DMA [`Transfer`]) and
+        /// start sending the provided buffer.
+        #[inline]
+        pub fn send_with_dma<Chan, B, W>(
+            self,
+            buf: B,
+            mut channel: Chan,
+            waker: W,
+        ) -> Transfer<Channel<ChannelId<Chan>, Busy>, transfer::BufferPair<B, Self>, (), W>
```
bradleyharden, line 1850: I think `Chan::Id` might be clearer here, instead of `ChannelId<Chan>`. I'm pretty sure that should work.

bradleyharden: Actually, I just realized that `Chan` was a type parameter, not a type. That confused me for a second. Is there a reason you use `Chan` instead of `C` here? To me, that's clearer. A single letter immediately tells me it's a type parameter, so I go look for its definition. That also cuts down on the return type:

```rust
Transfer<Channel<C::Id, Busy>, transfer::BufferPair<B, Self>, (), W>
```

bradleyharden: Finally, since this is already namespaced in its own `uart_dma` module, you might get rid of the `channel` and `transfer` prefixes.

bradleyharden:
By the way, I think the type-level stuff meshed really well here. You were able to use the `Word` type to define the size of your `Beat`. I like that.

hal/src/thumbv6m/sercom/v2/uart.rs
jbeaurivage: > Actually, I just realized that `Chan` was a type parameter, not a type. That confused me for a second. Is there a reason you use `Chan` instead of `C` here? To me, that's clearer. 

Just a matter of preference. When there are too many type parameters, I tend to use longer names as to not get confused. Though I can definitely see the case for keeping them single-letter.

> You might want to rethink how splitting works. Right now, you let users take out the `UartRx` and `UartTx` pieces. But if they do that and then drop the rest, they are dropping the `Config`, which contains the GPIO pins. To retrieve the pins, you would have to do partial moves out of the `Uart` and then back in, before calling the `disable` function.
>       
> One approach would be to copy the `Config` struct into each half. You would have `split` and `join` methods to handle the logic. The `Config` struct is zero-sized except for the `freq` field, so it's minimal memory overhead. Doing so would also save you from having to copy the `Sercom` directly.
> However, with this approach, would wouldn't want to let people `free` one half directly, because that could let you duplicate GPIO pins. You would need to force a `join`, then `disable`, then `free`.

I don't really see how that's different from partially moving the `rx` and `tx` fields out of (and back into) a `Uart`. I guess the only difference is you could have `rx` and `tx` halves without keeping the underlying `Uart` struct around. I also experimented with using `split` and `join` methods, but it gets complicated quite fast because of all the typelevel stuff going on.

bradleyharden: The problem isn't really keeping the `Uart` struct around. It's that you can't *move* the `Uart` struct once you take out the halves. I'm not really sure how often anyone would need to `join` anyway, so maybe it doesn't matter all that much. But partial moves will substantially limit when/how you can join them.

Do you happen to remember what problem you had with the type-level stuff? Maybe there's a cleaner way to do it?

jbeaurivage: I don't remember exactly. I think it was something to do with how to split the struct depending on whether `Uart` was `Tx`, `Rx` or `Tx+Rx`. 

I think now I'm leaning towards eliminating the `Uart` struct entirely, and requiring `Config` to be split into Tx and Rx halves to be enabled. This would also allow the Rx and Tx to be enabled separately at the hardware level (specifically, the USART receiver should only be enabled immediately before actually receiving data, otherwise buffer overflow errors will be thrown. - See comment below)

hal/src/thumbv6m/sercom/v2/uart.rs
```diff
@@ -0,0 +1,1930 @@
...
+        // Workaround is to read a few bytes to clear the RX buffer (3 bytes seems to do
+        // the trick), then manually clear the BUFOVF bit Clear rx buffer
+        for _ in 0..=2 {
+            let _data = usart.data.read();
+        }
+
+        // Clear buffer overflow error
+        self.clear_errors(Error::Overflow.into());
+    }
+}
```
jbeaurivage, line 1765: I wrote a `UartRx::flush` method to clear the rx buffer, as well as buffer overflow errors, in the event the receiver was enabled some time before data actually starts being read. However, I think I encountered a hardware bug (see commented code), leading to this inelegant workaround. 

Also, I couldn't find anywhere in the datasheet the actual length of the Rx buffer. After experimentation, it seems to be 3 bytes, as the buffer overflow errors pop up after reading the second byte. The BUFOVF bit gets set to true when the *next* word to be read has the error. I also don't know if the buffer for thumbv7em targets is also 3 bytes long.

hal/src/thumbv6m/sercom/v2/uart.rs
```diff
@@ -0,0 +1,2024 @@
...
+};
+
+use crate::gpio::v2::{AnyPin, OptionalPin, SomePin};
+use crate::sercom::v2::*;
+use crate::typelevel::{Is, NoneT, Sealed};
+
+use bitflags::bitflags;
+use embedded_hal::blocking;
+use embedded_hal::serial::{Read, Write};
+use nb::Error::WouldBlock;
```
ianrrees, line 238: I believe this needs to be `pub use nb::Error::WouldBlock;`, so that callers of `.read()` and `.write()` can match on the Result.

jbeaurivage: I'd prefer if we reexported `nb` at the crate level instead (if we don't already do that? Maybe in the prelude?) since a lot of peripherals use e-hal blocking traits

bradleyharden:
@jbeaurivage, since you still have boxes unchecked, I converted this to a draft PR. I hope that's ok. It helps me identify merge candidates when browsing the PR list.

jbeaurivage:
@bradleyharden no problem. Hopefully it should be ready soon!

hal/src/thumbv6m/sercom/v2/uart.rs
ianrrees: I may be missing something, but couldn't find a file in the HAL that made `nb` public.  So, I've made https://github.com/atsamd-rs/atsamd/pull/458

hal/src/thumbv6m/sercom/v1/spi.rs
```diff
@@ -1,4 +1,8 @@
-use core::marker::PhantomData;
+#![deprecated(
+    since = "0.12.0",
+    note = "Please use sercom::v2::uart instead.
```
bradleyharden, line 3: spi not uart here

hal/src/sercom/mod.rs
```diff
@@ -18,9 +18,10 @@
 //! The [v2] module will eventually replace [v1]. New users are encouraged to
 //! use [v2] instead of [v1].
 //!
-//! The new [`v2::spi`] module is substantially more configurable and safe than
-//! the existing, [`v1::spi`] module. To assist in migration, the
-//! [`v2::spi::Pads`] struct accepts both [`v1::Pin`]s and [`v2::Pin`]s.
+//! The new [`v2::spi`] and [`v2::uart`] modules are substantially more
+//! configurable and safe than the existing, [`v1::spi`] and [`v1::uart`]
+//! modules. To assist in migration, the [`v2::spi::Pads`] and
+//! [`v2::uart::Pads`] structs accepts both [`v1::Pin`]s and [`v2::Pin`]s.
```
bradleyharden, line 24: accept

hal/src/sercom/v2/dma.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        #[cfg(feature = "min-samd51g")]
+        let trigger_action = TriggerAction::BURST;
+
+        #[cfg(any(feature = "samd11", feature = "samd21"))]
+        let trigger_action = TriggerAction::BEAT;
+
+        // SAFETY: We use new_unchecked to avoid having to pass a 'static self as the
+        // destination buffer. This is safe as long as we guarantee the source buffer is
+        // static.
```
bradleyharden, line 128: Shouldn't this say "as long as the *destination* buffer is static"?

hal/src/sercom/v2/dma.rs
```diff
@@ -0,0 +1,271 @@
...
+    }
+}
+
+//=============================================================================
+// SPI DMA transfers
+//=============================================================================
+
+#[cfg(feature = "min-samd51g")]
+#[doc(hidden)]
+pub trait CharSize: spi::Length {}
```
bradleyharden, line 141: This isn't going to work. I think your DMA approach here will only work for SAMD5x chips when `Length` is `U4` or less. For those lengths, each write to the `DATA` register is essentially atomic. It represents a complete transaction. But for lengths greater than `U4`, the SPI peripheral keeps track of the state of the transfer in its `LENGTH` register, so your concept of `buffer_len = 1` doesn't match.

In general, I think it might make more sense to define respective implementations in the `thumbv*` modules and re-export them here. Otherwise, you'll end up with a ton of `cfg`s.

hal/src/thumbv6m/sercom/v2/uart/capability.rs
```diff
@@ -0,0 +1,278 @@
...
+
+use crate::typelevel::Sealed;
+use bitflags::bitflags;
+use core::convert::TryFrom;
+
+/// Type-level enum representing the capabilities of a UART peripheral
+pub trait Capability: Sealed {
+    const FLAG_MASK: u8;
+    const STATUS_MASK: u16;
+}
```
bradleyharden, line 13: It looks like SPI, UART and I2C all have `u8` flag and `u16` status registers. Let's go ahead and elevate this to a `sercom` module trait.

hal/src/sercom/v2/dma.rs
jbeaurivage: No, a transmission is memory -> SERCOM. The destination buffer is obviously `'static`, we need to guarantee that the buffer in memory is static. Though you pointing that out made me realize I forgot the static bounds for B and W!

hal/src/thumbv6m/sercom/v2/uart/capability.rs
```diff
@@ -0,0 +1,278 @@
...
+        // RXC
+        0x04 |
+        // RXS
+        0x08 |
+        // CTSIC
+        0x10 |
+        // RXBRK
+        0x20 |
+        // ERROR
+        0x80;
```
bradleyharden, line 47: Maybe you could express this as a `Status` type and then convert it to a `u8`? That might be clearer than all the comments. I'm not sure if the conversion to a `u8` is a `const` function or not though.

jbeaurivage: Unfortunately I tried doing that, and it seems that the flags served out by `bitflags` aren't actually `const` friendly. I'll play a bit more with it, but it might have to stay a plain `u8`. It would make for a good feature suggestion for `bitflags` though.

bradleyharden: Hmm, it looks like the `.bits()` function is indeed `const`. But I don't see any way to create the value, because the `BitOr` trait function isn't `const`. And I don't see any other way to construct the correct value that is `const`.

hal/src/thumbv6m/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,233 @@
+//! Register-level access to UART configuration
+
+use crate::sercom::v2::*;
+
+use super::{BaudMode, Parity, StopBits};
+use crate::time::Hertz;
+
+pub(super) trait RegisterInterface {
+    type Sercom: Sercom;
+    fn sercom(&self) -> &Self::Sercom;
```
bradleyharden, line 10: Unless you have some opposition, I would convert this to a type rather than a trait.

```rust
pub(super) type Registers<S: Sercom> {
    sercom: S,
}

impl<S: Sercom> Registers<S> {
    fn msb_first(&mut self, msb_first: bool) {
        // ...
    }
}
```
Then have the `Config` struct store a `Registers` struct *instead* of a `Sercom`. Basically, the `Registers` type acts to wrap the `Sercom` peripheral and expose a safer interface. You could even implement `Sync` for `Registers`, because you erase all of the interior mutability on the `Sercom` struct by requiring `&mut self` when appropriate.

This has an advantage, because you don't have any name clashes between public and private functions. It's also more consistent with what we're doing in the `clock::v2` module.

Note that I *had* to use a trait for `DynPin`, because `DynPin` erases the `PinId` type, so the corresponding `DynPinId` value can only be retrieved at run-time. A trait was the only way to implement it for both `Pin` and `DynPin`. But I still defined a `Registers<I: PinId>` type for `Pin`.

hal/src/thumbv7em/sercom/v2/spi.rs
```diff
@@ -538,10 +538,10 @@ pub type PadsFromIds<S, I, DI = NoneT, DO = NoneT, CK = NoneT, SS = NoneT> = Pad
...
     type Sercom: Sercom;
     type IoSet: IoSet;
-    type DataIn: OptionalPad;
-    type DataOut: OptionalPad;
-    type Sclk: OptionalPad;
-    type SS: OptionalPad;
+    type DataIn: OptionalPin;
+    type DataOut: OptionalPin;
+    type Sclk: OptionalPin;
+    type SS: OptionalPin;
```
bradleyharden, line 544: I think you didn't update something correctly. This should definitely be `OptionalPad`. Looks like there are problems throughout this module.

hal/src/thumbv6m/sercom/v2/uart/capability.rs
bradleyharden: Looks like this is a workaround.

https://github.com/bitflags/bitflags/issues/180#issuecomment-499302965

But I don't know how much I like it compared to the manual approach.

hal/src/dmac/channel/reg.rs
```diff
@@ -256,6 +256,7 @@ reg_proxy!(swtrigctrl, bit, rw);
 /// Acts as a proxy to the PAC DMAC object. Only registers and bits
 /// within registers that should be readable/writable by specific
 /// [`Channel`]s are exposed.
+#[allow(dead_code)]
 pub(super) struct RegisterBlock<Id: ChId> {
```
bradleyharden, line 260: In both the `gpio::v2` and `clock::v2` modules, we seem to have settled on an architecture where a `Registers` struct acts to split a PAC peripheral, like a `GPIO` or an `OSCCTRL`, into individual pieces that are owned, `Send` and `Sync`, like `Pin`s and different oscillator types. On top of that, these structs provide a "task-centric" API rather than a "register-centric" API, if that makes sense.

This implementation strays from that convention. Each `RegisterBlock` *does* represent an individual piece of the full peripheral, but each piece is still `!Sync`, because it stores a `DMAC`. It also has a very register-centric API.

It's not really a big deal, but if you don't have a good reason to keep it this way, I think it might be helpful in the long run to stick with a similar convention across the HAL. Just a thought.

hal/src/thumbv6m/sercom/v2/uart/pad.rs
```diff
@@ -0,0 +1,452 @@
...
+//=============================================================================
+
+/// Marker trait for valid sets of [`Pads`]
+///
+/// This trait labels sets of [`Pads`] that satisfy the [`RxpoTxpo`] trait. It
+/// guarantees to the [`Config`] struct that this set of `Pads` can be
+/// configured through that trait.
+///
+/// [`Config`]: super::Config
+pub trait ValidPads: PadSet + RxpoTxpo {}
```
bradleyharden, line 450: I would get rid of the `GetCapability` trait and add `Capability` as an associated type on `ValidPads`. I think that makes sense, because a `PadSet` can't be valid unless it has an associated `Capability` and satisfies `RxpoTxpo`.

hal/src/thumbv6m/sercom/v2/uart/pad.rs
```diff
@@ -0,0 +1,452 @@
...
+/// [`Config`] struct doesn't need access to the [`Pin`]s directly.  Rather, it
+/// only needs to apply the [`SomePad`] trait bound when a `Pin` is required.
+/// The [`PadSet`] trait allows each [`Config`] struct to store an instance of
+/// [`Pads`] without itself being generic over all six type parameters of the
+/// [`Pads`] type.
+///
+/// [`Pin`]: crate::gpio::v2::Pin
+/// [type-level function]: crate::typelevel#type-level-functions
+/// [`Config`]: super::Config
+pub trait PadSet: Sealed {
```
bradleyharden, line 406: A while back, I renamed `spi::AnyPads` to `spi::PadSet`. At the time, I think the type-level mapping was a bit more complicated, which made it more complicated to implement the full `AnyKind` pattern. However, with the latest approach, `PadSet` is basically `AnyPads` without the `AsRef` and `AsMut` implementations.

There's really no use case for *most* of the `AnyKind` machinery, like `SpecifcPads` or the super traits. *But*, defining this as `AnyPads` might make it clearer that it follows the exact same pattern for the same purposes.

What do you think? Should we convert this back to `AnyPads` or stick with `PadSet`?

hal/src/thumbv6m/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1353 @@
...
+    #[inline]
+    pub fn char_size<C2: CharSize>(self) -> Config<P, C2> {
+        C2::configure(&self.sercom);
+        self.change()
+    }
+
+    /// Change the bit order of transmission (MSB/LSB first)
+    #[inline]
+    pub fn msb_first(mut self, msb_first: bool) -> Self {
+        self._msb_first(msb_first);
```
bradleyharden, line 584: This would become `self.regs.msb_first(msb_first);` with the proposed change.

hal/src/thumbv6m/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1353 @@
...
+}
+
+//=============================================================================
+// Rx/Tx specific functionality
+//=============================================================================
+
+impl<C, D> Uart<C, D>
+where
+    C: ValidConfig,
+    C::Word: PrimInt,
```
bradleyharden, line 1160: If you need `PrimInt`, I would add it to the definition of `CharSize::Word` and `AnyConfig::Word`.

hal/src/thumbv6m/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1353 @@
...
+        } else {
+            Err(WouldBlock)
+        }
+    }
+}
+
+impl<C, D> Write<C::Word> for Uart<C, D>
+where
+    C: ValidConfig,
+    C::Word: PrimInt + AsPrimitive<u16>,
```
bradleyharden, line 1264: You can also add `AsPrimitive<u16>` to `CharSize::Word`. You can add `u16: AsPrimitive<Word>` as well, but it won't be inferred anywhere, so you'll still need to propagate the bound.

hal/src/thumbv6m/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1353 @@
...
+                Bits8 => 2,
+            },
+
+            Fractional(n) => match n {
+                Bits16 => 1,
+                Bits8 => 3,
+            },
+        }
+    }
+}
```
bradleyharden, line 1353: This is a minor thing, but I generally like to be able to read a module from top to bottom and get each piece approximately when I need it. It seems like these should be defined before (above) `Config`. That's just a personal preference of mine, though.

I also like to add the `//=====================` headings, because I find it helps me skim the code. Not sure if you agree. It's up to you.

hal/src/thumbv6m/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1353 @@
...
+    }
+
+    /// Set the baud rate
+    ///
+    /// This function will calculate the best BAUD register setting based on the
+    /// stored GCLK frequency and desired baud rate. The maximum baud rate is
+    /// GCLK frequency/oversampling. Values outside this range will saturate at
+    /// the maximum supported baud rate.
+    ///
+    /// Note that 3x oversampling is not supported.
```
bradleyharden, line 629: I have not read about UART baud rates *at all*, but I'm curious why this isn't supported.

hal/src/thumbv6m/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1353 @@
...
+    }
+}
+
+impl<C: ValidConfig> AsMut<Uart<C, Duplex>> for (&mut Uart<C, RxDuplex>, &mut Uart<C, TxDuplex>) {
+    #[inline]
+    fn as_mut(&mut self) -> &mut Uart<C, Duplex> {
+        // SAFETY: Pointer casting &mut Uart<C, RxDuplex> into &mut
+        // Uart<C, Duplex> should be safe as long as RxDuplex and Duplex
+        // are both zero-sized.
+        unsafe { &mut *(self.0 as *mut _ as *mut Uart<C, Duplex>) }
```
bradleyharden, line 1149: I think it's true that *any* struct with only one non-zero sized field has the same memory layout as the field, even for `repr(Rust)`. I remember digging into that one time. So this should be fine.

hal/src/thumbv6m/sercom/v2/uart/reg.rs
bradleyharden: By the way, I'm planning to implement this approach in `v2::spi` as well, when I get around to it.

hal/src/thumbv6m/sercom/v2/uart.rs
bradleyharden: This is very weird. Did you ever follow up with an example sent to Microchip?

hal/src/sercom/v2/dma.rs
bradleyharden: Ah yes, but you have the same comment for both the `send_with` and `receive_with` versions, so one of them needs to be changed.

bradleyharden: Wait, you *do* have the `'static` bounds for `B` and `W`. Am I missing something?

hal/src/thumbv6m/sercom/v2/uart/capability.rs
bradleyharden: Whoops, you were commenting as I was reviewing.

bradleyharden:
@jbeaurivage, as a general comment, why do you need completely separate implementation for `thumbv6m` and `thumbv7em`? I needed them, because I changed from using `Length` to `CharSize` which had a lot of downstream implications on the `Spi` type. But it looks like you don't have that problem. It seems like the `Pads` type is the only really big difference. Maybe consider consolidating?

hal/src/thumbv7em/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1371 @@
...
+
+//=============================================================================
+// ValidConfig
+//=============================================================================
+
+/// Marker trait for valid UART [`Config`]urations
+///
+/// A functional UART peripheral must have, at a minimum either a Rx or a Tx
+/// [`Pad`].
+pub trait ValidConfig: AnyConfig {}
```
bradleyharden, line 777: If you move `Capability` to an associated type of `ValidPads`, I don't think you'll need separate implementations here.

hal/src/thumbv6m/sercom/v2/uart/capability.rs
bradleyharden: Wait, no. We can't make it general, because the corresponding implementers need custom constants. Grr

hal/src/sercom/v2/dma.rs
jbeaurivage: Whoops, you're right. Disregard, I was looking at the wrong code section.

hal/src/thumbv7em/sercom/v2/spi.rs
jbeaurivage: Something weird happened with git. Should be fixed now.

jbeaurivage:
> @jbeaurivage, as a general comment, why do you need completely separate implementation for `thumbv6m` and `thumbv7em`? I needed them, because I changed from using `Length` to `CharSize` which had a lot of downstream implications on the `Spi` type. But it looks like you don't have that problem. It seems like the `Pads` type is the only really big difference. Maybe consider consolidating?

Absolutely, I came to the same realization. I'm going to consolidate most of the `thumbv6m` and `thumbv7em` modules.

hal/src/thumbv6m/sercom/v2/uart/mod.rs
jbeaurivage: I made the decision not to support it. The only reason is that 3x oversampling only supports arithmetic baud mode, and not fractional mode. Supporting it would've added complexity to the baud mode enum system. Maybe this isn't a good reason though.

hal/src/thumbv6m/sercom/v2/uart/pad.rs
jbeaurivage: I like `PadSet`. To me it makes it clearer that it's "a set of pads for the specified SERCOM" rather than just "any set of Pads". Plus, I think the `Any_` naming convention should probably be reserved specifically for `AnyKind` implementations.

hal/src/dmac/channel/reg.rs
jbeaurivage: I can see the reasoning behind that. I'll probably change it in the future, but I don't think this really falls within the scope of this PR

hal/src/sercom/v2/dma.rs
jbeaurivage: I think adding an `AtomicLength` trait, which limits acceptable `Lengths` to U1 through U4, fixes that.

hal/src/thumbv6m/sercom/v2/uart.rs
jbeaurivage: I haven't yet. I might need to get a C example up.

hal/src/sercom/v2/dma.rs
bradleyharden: I'm reworking the `spi` module to add `Capability` right now, so why don't we hold off on DMA for `spi` for the moment. I think your proposed solution works though.

hal/src/dmac/channel/reg.rs
bradleyharden: Sure, it was just something that caught my eye.

hal/src/thumbv6m/sercom/v2/uart/pad.rs
bradleyharden: I wasn't proposing to keep the trait the same yet call it `AnyPads`. I was proposing to implement the full `AnyKind` pattern for `AnyPads`. It's just that most features would go unused. You wouldn't ever need `SpecificPads` nor any of the super trait bounds. However, the one advantage would be consistency. I'm not sure if that's worth it though.

hal/src/thumbv6m/sercom/v2/uart/mod.rs
bradleyharden: I think it's fine. I chose to ignore 9-bit character sizes for the `thumbv7` `spi` module. Same reasoning, added complexity for little benefit.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,221 @@
...
+impl<S: Sercom> Registers<S> {
+    /// Helper function to access the given `SERCOM`
+    pub(crate) fn sercom(&self) -> &S {
+        &self.sercom
+    }
+
+    /// Helper function to access the underlying `USART` from the given `SERCOM`
+    #[cfg(any(feature = "samd11", feature = "samd21"))]
+    pub(crate) fn usart(&self) -> &crate::target_device::sercom0::USART {
+        self.sercom().usart()
```
bradleyharden, line 21: You can just access the `sercom` field directly here. There's no need for the `.sercom()` method anymore.

hal/src/thumbv7em/sercom/v2/spi.rs
```diff
@@ -804,26 +804,33 @@ impl Length for DynLength {
 /// Marker trait for statically known transaction [`Length`]s
 pub trait StaticLength: Length {}
 
+/// Marker trait for transactions that are performed atomically
+pub trait AtomicLength: Length {}
+
```
bradleyharden, line 809: Ah, I see you already implemented it. That's fine. I'll adjust as necessary in my refactor.

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1528 @@
...
+//=============================================================================
+// Interrupt flags
+//=============================================================================
+const DRE: u8 = 0x01;
+const TXC: u8 = 0x02;
+const RXC: u8 = 0x04;
+const RXS: u8 = 0x08;
+const CTSIC: u8 = 0x10;
+const RXBRK: u8 = 0x20;
+const ERROR: u8 = 0x80;
```
bradleyharden, line 503: Much better solution. Really obvious too. Good call.

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1528 @@
...
+    pads: P,
+    chsize: PhantomData<C>,
+    freq: Hertz,
+}
+
+#[cfg(any(feature = "samd11", feature = "samd21"))]
+type Clock = pac::PM;
+
+#[cfg(feature = "min-samd51g")]
+type Clock = pac::MCLK;
```
bradleyharden, line 743: I forgot about this problem. Simple enough solution though. Hopefully we can unify this with `clock::v2`.

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1528 @@
...
+            .ctrla
+            .modify(|_, w| w.mode().variant(MODE_A::USART_INT_CLK));
+
+        Self {
+            registers,
+            pads,
+            chsize: PhantomData,
+            freq: freq.into(),
+        }
+    }
```
bradleyharden, line 797: I started implementing a `Registers` type myself, and I realized a few things.

1. I think a better name for this function is `default`. Do you agree? I never did like the name `create`.
2. You could add a `.reset()` method to `Registers` and get rid of the `.swrst()` method on `Config`.
3. I decided to change the `.configure()` trait method approach on `ValidPads`, `CharSize`, etc. Instead of passing a `&RegisterBlock` or something else to the trait, I pass the trait's associated constants to a method on `Registers`. I personally like that structure more. I think it better reflects that the `Registers` type is responsible for configuring the registers, and the trait is responsible for providing the constants. It also gets rid of the ugly `<T as Trait>::method()` syntax. Instead, you get `registers.set_rxpo(P::RXPO)`.

None of this is mandatory though. I have a tendency to iterate indefinitely if left to my own devices.

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1528 @@
...
+    /// Enable the UART peripheral and return a [`Uart`] struct.
+    ///
+    /// UART transactions are not possible until the peripheral is enabled.
+    /// This method is limited to [`ValidConfig`]s
+    #[inline]
+    pub fn enable(mut self) -> Uart<Self, P::Capability> {
+        self.registers.enable();
+        Uart {
+            config: self,
+            capability: PhantomData,
```
bradleyharden, line 949: I don't think you do, but I needed one of the `Capability` structs to contain real data. Instead of making this `PhantomData`, I made it the struct itself, and I added `Default` as a super trait of `Capability`. Just thought I would tell you, in case you ever come across a similar scenario.

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1528 @@
...
+    ///
+    /// **Warning:** The implementation of of [`Write::flush`] waits on and
+    /// clears the `TXC` flag. Manually clearing this flag could cause it to
+    /// hang indefinitely.
+    pub fn clear_flags(&mut self, flags: Flags) {
+        // Remove flags not pertinent to Self's Capability
+        self.usart()
+            .intflag
+            .modify(|_, w| unsafe { w.bits(Self::capability_flags(flags).bits()) });
+    }
```
bradleyharden, line 1208: You might consider moving these functions to `Registers` as well. I'm not sure if you agree, but I see value in having **all** of the register interaction occur in one and *only* one type.

That being said, it's mostly a preference.

In this particular case, I think you could take the `flags` argument and intersect it with the `.capability_flags()` value before passing the corresponding bits to `self.config.as_mut().registers.clear_flags()`.

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1528 @@
...
+            .intenclr
+            .write(|w| unsafe { w.bits(Self::capability_flags(flags).bits()) });
+    }
+
+    /// Read the status flags
+    #[inline]
+    pub fn read_status(&self) -> Status {
+        let bits = self.usart().status.read().bits();
+        Status::from_bits_truncate(bits)
+    }
```
bradleyharden, line 1248: By the way, I renamed my `Errors` type to be `Status` and renamed my corresponding functions. You were right to stick with `Status`. Having both `Error` and `Errors` in the same module is confusing.

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1528 @@
...
+{
+    /// Disable the UART peripheral and return the underlying [`Config`]
+    #[inline]
+    pub fn disable(self) -> C {
+        let mut config = self.config;
+        config.as_mut().registers.disable();
+        config
+    }
+
+    /// Update the UART [`Config`]uration.
```
bradleyharden, line 1300: Just because I find this fun. How about

```
[`Reconfig`]ure the UART
```

jbeaurivage:
@bradleyharden @ianrrees this is positively, definitely ready to go now

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1531 @@
...
+    /// containing the corresponding [`Flags`] or [`Error`]
+    #[inline]
+    fn read_flags_errors(&self) -> Result<Flags, Error> {
+        self.read_status().try_into()?;
+        Ok(self.read_flags())
+    }
+
+    /// Flush the RX buffer and clear errors
+    #[inline]
+    pub fn flush(&mut self) {
```
ianrrees, line 1421: This name clashes with the `flush()` impl for Write, for duplex Uarts, which can be a bit confusing since they do quite different things.  I wonder if a name like `read_init()` might be better?

ianrrees:
Minor thing, but I wonder if it might be reasonable to take the character size out of the type system?  I'm of a mind to only support 8 bit characters in https://github.com/ianrrees/atsamd-usbd-uart , but if others were to be supported, it would be easier if the type of the Uart only depended on the word size.

hal/src/sercom/v2/uart/mod.rs
bradleyharden: Or maybe `flush_rx_buffer`

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1531 @@
...
+    /// Change the parity setting
+    #[inline]
+    pub fn parity(mut self, parity: Option<Parity>) -> Self {
+        self.registers.parity(parity);
+        self
+    }
+
+    /// Change the stop bit setting
+    #[inline]
+    pub fn stop_bit(mut self, stop_bit: StopBits) -> Self {
```
ianrrees, line 840: Super-minor nitpicky observation: it would be more consistent to always use plural `stop_bits()` etc

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1531 @@
...
+    /// Change the bit order of transmission (MSB/LSB first)
+    #[inline]
+    pub fn msb_first(mut self, msb_first: bool) -> Self {
+        self.registers.msb_first(msb_first);
+        self
+    }
+
+    /// Change the parity setting
+    #[inline]
+    pub fn parity(mut self, parity: Option<Parity>) -> Self {
```
ianrrees, line 833: Clever reuse of `None`, but IMHO it's a little bit awkward to use :)

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1531 @@
...
+    type Word = u16;
+    const BITS: u8 = 0x1;
+}
+
+//=============================================================================
+// Stop bits, parity, baud rate
+//=============================================================================
+
+/// Number of stop bits in a UART frame
+pub enum StopBits {
```
ianrrees, line 440: To implement that USB-UART, I had to add `#[derive(Copy, Clone)]` to `StopBits` and `Parity`.  It also might be nice to have a way to read the current settings out of a Uart if that's not too much trouble to implement.

bradleyharden:
@jbeaurivage, I really like the new types you added. A few comments though.

I think it might make sense to rename many of the Registers functions to `set_*`, to mirror the `get_*` names. I wouldn't rename any of the methods on `Config` or `Reconfig`, though, because they are optimized for specifying a configuration.

I'm also not sure about making `ReadConfig` a trait. It's something users would have to know to import. It's just an extra, non-obvious step. Is there a reason you can't just use inherent methods? Just more boilerplate?

jbeaurivage:
> @jbeaurivage, I really like the new types you added. A few comments though.
> 
> I think it might make sense to rename many of the Registers functions to `set_*`, to mirror the `get_*` names. I wouldn't rename any of the methods on `Config` or `Reconfig`, though, because they are optimized for specifying a configuration.
> 
> I'm also not sure about making `ReadConfig` a trait. It's something users would have to know to import. It's just an extra, non-obvious step. Is there a reason you can't just use inherent methods? Just more boilerplate?

@bradleyharden no problem on set_* vs get_*. `ReadConfig` is a trait just to avoid boilerplate, otherwise the methods provided would need to be re-exported to `Config`, `Reconfig` and `Uart`. Plus, I figure that reading the USART's current configuration is probably uncommon enough that bringing a trait in scope is acceptable, most use cases are just "set up the peripheral then use it". The trait is mentioned in the module comments though.

bradleyharden:
I just had an idea.

We want the `Config` struct to have a fluent interface that takes and returns `self`, because we have two methods, `char_size` and `enable` that need to take `self`. We have the rest of the methods take and return `self` only for the sake of ergonomics. Normally, they would just take `&mut self` and return `()`. What if we added **both** approaches to `Config`?

```rust
impl<P, C> Config<P, C> {
    pub fn get_bit_order(&self) -> Bitorder { ... }
    pub fn set_bit_order(&mut self, bit_order: BitOrder) { ... }
    pub fn bit_order(self, bit_order: BitOrder} -> Self { ... }
}
```

That would do two things:
- It would preserve the fluent interface with the `self` methods.
- It would get rid of the `Reconfig` struct, because you could give out an `&mut Config`, which you could only use for the `set_*` methods.

I think the only potential downside is slight naming confusion.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,400 @@
...
+use pac::sercom0::usart::ctrla::MODE_A;
+
+#[cfg(feature = "min-samd51g")]
+use pac::sercom0::usart_int::ctrla::MODE_A;
+
+use crate::time::Hertz;
+
+pub struct Registers<S: Sercom> {
+    sercom: S,
+}
```
bradleyharden, line 18: Since you've removed interior mutability from the PAC `SERCOM` struct, you can `unsafe impl Sync for Registers`, even though the PAC struct isn't `Sync`.

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1741 @@
...
+    /// Change the [`CharSize`] to [`DynCharSize`]. The UART's character
+    /// size will be changed to the provided `C2`, which is a [`FixedCharSize`]
+    /// type, and can then be changed dynamically on an enabled [`Uart`]
+    /// without changing the underlying [`Config`]'s type through the
+    /// [`reconfigure`](Uart::reconfigure) method.
+    #[inline]
+    pub fn dyn_char_size<C2: FixedCharSize>(mut self) -> Config<P, DynCharSize> {
+        self.registers.configure_charsize(C2::BITS);
+        self.change()
+    }
```
bradleyharden, line 977: This seems weird to me. You're setting it to `DynCharSize`, but you still want a compile-time known type to use for the actual value? What if I don't know the value until run-time? Shouldn't there be a `CharSizeEnum` or something?

I let users set `DynLength` like any other `Length` value, with the `.length::<L2>()` method. But then I added `get_dyn_length` and `set_dyn_length` methods that are only implemented when the `Config` is in `DynLength` mode. I think that approach might make more sense here.

The only difference is that `DynLength` can be set to anything from 1-255, so a `u8` works well to represent it at run-time. But a value-level version of `CharSize` should probably be an `enum`.

bradleyharden:
@jbeaurivage, when I refactored the `spi` module, I moved all remaining pieces of it out of the `thumbv*` modules. Instead, I created two different modules in the new, consolidated `spi` module. You can see the directory structure [here](https://github.com/bradleyharden/atsamd/tree/refactor_v2_spi/hal/src/sercom/v2/spi).

What do you think of this approach? I like it better. I actually went a step further and moved the `impl_pad` modules to the main `sercom` module as well. Now, the only modules left in the `thumbv*` modules are the `v1` modules.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,400 @@
...
+            }
+            Parity::Even => {
+                self.usart().ctrlb.modify(|_, w| w.pmode().bit(false));
+                true
+            }
+        };
+
+        self.usart()
+            .ctrla
+            .modify(|_, w| unsafe { w.form().bits(!enabled as u8) });
```
ianrrees, line 119: There's a stray `!` here; form 0 has no parity and form 1 has parity.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    /// Get the current parity setting
+    #[inline]
+    pub(super) fn get_parity(&self) -> Parity {
+        let enabled = self.usart().ctrla.read().form().bits() != 0;
```
ianrrees, line 125: Just thinking about our future selves, it might be better if this checked for the format being either 0x1 (USART w/ parity) or 0x5 (auto baud w/ parity).  If auto-baud w/o parity were enabled (form 0x4), this would return the wrong result.

ianrrees (changes requested):

jbeaurivage:
> @jbeaurivage, when I refactored the `spi` module, I moved all remaining pieces of it out of the `thumbv*` modules. Instead, I created two different modules in the new, consolidated `spi` module. You can see the directory structure [here](https://github.com/bradleyharden/atsamd/tree/refactor_v2_spi/hal/src/sercom/v2/spi).
> 
> What do you think of this approach? I like it better. I actually went a step further and moved the `impl_pad` modules to the main `sercom` module as well. Now, the only modules left in the `thumbv*` modules are the `v1` modules.

I think that's a great idea. I followed suit for this PR.

hal/src/sercom/v2/uart/mod.rs
jbeaurivage: You're right, that was weird. I changed the `set_dyn_char_size` method to take a `CharSizeEnum`, and added a `get_dyn_char_size` method.

TDHolmes:
For 36873cee37c6931b1fbfc315eff7ab62c4eb8289, you could patch the PAC instead of working around it in the HAL

TDHolmes:
Check out #428 for a reference on how

hal/src/sercom/v2/uart/charsize.rs
```diff
@@ -0,0 +1,114 @@
...
+        match item {
+            FiveBit::BITS => CharSizeEnum::FiveBit,
+            SixBit::BITS => CharSizeEnum::SixBit,
+            SevenBit::BITS => CharSizeEnum::SevenBit,
+            EightBit::BITS => CharSizeEnum::EightBit,
+            NineBit::BITS => CharSizeEnum::NineBit,
+            _ => unreachable!(),
+        }
+    }
+}
```
bradleyharden, line 72: I don't know that you need this conversion. And if you do, I wouldn't implement it with `From`. I would use `TryFrom`. The `unreachable!` option is certainly not unreachable.

In the `spi` module, I realized that the only time I needed to convert from the `enum` to the raw bits is inside the corresponding `reg` function. So rather than abstract it into a trait that is only used once, I kept it isolated to those functions.

hal/src/sercom/v2/uart/charsize.rs
```diff
@@ -0,0 +1,114 @@
...
+
+/// `enum` version of [`CharSize`]
+#[repr(u8)]
+pub enum CharSizeEnum {
+    FiveBit = FiveBit::BITS,
+    SixBit = SixBit::BITS,
+    SevenBit = SevenBit::BITS,
+    EightBit = EightBit::BITS,
+    NineBit = NineBit::BITS,
+}
```
bradleyharden, line 59: I think this is a bit roundabout. You define the bits on the type, then use the type to define the enum, then use the enum to get back the bits. I would switch things around.

What if you defined the variants of `CharSizeEnum` directly? Then then `FixedCharSize` trait can define a `const CHAR_SIZE: CharSizeEnum;`. That would also let you modify the corresponding functions in the `reg` module to take a `CharSizeEnum` rather than raw bits. It's a bit dangerous right now, because the `reg` function does not check the bits at all.

hal/src/sercom/v2/uart/charsize.rs
```diff
@@ -0,0 +1,114 @@
...
+    type Word = u16;
+    const BITS: u8 = 0x1;
+}
+impl FixedCharSize for NineBit {}
+
+impl Sealed for DynCharSize {}
+impl CharSize for DynCharSize {
+    type Word = u16;
+    // Irrelevant for DynCharSize
+    const BITS: u8 = 0x0;
```
bradleyharden, line 113: By moving the `CHAR_SIZE` constant to `FixedCharSize`, you remove the need for this.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,411 @@
...
+    pub(super) fn configure_pads(&mut self, rxpo: u8, txpo: u8) {
+        self.usart().ctrla.modify(|_, w| unsafe {
+            w.rxpo().bits(rxpo);
+            w.txpo().bits(txpo)
+        });
+    }
+
+    /// Configure the character size
+    #[inline]
+    pub(super) fn set_char_size(&mut self, bits: u8) {
```
bradleyharden, line 84: The other functions take the corresponding `enum`, but not this one. The changes I proposed in the `char_size` module should help that.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,411 @@
...
+                    });
+                }
+            }
+        };
+    }
+
+    /// Get the contents of the `BAUD` register and the current baud mode. Note
+    /// that only the CONTENTS of `BAUD` are returned, and not the actual baud
+    /// rate. Refer to the datasheet to convert the `BAUD` register contents
+    /// into a baud rate.
```
bradleyharden, line 232: Just out of curiosity, why not do the conversion for users? Do you think the raw pieces are more useful? Or would the actual frequency be more useful? I'm imagining a situation where you might want to set the desired baud rate, then read it back to see how close the actual baud rate is.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,411 @@
...
+
+    /// Get the current IrDA encoding setting. The return type is the pulse
+    /// length wrapped in an [`Option`].
+    #[inline]
+    pub(super) fn get_irda_encoding(&self) -> Option<u8> {
+        if !self.usart().ctrlb.read().enc().bit() {
+            return None;
+        }
+
+        Some(self.usart().rxpl.read().bits())
```
bradleyharden, line 296: Minor thing, but you could use an expression here, i.e.

```rust
        if self.usart().ctrlb.read().enc().bit() {
            Some(self.usart().rxpl.read().bits())
        } else {
            None
        }
```

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    /// Clear specified interrupt flags
+    #[inline]
+    pub fn clear_flags(&mut self, bits: u8) {
```
bradleyharden, line 301: I think the functions you copied from me are marked `pub`, but your own functions are marked `pub(super)`. Because the type is marked `pub(super)`, I don't think it matters, but I'm not 100% sure. Either way, I think it might make sense to make them all uniform. I don't really care which though.

hal/src/sercom/v2/uart/config.rs
```diff
@@ -0,0 +1,480 @@
...
+        self.change()
+    }
+
+    /// Change the [`CharSize`] to [`DynCharSize`]. The UART's character
+    /// size will be changed to the provided `C2`, which is a [`FixedCharSize`]
+    /// type, and can then be changed dynamically on an enabled [`Uart`]
+    /// without changing the underlying [`Config`]'s type through the
+    /// [`reconfigure`](Uart::reconfigure) method.
+    #[inline]
+    pub fn dyn_char_size<C2: FixedCharSize>(mut self) -> Config<P, DynCharSize> {
```
bradleyharden, line 150: Hmm. I still find this a bit strange. In one function, you both set the `Config` struct to `DynCharSize` mode and also set the actual character size. To me, it would make more sense for the `dyn_char_size` function to change it to `DynCharSize` mode *only*, leaving it with the default character size (`CharSizeEnum::EightBit`). Then a separate call to `set_dyn_char_size` would actually set the desired character size. That way, if you're using `DynCharSize`, you only need to interact with `CharSizeEnum` and never the `FixedCharSize` types.

Thoughts?

hal/src/sercom/v2/uart/config.rs
```diff
@@ -0,0 +1,480 @@
...
+
+impl<P: ValidPads> Config<P, DynCharSize> {
+    /// Dynamically change the character size
+    #[inline]
+    pub fn set_dyn_char_size(&mut self, char_size: CharSizeEnum) {
+        self.registers.set_char_size(char_size as u8);
+    }
+
+    /// Get the current character size setting
+    pub fn get_char_size(&self) -> CharSizeEnum {
```
bradleyharden, line 387: I think you used `get_char_size` instead of `get_dyn_char_size`, because it could be used to read the character size even with a `FixedCharSize`. But I don't know that it's worth it to break the naming symmetry for that. If your character size is a fixed value at compile time, why would you ever need to read it back? Seems like we'll send people searching for the correct function name more often than we'll provide a needed API.

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1005 @@
+//! Use the SERCOM peripheral for UART communications
```
bradleyharden, line 1: By the way, I plan to propose that we move away from using `mod.rs`. I have a few reasons:

1. When you have multiple `mod.rs` files open in the same editor, it becomes difficult to know which file you are looking at. Taken to it's extreme, this approach would name every single module `mod.rs`, which doesn't really make sense.
2. If you have a preexisting module without any sub-modules and you add a new sub-module, the `mod.rs` approach requires that you rename the module and move it to a sub-directory. The other approach does not.
3. With the new approach, the directory hierarchy exactly matches the module hierarchy. Sub-modules of a given module are always located one directory level below the module.

The only downside is that it requires a slight change in perspective. Whereas the `uart/` directory *used* to contain the **contents** of the module, it now only contains the **sub-modules** of the module. But once you get used to it, I find that approach much more intuitive.

I discussed this issue with @vccggorski and @vcchtjader, and they agreed with it at the time. I haven't discussed it with anyone else though. I should probably make an issue.

The HAL mixes both approaches right now, so it's not a priority to change it. But I thought I should bring it up, since you took the time to separate out your module into neat little pieces. What do you think?

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1005 @@
...
+
+impl From<StopBits> for bool {
+    #[inline]
+    fn from(item: StopBits) -> bool {
+        match item {
+            StopBits::OneBit => false,
+            StopBits::TwoBits => true,
+        }
+    }
+}
```
bradleyharden, line 438: Do you think users will actually use these trait definitions at all? It seems like they are only relevant for setting the registers. In that case, they would only be used once within the HAL and nowhere else. Would it make sense to move them there?

Do you see any use cases I'm missing?

hal/src/sercom/v2/uart/mod.rs
```diff
@@ -0,0 +1,1005 @@
...
+    #[inline]
+    pub fn flush_rx_buffer(&mut self) {
+        // TODO
+        // The datasheet states that disabling the receiver (RXEN) clears
+        // the RX buffer, and clears the BUFOVF, PERR and FERR bits.
+        // However, in practice, it seems like BUFOVF errors still pop
+        // up after a disable/enable cycle of the receiver, then immediately begin
+        // reading bytes from the DATA register.
+        //
+        // Is this a hardware bug???
```
bradleyharden, line 967: I think you should elevate this from a comment to part of the documentation. It seems important enough to tell users about.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,411 @@
...
+    #[inline]
+    pub fn clear_status(&mut self, bits: u16) {
+        self.usart().status.modify(|_, w| unsafe { w.bits(bits) });
+    }
+
+    /// Read status flags
+    #[inline]
+    pub fn read_status(&self) -> u16 {
+        self.usart().status.read().bits()
+    }
```
bradleyharden, line 333: I don't think I follow my own advice here, but I just realized that it might be better for these functions to take `Flags` and `Status` types, rather than raw bits.

Right now, the `unsafe` code has to rely on the caller to provide valid values, which means it could technically be unsound if a HAL author made a mistake. But if you used the `Flags` and `Status` types, it would make the check entirely local to the function and truly safe.

I think it would clean up some of the downstream code too.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -0,0 +1,411 @@
...
+
+#[cfg(any(feature = "samd11", feature = "samd21"))]
+use pac::sercom0::usart::ctrla::MODE_A;
+
+#[cfg(feature = "min-samd51g")]
+use pac::sercom0::usart_int::ctrla::MODE_A;
+
+use crate::time::Hertz;
+
+pub struct Registers<S: Sercom> {
```
bradleyharden, line 16: Actually, it looks like you made the type `pub`, but the module is private. Again, I think consistency is the important thing. Either approach should be fine.

hal/src/sercom/v2/uart/impl_ehal.rs
```diff
@@ -0,0 +1,69 @@
...
+where
+    C: ValidConfig,
+    D: Transmit,
+{
+    type Error = core::convert::Infallible;
+
+    /// Wait for a `DRE` flag, then write a word
+    #[inline]
+    fn write(&mut self, word: C::Word) -> nb::Result<(), Self::Error> {
+        // Ignore buffer overflow errors
```
bradleyharden, line 41: I think you copied a comment that doesn't make sense for UART.

hal/src/sercom/v2/uart/impl_ehal.rs
```diff
@@ -0,0 +1,69 @@
...
+            Ok(())
+        } else {
+            Err(WouldBlock)
+        }
+    }
+
+    /// Wait for a `TXC` flag
+    #[inline]
+    fn flush(&mut self) -> nb::Result<(), Self::Error> {
+        // Ignore buffer overflow errors
```
bradleyharden, line 53: Same here

hal/src/sercom/v2/uart/charsize.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::typelevel::Sealed;
+use num_traits::{AsPrimitive, PrimInt};
+
+/// Size of the SERCOM's `DATA` register
+#[cfg(any(feature = "samd11", feature = "samd21"))]
+pub type DataSize = u16;
+
+/// Size of the SERCOM's `DATA` register
+#[cfg(any(feature = "min-samd51g"))]
+pub type DataSize = u32;
```
bradleyharden, line 12: Seems like this belongs somewhere else. I'm not sure where though. Maybe the top level module? It doesn't affect the character size though.

hal/src/sercom/v2/uart/pads_thumbv6m.rs
```diff
@@ -0,0 +1,486 @@
...
+    const TXPO: u8 = <(RX::PadNum, TX::PadNum, RTS::PadNum, CTS::PadNum)>::TXPO;
+}
+
+//=============================================================================
+// Implement RxpoTxpo
+//=============================================================================
+
+/// Filter [`PadNum`] permutations and implement [`RxpoTxpo`]
+    #[rustfmt::skip]
+    macro_rules! impl_rxpotxpo {
```
bradleyharden, line 59: Does UART have loopback mode, like SPI? I recently modified this macro for SPI to avoid accidentally using loopback mode when one of the pads is `NoneT`. If that matters at all to you, see the updated version in #467.

hal/src/sercom/v2/uart/charsize.rs
bradleyharden: I realized a better name for this might be DataReg. I'm going to change it in the spi module.

I think DataSize sounds too related to CharSize. And in the spi module, I have a Size trait that could also be confused.

The intent is for this alias to represent the data register, so I think DataReg probably makes the most sense while also being unambiguous. Thoughts?

hal/src/sercom/v2/uart/pads_thumbv6m.rs
ianrrees: Out of curiosity, I looked at the SAMD21 datasheet and it doesn't explicitly say that the UART supports loopback like for SPI, but it appears that RXPO can be set to be set to any of the pads so it probably does support a loopback so long as you're not using hardware flow control.

hal/src/sercom/v2/uart/charsize.rs
jbeaurivage: Would you say panicking is acceptable behaviour if we encounter an invalid bit pattern? Normally this shouldn't happen as the driver guarantees a valid pattern should be in the register field.

hal/src/sercom/v2/uart/config.rs
jbeaurivage: Yeah I think that makes perfect sense.

hal/src/sercom/v2/uart/charsize.rs
bradleyharden: > Normally this shouldn't happen as the driver guarantees a valid pattern should be in the register field.

I guess that's my point. It sounds like you're saying that this code should only be used by the driver, because it can guarantee that the given bit pattern is valid before the conversion. But if that's the case, the conversion probably shouldn't be implemented as `From`, because that gives users access as well. A user could easily provide an invalid bit pattern and get to the `unreachable!` branch.

I was arguing that this code probably belongs in the `reg` module. The user-facing type is `CharSizeEnum`, and users shouldn't have to know or care about its corresponding hardware bit representation. And once it's in the `reg` module, you can truly guarantee that the branch is `unreachable!`.

bradleyharden: By "in the `reg`" module, I mean as part of one of the `Registers` methods.

jbeaurivage: Okay, yeah, that makes total sense.

hal/src/sercom/v2/uart/mod.rs
jbeaurivage: I don't really have preference either way. At first I thought it was weird that module files were called `mod.rs`. I got used to it, so I don't really give it much thought anymore. Though I can definitely see the argument for avoiding `mod.rs`. I'll update the PR to reflect that.

hal/src/sercom/v2/uart/charsize.rs
jbeaurivage: Yup, that works!

hal/src/sercom/v2/uart/pads_thumbv6m.rs
jbeaurivage: From the datasheet:
> 25.6.3.6 Loop-back Mode
By configuring the Receive Data Pinout (CTRLA.RXPO) and Transmit Data Pinout (CTRLA.TXPO) to use the same data
pins for transmit and receive, loop-back is achieved. The loop-back is through the pad, so the signal is also available
externally.

hal/src/sercom/v2/uart/mod.rs
jbeaurivage: No, you're right. Removed them.

hal/src/sercom/v2/uart/reg.rs
jbeaurivage: The only reason I return the raw contents of the `BAUD` register is to avoid having to hold the underlying GCLK frequency in `Config`. I'm trying to avoid synchronization issues where the clock frequency gets changed from under the UART. Thankfully this should be impossible when `clock::v2` gets here.

bradleyharden: Yes, with `clock::v2`, the `Config` will hold a `Pclk` (peripheral channel clock) struct. Holding that struct will guarantee that the GCLK doesn't change, and it also contains a copy of the GCLK frequency.

Maybe you can update this once `clock::v2` lands.

hal/src/sercom/v2/uart/pads_thumbv6m.rs
bradleyharden: This was a problem for me, because if you create an SPI with `Rx` `Capability`, you still need to send dummy data to initiate the transfer. But if you were accidentally in loop-back mode, you would be reading back that dummy data.

I'm not sure if that's an issue for you though. I guess not? If you're `Rx`, then you're never sending data, so you can't read it back. But would you accidentally have contention on the line? If the SAMD transmit pin is trying to hold the line high, for idle, but the sender on the other end is trying to bring it low, for a transfer, would you have a problem?

bradleyharden: Here is where I corrected it:

https://github.com/bradleyharden/atsamd/blob/3d9294b6d7f87b2fd5eae9d762f2bc73fca1b8b9/hal/src/sercom/v2/spi/pads_thumbv6m.rs#L103

jbeaurivage: Well, I don't *think* it would create any problems. However I'd rather be proactive and avoid potential problems in the future, so I'm going to avoid entering loopback mode by only enabling `RXEN`/`TXEN`/Both depending on the `Capability`. Right now both unconditionally get enabled.

PR #445, opened 2021-05-26
Added Vdiv (A6) as pin b1

ewwaller:
The Feather-M4 has a 1:2 resistor divider from VBat to what anrduino
considers A6 (mapped to port b 1).  This permits the battery voltage
to be monitored using the ADC.

### DIFF
--- a/boards/feather_m4/src/lib.rs
+++ b/boards/feather_m4/src/lib.rs
@@ -48,7 +48,9 @@ define_pins!(
     pin a4 = a4,
     /// Analog Pin 5
     pin a5 = a6,
-
+	/// Analog Vdiv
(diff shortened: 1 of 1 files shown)


ewwaller:
This change adds a VBat monitor on the Feather M4 board.  This pin is not broken out, but is internally tied to port b1 through  divide by two resistor newwork.

sajattack:
I think the other BSPs are calling these sorts of pins "battery", so we should probably keep it consistent.

bradleyharden:
This will be handled in #487. I'm going to close.

TDHolmes:
@ewwaller  sorry we missed this until it became too hard to merge due to a refactor

PR #449, opened 2021-06-07
Allow setting ADC prescaler and resolution

DCRichards:
## Motivation

Based on [this conversation](https://matrix.to/#/!JbgTnIGpDxEHkjzSep:gitter.im/$26l-B8HhoAf-XxsL8re4djF-77b6vxz5VEfODJCFQrw?via=matrix.org), adding the ability to configure ADC for boards with different prescaling and resolution. On the [Arduino Nano 33 IOT](https://github.com/arduino/ArduinoCore-samd/blob/master/cores/arduino/wiring.c#L103), for example, the prescaler is 1/512 with a 10bit resolution rather than the default 1/32 and 12bit.

## Changes

* Added `prescaler(&mut self, prescaler: Prescaler)` and `resolution(&mut self, resolution: Resolution)` functions for setting ADC resolution and prescaler to allow changing these for devices which require different configurations.
* Added exported aliases for ADC settings.
* Added doc comments for clarity.
* Update Arduino Nano 33 IOT README.md to clarify steps and add Arduino CLI steps seeing as I've been working with it recently.


hal/src/thumbv6m/adc.rs
```diff
@@ -59,18 +68,37 @@ impl Adc<ADC> {
...
+    /// Set the voltage reference source
     pub fn reference(&mut self, reference: adc::refctrl::REFSEL_A) {
         self.adc
             .refctrl
             .modify(|_, w| w.refsel().variant(reference));
         while self.adc.status.read().syncbusy().bit_is_set() {}
     }
 
+    /// Set the prescaler for adjusting the clock relative to the system clock
+    pub fn prescaler(&mut self, prescaler: adc::ctrlb::PRESCALER_A) {
```
bradleyharden, line 86: I would suggest adding a type alias for the `PRESCALAR_A` type, so that users can import it from this module. Otherwise they'll have to import it from the PAC directly. You can also give it a better name and one that is more consistent with Rust naming conventions.

Maybe you could do the same for the other types in this module as well.

boards/arduino_nano33iot/README.md
```diff
@@ -1,32 +1,42 @@
...
-       - **OSX**: `ArduinoData` is likely something like `~/Library/Arduino15`
-    - Probably best to install an example sketch via the IDE just to make sure everything is working
-    - Note that the [arduino cli](https://github.com/arduino/arduino-cli) (or just regular bossac) may soon replace this section
- - arm-none-eabi tools installed, you need gcc and objcopy.
-   -  **Note**: Alternatively, you can use [cargo-binutils](https://github.com/rust-embedded/cargo-binutils), which is likely easier to install on OSX and also easier to use, as it will automatically detect the target
- - thumbv6m-none-eabi rust target installed via `rustup target add thumbv6m-none-eabi`
-
-#### Steps
+
+* Blinky basic
```
sajattack, line 7: I don't like having the examples listed in the README. Too burdensome to update if new ones get added.

DCRichards: One was already listed so just tidied it up, but I agree, it's easy enough to find them so no need to list

sajattack:
Could you do v7 too? I don't want them to get too out of sync in functionality. I can test it if you don't have a board.

DCRichards:
> Could you do v7 too? I don't want them to get too out of sync in functionality. I can test it if you don't have a board.

Sure, AFK for the weekend but will get to it in the week

PR #452, opened 2021-06-15
Remove DMAC Transfer payloads

jbeaurivage:
This PR removes the `payload` concept from DMAC `Transfers`. It was initially intended to tie in a PAC/peripheral struct while a Transfer is ongoing. Since then, the design has evolved, and the peripherals themselves implement `Buffer`. Therefore they're already moved into a `Transfer`, rendering the use of `payload`s redundant.

PR also fixes doc warnings for the `dmac` module, and re-exports all public items from its submodules for slightly better ergonomics.


bradleyharden:
This looks good to me. @sajattack?

sajattack:
> This looks good to me. @sajattack?

DMA has always been an area I'm not too knowledgable about, the code looks fine afaict.

PR #453, opened 2021-06-15
Add section to README: How to use a BSP

nQue:
I was advised by @jbeaurivage to add a section to the readme, detailing the steps I found difficult as a newbie getting started.
This is my first Github pull request ever.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,29 @@ $ ./build-all.py
 
 Please note that this script requires Python 3.
 
-## Getting code onto the device: Gemma M0
+## How to use a BSP (i.e. getting started writing your own code)
+
+To bootstrap your own project you should be able to copy/paste the Rust code from the examples folder within the folder of the BSP you've chosen. But you shouldn't copy the `Cargo.toml` file from there, since that's not only used for the examples, but also for the whole BSP itself. You want to make your own `Cargo.toml` file. If you're new to this and have no clue what you're doing then this is probably the line you want in there:
+
+```rust
+[dependencies]
+feather_m0 = { git = "https://github.com/atsamd-rs/atsamd" }
+```
+
+Replace "feather_m0" with the name of the BSP you want to use. Note a few things:
+* By specifying the dependency as a remote git repo without specifying a branch, rather than pinning a specific version, we ask Cargo to always grab the latest master branch from Github. This is probably what you want in the beginning, at least now when the project is young, but expect that you'll later want to pin down a specific version.
+* There's a whole bunch of crates inside that git repo, and Cargo automatically figures out which one you want to use by checking what you're naming the dependency. In this case we named it "feather_m0" so Cargo will try to find a BSP called that, somewhere inside the repo.
+
+Now the imports in your Rust code should work, if they look something like this:
+```rust
+use feather_m0 as bsp;
+use bsp::hal;
+use bsp::pac;
+```
+
+You should now have objects called `bsp` and `hal` and `pac`, which contain all the good stuff.
+
+## Getting code onto the device: Adafruit M0/M4 board (such as Gemma M0 & Feather M0)
 
 If you want to flash the device using the tools that come with the Adafruit arduino support package:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -118,9 +140,13 @@ This same technique should work for all of the Adafruit M0/M4 boards, as they al
 
 ## Getting code onto the device: JLink
 
-If you have a board with a SWD debug header, such as the [Metro M0][metro_m0], or if you attached the header yourself, you can  use your JLink together with gdb. @wez prefers using the JLinkGDBServer, but you can also use OpenOCD.
+If you have a board with a SWD debug header, such as the [Metro M0][metro_m0], or if you attached the header yourself, you can use your JLink programmer to either flash the device or debug it (together with gdb).
+
+You can use JFlashLiteExe to just flash the .bin file (note that for some boards, such as [Feather M0][feather_m0], you need to specify the program memory starting address offset or you'll just get a nondescript flashing error).
+
+To debug your board you can run JLinkGDBServer instead of JFlashLiteExe. @wez prefers using the JLinkGDBServer, but you can also use OpenOCD. Note that starting a debug session normally triggers a flashing, so you don't need a separate flashing step before this.
```
bradleyharden, line 147: Isn't this only true if the `load` command is in your .gdbinit?

README.md
```diff
@@ -97,7 +97,29 @@ $ ./build-all.py
...
 Please note that this script requires Python 3.
 
-## Getting code onto the device: Gemma M0
+## How to use a BSP (i.e. getting started writing your own code)
+
+To bootstrap your own project you should be able to copy/paste the Rust code from the examples folder within the folder of the BSP you've chosen. But you shouldn't copy the `Cargo.toml` file from there, since that's not only used for the examples, but also for the whole BSP itself. You want to make your own `Cargo.toml` file. If you're new to this and have no clue what you're doing then this is probably the line you want in there:
+
+```rust
+[dependencies]
+feather_m0 = {  git = "https://github.com/atsamd-rs/atsamd", features = ["unproven"]  }
```
bradleyharden, line 106: @nQue, maybe you could add `unproven` as a default feature for the HAL in this commit? I think that would be helpful, and perhaps a good exercise for someone new to Rust.

Separately, you might want to clarify that this pulls the BSP from `master`. I'm not sure if that's the best thing for new users. It's probably the best choice right now, given the flux of things. But it may not be the best choice in the future.

nQue: @bradleyharden Sure I could add that, but are you sure that's a good idea? It sounds kinda fishy. Are you one of the main developers of this repo? I say it seems fishy because it would hide that they're unproven, thus kinda removing the reason for having them labeled as "unproven" to begin with. We might as well just remove the "unproven" tag from them? That sounds like a less fishy fix :)

README.md
nQue: Ah, true. I'll edit that.

README.md
nQue: @bradleyharden Oh wait, now I see that you suggested it in the Matrix chat, with a motivation, and that @jbeaurivage thumbed-upped it, so then I feel confident to add it :)

bradleyharden: @nQue, I think your instincts are good, but this case is a bit different. `embedded-hal` never removed the `unproven` feature gate for several traits in `v0.2.0+`. Since then, they jumped to alpha releases of `v1.0`. We plan to adopt `v1.0` once it's released, but we want to wait for it to stabilize first. In the meantime, *many* features of this HAL depend on the `unproven` feature, which is frustrating. Ultimately, I think it will be easier for everyone if we make `unproven` a default feature.

@sajattack, do you agree?

bradleyharden:
@nQue, do you know about the [cortex-m-quickstart](https://github.com/rust-embedded/cortex-m-quickstart) template? When I first got started with embedded Rust, that's what I started from. I'm not sure if that appears anywhere in the README. If not, it probably should.

bradleyharden:
This looks good to go, pending any comments from @sajattack. Adding a reference to the quickstart template might be nice, but isn't required.

README.md
Sympatron: @nQue bradleyharden is one of the maintainers.
I agree that `unproven` should be standard given how unusable a lot of this HAL is without it and how long these traits have been "unproven".

While publishing the next release this example should be updated to point to crates.io instead IMO.

sajattack:
Sorry didn't see this.

Issue #397, opened 2021-02-01
USB methods disable interrupts for too long(?)


ianrrees:
I'm working on a SAMD21 based USB device, which does the USB work in the USB ISR, along the lines of the usb_echo example.  The USB interrupt is set to a lower priority (higher number) than another interrupt which needs to be serviced fairly quickly, but sometimes that doesn't happen quickly enough.  The issue seems to be that `UsbBus`'s methods do all their work in interrupt free contexts.  For example, [these](https://github.com/atsamd-rs/atsamd/blob/4816bb13a12a604e51f929d17b286071a0082c82/hal/src/samd21/usb/bus.rs#L1046-L1056).

Might it make sense for `UsbBus` to have an "in use" flag in the mutex rather than the RefCell, so that the interrupts only need to be disabled for setting/checking that flag?

twitchyliquid64:
Can you elaborate on 'too long'? We tried to do as little as possible in the `UsbBus` inner implementation.

>  The USB interrupt is set to a lower priority (higher number) than another interrupt which needs to be serviced fairly quickly

Depending on how many priority bits your chip has, could you boop the USB to have less priority than your other CSR?

> Might it make sense for UsbBus to have an "in use" flag in the mutex rather than the RefCell, so that the interrupts only need to be disabled for setting/checking that flag?

I'm not sure i follow, can you give me an example?

ianrrees:
> Can you elaborate on 'too long'?

Maybe more importantly than the specific timing constraint, I was surprised to find that the entirety of `read()` and `write()` is in an interrupt free block since they call `copy_to_nonoverlapping()`.  Isochronous endpoints can have up to 1023B per read/write, so there would be a pretty wide range of time that interrupts will be disabled.

Specifically, I've got about a 16us window to respond to a DMA interrupt, and I believe that response usually takes about 3us so I should be able to tolerate about 12us of interrupts being disabled.  Admittedly I haven't put much effort in to profiling - an early test showed a case where the DMA interrupt wasn't handled when the processor was in the USB ISR, so I went digging to see how that could happen...

> could you boop the USB to have less priority than your other CSR

Yep, I'm pretty sure that is the current arrangement, but perhaps got something wrong.  Will investigate further.

> I'm not sure i follow, can you give me an example?

[...]

PR #456, opened 2021-06-22
Add methods for the USB Start Of Frame interrupt

ianrrees:
After using https://github.com/atsamd-rs/atsamd/pull/398 a bit, I found that checking the SOF flag in `poll()` was problematic, because from the outside it's not clear whether or not the SOF flag was set.

This functionality is helpful in the context of USB classes which are always accessed in the USB ISR (working around https://github.com/atsamd-rs/atsamd/issues/397).  If the host isn't read/writing to the endpoint, but the device has info it wants to send, then somehow execution needs to get back in to the USB ISR.

Since the PR is so small, and the original one hasn't been updated in a while, I've made a new PR.  If that's problematic, I'd be happy to re-do this as a PR against https://github.com/henrikssn/atsamd/tree/usb-sof

### DIFF
--- a/hal/src/thumbv6m/usb/bus.rs
+++ b/hal/src/thumbv6m/usb/bus.rs
@@ -717,6 +717,15 @@ impl Inner {
         usb.ctrlb.modify(|_, w| w.detach().clear_bit());
     }
 
+    /// Enables/disables the Start Of Frame (SOF) interrupt
+    fn sof_interrupt(&self, enable: bool) {
+        if enable {
+            self.usb().intenset.write(|w| w.sof().set_bit());
+        } else {
+            self.usb().intenclr.write(|w| w.sof().set_bit());
+        }
+    }
+
     /// Configures all endpoints based on prior calls to alloc_ep().
     fn flush_eps(&self, mode: FlushConfigMode) {
         for idx in 0..8 {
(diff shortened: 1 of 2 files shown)


ianrrees:
Whoops!  I forgot to add the change to thumbv7, so have updated the PR.

bradleyharden:
@sajattack, I haven't worked with USB at all. Do you want to take a look at this?

sajattack:
Me neither tbh. @twitchyliquid64 ?

twitchyliquid64 (approved):

hal/src/thumbv7em/usb/bus.rs
```diff
@@ -663,6 +663,15 @@ impl Inner {
         usb.ctrlb.modify(|_, w| w.detach().clear_bit());
     }
 
+    /// Enables/disables the Start Of Frame (SOF) interrupt
+    fn sof_interrupt(&self, enable: bool) {
+        if enable {
+            self.usb().intenset.write(|w| w.sof().set_bit());
+        } else {
+            self.usb().intenset.write(|w| w.sof().clear_bit());
```
jbeaurivage, line 671: Same here

hal/src/thumbv6m/usb/bus.rs
```diff
@@ -717,6 +717,15 @@ impl Inner {
         usb.ctrlb.modify(|_, w| w.detach().clear_bit());
     }
 
+    /// Enables/disables the Start Of Frame (SOF) interrupt
+    fn sof_interrupt(&self, enable: bool) {
+        if enable {
+            self.usb().intenset.write(|w| w.sof().set_bit());
+        } else {
+            self.usb().intenset.write(|w| w.sof().clear_bit());
```
jbeaurivage, line 725: I think you might want `self.usb().intenclr.write(|w| w.sof().set_bit());` here instead. Per the datasheet, writing a zero to the `INTENSET` or `INTENCLR` registers' bits has no effect.

ianrrees: Good catch, thanks!  I'll update the PR shortly.

PR #459, opened 2021-06-24
Rework USB OUT endpoint interrupts

ianrrees:
(this PR has been heavily edited from the original - sorry)

Previously, when the USB interrupt was unmasked and a host issued an OUT transfer to an endpoint, the USB interrupt would continually fire until `read()` was called on that endpoint.  This could lead to a hang if the USB class controlling the endpoint was dependent on some lower-priority task before it could successfully read.  For example: if the class received data over USB to feed out to an interrupt-driven serial port, the class might have a full buffer so not call the endpoint `read()`, but the buffer would not drain because the serial port interrupt was lower priority than the USB one.  This behaviour came about because the USB `poll()` method didn't clear the endpoint interrupts for OUT transfers, but relied on the `read()` methods on endpoints to clear them.

The endpoints have a non-interrupting flag that indicates when the endpoint is ready to be read, and this change switches to using that to indicate data is available.

One subtle consequence of this change is that if a USB class doesn't `read()` a successful OUT transfer, the OUT data will sit idle indefinitely, until either it or the host transfers data.  A simple workaround is to enable the SOF interrupts using https://github.com/atsamd-rs/atsamd/pull/456 so that `poll()` gets called within the next millisecond.

### DIFF
--- a/hal/src/thumbv6m/usb/bus.rs
+++ b/hal/src/thumbv6m/usb/bus.rs
@@ -820,18 +820,23 @@ impl Inner {
         let mut ep_in_complete = 0;
         let mut ep_setup = 0;
 
+        let intbits = self.usb().epintsmry.read().bits();
+
         for ep in 0..8u16 {
             let mask = 1 << ep;
 
             let idx = ep as usize;
 
-            if let Ok(bank1) = self.bank1(EndpointAddress::from_parts(idx, UsbDirection::In)) {
-                if bank1.is_transfer_complete() {
-                    bank1.clear_transfer_complete();
-                    ep_in_complete |= mask;
+            if (intbits & mask) != 0 {
+                if let Ok(bank1) = self.bank1(EndpointAddress::from_parts(idx, UsbDirection::In)) {
+                    if bank1.is_transfer_complete() {
+                        bank1.clear_transfer_complete();
+                        ep_in_complete |= mask;
+                    }
                 }
             }
 
+            // Can't test intbits, because bk0rdy doesn't interrupt
             if let Ok(bank0) = self.bank0(EndpointAddress::from_parts(idx, UsbDirection::Out)) {
                 if bank0.received_setup_interrupt() {
                     ep_setup |= mask;
--- a/hal/src/thumbv6m/usb/bus.rs
+++ b/hal/src/thumbv6m/usb/bus.rs
@@ -394,13 +394,6 @@ impl<'a> Bank<'a, OutBank> {
             .write(|w| w.trcpt0().set_bit().trfail0().set_bit());
     }
 
-    /// Checks if data has been received. Returns true for failed transfers
-    /// as well as successful transfers.
-    #[inline]
-    fn is_transfer_complete(&self) -> bool {
-        self.epintflag(self.index()).read().trcpt0().bit()
-    }
-
     /// Returns true if a Received Setup interrupt has occurred.
     /// This indicates that the read buffer holds a SETUP packet.
     #[inline]
@@ -750,6 +743,7 @@ impl Inner {
     }
 
     fn suspend(&self) {}
+
     fn resume(&self) {}
 
     fn alloc_ep(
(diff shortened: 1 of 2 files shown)


ianrrees:
Just noticed this might've broken isochronous transfers; I've got some other local changes that might also be where the problem comes from, but need to do some more digging.

hal/src/thumbv6m/usb/bus.rs
```diff
@@ -995,6 +997,7 @@ impl Inner {
                 }
                 Err(err) => {
                     dbgprint!("UsbBus::read from ep {:?} -> {:?}\n", ep, err);
+                    drop(bank); // bank holds a reference to self.desc, which print_epstatus() needs
```
ianrrees, line 1000: > Just noticed this might've broken isochronous transfers; I've got some other local changes that might also be where the problem comes from, but need to do some more digging.

This was the problem - previously this PR just deleted the `drop(bank);` above.

ianrrees:
I've made a few changes from the original version of this PR, the main one is that the transfer complete and transfer failed flags are cleared in `poll()`.  usb-device says that the "[OUT] event should continue to be reported until the packet is read" - I think we're still doing that because `poll()` will continue to report the event, and will be called regularly because endpoint's transfer failed interrupt will now fire when the host retries.

hal/src/thumbv6m/usb/bus.rs
```diff
@@ -889,40 +823,48 @@ impl Inner {
...
+                    ep_out |= mask;
+                }
+
+                // Clear the transfer complete and transfer failed interrupt flags
+                // so that execution leaves the USB interrupt until the host makes
+                // another transaction.  The transfer failed flag may have been set
+                // if an OUT transaction wasn't read() from the endpoint by the
+                // Class; the hardware will have NAKed (unless the endpoint is
+                // isochronous) and the host may retry.
+                bank0.clear_transfer_complete();
```
ianrrees, line 867: This should happen in the block where ep_out gets set, otherwise there's a timing hole.  I'm chasing down a subtle bug in the OUT endpoint handling, once I've resolved it I'll update this PR.

hal/src/thumbv7em/usb/bus.rs
```diff
@@ -754,27 +754,32 @@ impl Inner {
...
-                    // Continuing (and hence not setting masks to indicate complete
-                    // OUT transfers) is necessary for operation to proceed beyond
-                    // the device-address + descriptor stage. The authors suspect a
-                    // deadlock caused by waiting on a write when handling a read
-                    // somewhere in an underlying class or control crate, but we
-                    // can't be sure. Either way, if a write has finished, we only
-                    // set the flag for a completed write on that endpoint index.
-                    // Future polls will handle the reads.
-                    continue;
+            if (intbits & mask) != 0 {
```
ianrrees, line 764: This is a surprisingly significant optimisation - I've set up a very basic test with a SAMD21 firmware that has 3 USB classes, and at idle it was spending about 3.8% of time in the USB ISR, but with this it's down to 2.6%.  I think it might be worth adding a field to `Inner` that tracks whether any OUT endpoints are ready, then reinstating some of the short-circuits that this PR currently removes.  Will do some tests soon.

hal/src/thumbv7em/usb/bus.rs
```diff
@@ -800,16 +789,33 @@ impl Inner {
                     //  instead it is cleared in the read handler.
```
TDHolmes, line 789: Does this comment block contradict the core change of this PR?

hal/src/thumbv6m/usb/mod.rs
```diff
@@ -11,10 +11,13 @@ mod devicedesc;
 use self::devicedesc::Descriptors;
 
 /// Emit SOF at 1Khz on this pin when configured as function G
+#[allow(deprecated)]
```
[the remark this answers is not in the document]
ianrrees: Good catch, I've removed that change from this PR.

hal/src/thumbv7em/usb/bus.rs
ianrrees: Interesting question!  I think you're correct that it does in spirit, so I'll reword it.

The setup bit is only relevant for control endpoints, and I believe these are nearly-always endpoint 0 in the real world - usb-device currently only supports control pipes on EP0 anyway.  The [usb-device poll()](https://github.com/mvirkkunen/usb-device/blob/b6037b0f98c933ff0a18a864fd34a42f89a6316a/src/device.rs#L206-L207) (which calls this code) immediately handles control transfers on endpoint 0, calling `read()` in the process, so there's practically no difference from immediately clearing the setup interrupt flag.

hal/src/thumbv7em/usb/bus.rs
ianrrees: Sorry, I've not gotten around to doing this yet.  In the spirit of "make it work, then make it fast", I think we're better off leaving optimisations for a different PR.

Issue #426, opened 2021-04-07
API cleanup


glaeqen:
Effort to standardize API across HAL & BSCs. 

- Remove chip-specific modules
  PR #424 :heavy_check_mark: merged
- Flatten HAL by removing redundant `common` module
  PR #425 :heavy_check_mark: merged
- Update the BSPs to re-export the HAL itself, rather than modules within the HAL (open discussion)
  PR #357 :x: under investigation
- Make the `thumb` modules private, to indicate that they are an implementation detail; consider the remaining modules public and let them become the official API
  PR XXX
- Rename imports across the HAL and the BSPs to increase consistency and clarity. (open discussion)
  PR XXX

vccggorski:
As far as I could see, discussion sorta started already in #357. Nonetheless, properly tracked issue seems more appropriate for such thing.

bradleyharden:
Here's another proposed change. I would like to change `target_device` to `pac`. At the very least, maybe we could support both. Thoughts?

Sympatron:
Maybe we could review the `unproven` feature. Last time I looked a lot of basic stuff was "unproven" (I think because the traits were unproven in `embedded-hal`).

Sympatron:
> Here's another proposed change. I would like to change `target_device` to `pac`. At the very least, maybe we could support both. Thoughts?

Seems like this is already part of PR #357.

bradleyharden:
I meant at the HAL level, not the BSP level. I don't think that PR affects the HAL name.

bradleyharden:
@Sympatron, I would expect unproven to disappear with embedded-hal 1.0, but who knows when that will be.

Sympatron:
Thinking about the `unproven` thing more, I came to the conclusion that we probably should not mark whole modules as "unproven" just because they implement some unproven trait. Everything else in these modules is "proven" (to the same degree as most of the hal). So I would vote to mark only the e-h trait impls as `unproven`.

TDHolmes:
#491

TDHolmes:
> Rename imports across the HAL and the BSPs to increase consistency and clarity. (open discussion)
PR XXX

done in #491

ianrrees:
It seems like the work discussed here has all been done, any reason not to close this?

glaeqen:
I think this issue got outdated considering that I stopped following up on the changes that happened in the meantime. We can close it.

PR #460, opened 2021-06-24
Moved unproven feature gates to trait impl blocks

Sympatron:
As suggested in #426, this PR moves the feature gates to the impl blocks. That way, you can still use parts of these modules without `unproven`.

If this is something we want, I would probably add replacement methods in case `unproven` isn't active to make them useful again. And make the e-h trait impls wrappers around those. 

At this state some modules' main functions wouldn't be accessible.

### DIFF
--- a/hal/src/thumbv7em/watchdog.rs
+++ b/hal/src/thumbv7em/watchdog.rs
@@ -30,6 +30,7 @@ impl Watchdog {
     }
 }
 
+#[cfg(feature = "unproven")]
 impl watchdog::Watchdog for Watchdog {
     /// Feeds an existing watchdog to ensure the processor isn't reset.
     /// Sometimes commonly referred to as "kicking" or "refreshing".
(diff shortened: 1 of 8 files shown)


hal/src/thumbv7em/watchdog.rs
```diff
@@ -30,24 +30,60 @@ impl Watchdog {
     }
 }
 
+fn feed(&mut wdt: WDT) {
```
Sympatron, line 33: This could be the way to go, until e-h traits aren't "unproven" anymore. But I am open for suggestions/better options.

bradleyharden:
Honestly, I'm starting to think we should enable `unproven` unconditionally. It's becoming a real pain. Is there *anyone* who wants to turn it off? Given the state of embedded-hal, I doubt it. Once embedded-hal reaches v1.0, we can return it to an optional feature.

Sympatron:
That would another solution.

bradleyharden:
@sajattack, what do you think about making `unproven` unconditional for the time being? Seems more trouble than it's worth

sajattack:
> @sajattack, what do you think about making `unproven` unconditional for the time being? Seems more trouble than it's worth

Yeah go for it.

PR #461, opened 2021-06-28
Update metro_m0 BSP and examples

ianrrees:
The main change is a switch to GPIOv2.  Also made a couple minor fixes to the Feather M0 BSP.

This drops the Metro M0 "rando" example, if we'd prefer to keep it I would be happy to update it too.


sajattack:
Thanks, the Metro M0 is the oldest BSP I think. Fine with dropping rando.

boards/metro_m0/examples/blinky_rtic.rs
```diff
@@ -1,79 +1,70 @@
...
+    use rtic_monotonic::Extensions;
 
-#[app(device = crate::hal::pac, peripherals = true)]
-const APP: () = {
+    #[resources]
     struct Resources {
-        red_led: hal::gpio::Pa17<hal::gpio::Output<hal::gpio::OpenDrain>>,
-        timer: hal::timer::TimerCounter3,
+        red_led:
+            hal::gpio::v2::Pin<hal::gpio::v2::PA17, hal::gpio::v2::Output<hal::gpio::v2::PushPull>>,
```
bradleyharden, line 30: Couldn't you use the `RedLed` alias here? That would be a lot cleaner.

bradleyharden:
@ianrrees, this looks good to me. Just fix that one `Pin` alias and it's ready to go.

boards/metro_m0/examples/blinky_rtic.rs
ianrrees: Indeed!  Does `bsp_pins` expose `DynPinId`s for the un-aliased pins?

bradleyharden: No, it doesn't. But you can get it from the normal `PinId` with `<PA08 as PinId>::DYN`. What do you need them for? If that's a common use, I can add a type alias for that.

ianrrees: (in case it's not clear, I force-pushed a change to this PR to use `RedLed` as above) In my own usage so far, I quite like the new BSP type aliases for the Id+Mode.  It's a minor thing, but a piece of example code (for instance on a blog, not packaged with the HAL) could be entirely independent of the pin mapping if there were an alias that allowed code like

```Rust
static mut DEBUG_PIN: Option<Pin<PA23, Output<PushPull>>> = None;
```
to instead be written like:
```Rust
static mut DEBUG_PIN: Option<PinD4<Output<PushPull>>> = None;
```

bradleyharden: Hmm. I'm not sure I follow how that relates to `DynPinId`.

I was actually planning to do what you describe once we get ride of `gpio::v1`. It would look like this:
```rust
pub type PinA08<M> = Pin<PA08, M>;
```
I haven't implemented it yet, because the might be confusing alongside the existing `v1` types. But maybe that's not true.

bradleyharden: Oh wait, I misunderstood. You want to create a similar alias, but you want to use the BSP pin name instead of the `PinId`. Now I understand. The `D4` confused me. Yeah, I could add that.

ianrrees: Sorry, I'm still not fully up to speed with the Dyn types, or really Rust's type system (lately I'm operating in a perpetual state of distracted burnout).  Yeah, the idea would be for the BSP macro to expose a type too, instead of just making a `Pin` instance available, so that the main application source doesn't need to refer to the PinId directly.

bradleyharden: Let's create a new issue to discuss this. I have a few points to add, and I don't want them to get lost here.

bradleyharden: @ianrrees, see #475 for the promised discussion.

bradleyharden:
@ianrrees, I rebased this and updated the RTIC example for v0.6.alpha5

PR #462, opened 2021-06-28
Get ready for v0.13 release

jbeaurivage:
This PR aims to tie some loose ends before pushing the v0.13 release:

- [x] Add changelog. This should be updated in the future as individual PRs get merged
- [x] Add deprecation notices to modules that exist solely for backwards compatibility
- [x] Add deprecation notice to `gpio::v1`
- [x] Add deprecation notice to `sercom::v1::{spi, uart}`
- [x] Carry out whatever decision is made regarding Tier 1 vs Tier 2 BSPs and add an explanation to the README
- [ ] Update all Tier 1 BSPs to v2 modules:

* ~~`feather_m0`~~
* ~~`metro_m0`~~
* `feather_m4`
* `metro_m4` (@sajattack - #478)
* ~~`samd11_bare`~~
* `pygamer` (@bradleyharden - #455)
* `wio_terminal`

- [x] Tie all Tier 2 BSPs' `atsamd-hal` dependency to crates.io
- [x] Manually bump T2 BSPs (GHA workflow only bumps T1 boards)
- [ ] Bump T1 BSP and PAC versions using GHA workflow (HAL is already bumped)
- [ ] Release to crates.io and tag commit with v0.13.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Unreleased Changes
+
+---
+
+Changelog tracking started at v0.13
--- a/README.md
+++ b/README.md
@@ -89,6 +89,79 @@ $ cargo build --examples
 
 ## How to use a BSP (i.e. getting started writing your own code)
 
+A BSP (**B**oard **S**upport **P**ackage) is a crate that contains definitions specific to a particular board. These generally contain pin definitions, sometimes helper functions to setup certain peripherals, as well as examples to quickly get up and running with your board. BSPs are separated into 2 tiers:
+
+* Tier 1 boards are guaranteed to be up to date with the latest version of `atsamd-hal`
+
+* Tier 2 boards are tied to a specific version of `atsamd-hal`. They are not guaranteed to be updated when a new version of `atsamd-hal` gets released. 
+
+### Tier 1 BSPs
+
+* `feather_m0`
+
+* `feather_m4`
+
+* `metro_m0`
+
+* `metro_m4`
+
+* `pygamer`
+
+* `samd11_bare`
+
+* `wio_terminal`
+
+
+### Tier 2 BSPs
+
+* `arduino_mkr1000`
+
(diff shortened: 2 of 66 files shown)


sajattack:
> I've changed the `atsamd-hal` dependency to crates.io for all Tier 2 BSPs. Now obviously these will all fail CI until a 0.13 version is actually published to crates.io

I typically update the version numbers and publish at the same time, no biggie though.

jbeaurivage:
> I typically update the version numbers and publish at the same time, no biggie though.

That's probably a good idea. I think the PAC versions should also get bumped since we've had SVD patches since 0.12

sajattack:
> > I typically update the version numbers and publish at the same time, no biggie though.
> 
> That's probably a good idea. I think the PAC versions should also get bumped since we've had SVD patches since 0.12

yep

jbeaurivage:
@sajattack, @twitchyliquid64: @bradleyharden brought this to my attention: Now that Tier 2 BSPs will be tied to the crates.io version of `atsamd-hal`, it seems exceedingly wasteful to run CI checks to these boards *all the time*. It would instead make more sense to only check that the Tier 2 BSP+examples compile when they are actually changed. Is this something with which can get GHA to play nice?

twitchyliquid64:
As an open-source project doing actual CI we have unlimited quota here, so we don't have to be cautious with our use of CI.

That said, if you want to do this, we can do this pretty easily as long as `crates.json` captures the tier status of each BSP.

jbeaurivage:
As discussed on Matrix, we could pull the trigger on a release even though not all T1 boards have been updated yet - We would need to upgrade them before the v0.14 release.

bradleyharden:
This is good to go, I think. If anyone wants to jump in and handle the last steps, go for it. Otherwise I'll try to figure it out soon

TDHolmes:
[`feather_m4` update](https://github.com/atsamd-rs/atsamd/pull/487)

PR #468, opened 2021-07-05
Minor `gpio` updates

### LINKED ISSUE #464: gpio::v1: Faulty initial state of pins returned from `PORT::split()`
By design, all GPIO pins are in a `FloatingDisabled` (aka `Reset`) state. Because of backward compatibility in `gpio::v1`, initial state is incorrectly assumed as `FloatingInput`. `v2::Pins::into_mode` based routines make an additional check if mode is in fact changed, otherwise it is a noop. `v1::Pins` are using `v2::Pins` internally. Because of misattributed initial mode, it is impossible to change a mode of a pin to `FloatingInput` as it is incorrectly assumed to be `FloatingInput` already.

Proposed solution: Create a similar `into_mode` mechanism inside of a `gpio::v1` module that ignores a check. It will make a behaviour a compatibility shim at least a little bit more sane.

Tagging @bradleyharden

bradleyharden:
### Remove some unsafe and improve documentation for gpio::v2

The `RegisterInterface` trait in `gpio::v2` used an approach that
produced an unnecessary amount of `unsafe` code. Because each register
in a PAC struct is wrapped in an `UnsafeCell`, it is safe to mutate them
through shared references. The previous approach used a combination of
raw pointers and unique references. However, raw pointers are
unnecessary and unique references offer no benefits, yet they carry the
risk of UB if not implemented correctly.

Update the `RegisterInterface` trait to use shared references
throughout. Also improve the `pin` module documentation slightly.

### Fix regression in gpio::v1 compatibility

At reset, the `gpio::v1` module gives `Pin`s to users in `FloatingInput`
mode, but this is not correct, because the pins are actually in
`FloatingDisabled` mode. This problem was never discovered, because
users always called a function to change the pin mode before using a
given pin, even when that function was `.into_floating_input()`.

In a recent change to `gpio::v2`, the `into_mode` function was altered
to only update the register settings when actually changing modes. The
goal of this change was to prevent spurious writes to the GPIO
registers.

Combined with the incorrect behavior of `gpio::v1`, this change created
a situation where it was impossible to use a pin as a floating input
without first converting it to some other mode and back. Pins would be
labelled as `FloatingInput`, even though they were actually
`FloatingDisabled`. But calls to `.into_floating_input()` would have no
effect, because there was no change in `PinMode`.

Introduce an `into_mode` method for `v1::Pin`s that mirrors the design
of the corresponding method of `v2::Pin`, but in this case change the
register values unconditionally. This does not fix the fundamental flaw
in `gpio::v1`, but it does return it to the pre-existing behavior.

Closes #464

### DIFF
--- a/hal/src/gpio/v2/reg.rs
+++ b/hal/src/gpio/v2/reg.rs
@@ -1,13 +1,12 @@
 #[cfg(any(feature = "samd11", feature = "samd21"))]
 use crate::target_device::port::{
-    pincfg0_::R as PINCFG_R, CTRL, DIR, DIRCLR, DIRSET, DIRTGL, IN, OUT, OUTCLR, OUTSET, OUTTGL,
-    PINCFG0_ as PINCFG, PMUX0_ as PMUX, WRCONFIG,
+    CTRL, DIR, DIRCLR, DIRSET, DIRTGL, IN, OUT, OUTCLR, OUTSET, OUTTGL, PINCFG0_ as PINCFG,
+    PMUX0_ as PMUX, WRCONFIG,
 };
 
 #[cfg(feature = "min-samd51g")]
 use crate::target_device::port::group::{
-    pincfg::R as PINCFG_R, CTRL, DIR, DIRCLR, DIRSET, DIRTGL, IN, OUT, OUTCLR, OUTSET, OUTTGL,
-    PINCFG, PMUX, WRCONFIG,
+    CTRL, DIR, DIRCLR, DIRSET, DIRTGL, IN, OUT, OUTCLR, OUTSET, OUTTGL, PINCFG, PMUX, WRCONFIG,
 };
 
 use crate::target_device::PORT;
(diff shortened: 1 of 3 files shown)


hal/src/gpio/v2/reg.rs
```diff
@@ -262,20 +260,8 @@ pub(super) unsafe trait RegisterInterface {
...
-    fn pincfg_mut(&mut self) -> &mut PINCFG {
-        let group = self.group_mut();
-        // A mutable reference is safe here, because it points to the PINCFG
-        // register of the corresponding `PinId`. Each `Pin` is a singleton, and
-        // we have exclusive control of it through `&mut self`, so the reference
-        // is guaranteed to be unique
-        unsafe { &mut (*group).pincfg[self.id().num as usize] }
+    fn pincfg(&self) -> &PINCFG {
+        &self.group().pincfg[self.id().num as usize]
     }
```
jbeaurivage, line 265: `pincfg_mut` was indeed unnecessary, since interior mutability is erased at the method level.

jbeaurivage:
Overall I'm happy with this. Glad to see potential UB and regressions going away. @bradleyharden, if you didn't have any additional changes to submit, I'd be ready to merge.

bradleyharden:
Nope, I think this is good. Thanks

bradleyharden:
Could you include the contents of my first post in the commit message?

jbeaurivage:
No problem. Seems like CI is stuck though. Maybe a force push would give it a kick?

sajattack:
I gave it a kick here https://github.com/atsamd-rs/atsamd/actions/runs/1075234166

bradleyharden:
Same problem. What's going on?

jbeaurivage:
Aha. Seems like GHA finally woke up.

PR #469, opened 2021-07-06
Compile errors for some invalid feature combinations

ianrrees:
Following https://github.com/atsamd-rs/atsamd/pull/394 , the HAL will compile when there is no feature flag to specify a particular target chip.  However, when a crate depends on parts of the HAL that require hardware support, but a target chip isn't specified, a compile will often result in a big pile of errors that don't usually indicate the fundamental problem.

This issue is particularly noticeable in the context of crates that depend on the HAL, but not on a specific device.  For example, [this crate](https://github.com/ianrrees/atsamd-usbd-uart) provides a USB CDC device using a SERCOM peripheral, and is meant to be added as a dependency to a main firmware.  If this crate is built on its own, currently the compiler will generate errors about missing modules, types, and traits relating to UARTs; but the actual problem is that no specific target chip is enabled in the HAL.  That crate cannot know the specific target chip, it would usually be specified in the BSP and resolved by [Feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification).

This PR catches an instance of this type of problem, but there's room for discussion around the more general solution to cover cases where the external crate doesn't depend on a feature of the HAL like USB.


ianrrees:
Following some chat discussion, this has been extended to error out of compiles where no device feature is enabled.  @henrikssn and @twitchyliquid64 , it would be good to get your input since this might be counter to the goals of #394 .

bradleyharden:
What pieces of the HAL can be used without specifying a chip? My argument was this. If the code doesn't interact with the hardware, then why is it in a hardware abstraction layer crate? Maybe I'm missing something though.

ianrrees:
Just a note about the failing check - I think if we add `device` to the features, it will work as before.

bradleyharden:
@ianrrees, since you added a compiler error in the absence of the `device` feature, do you also need the error for the `usb` feature? Doesn't the former ensure a device is specified?

I'm inclined to say that you **must** specify a chip when compiling the HAL. The HAL should be for *hardware abstraction*. If there are pieces that are independent of hardware, then we should split them out into separate crates. I think the biggest example of that right now is the `time` module. But we should be aiming to replace that with `embedded-time`. I'm not sure if there is anything else.

bradleyharden:
FYI, I think you can add `--features=device` [here](https://github.com/atsamd-rs/atsamd/blob/3988b5950751f28aa9cde4b8b796c5139e183315/.github/workflows/build-bsp.yml#L72) to get rid of the error.

bradleyharden:
After looking more at #394, I think this change does run counter to its goals. But I don't particularly agree with those goals, given my arguments above. I'm open to counter arguments though.

Sympatron:
I don't agree with those goals either. Not specifying a device kind of pointless IMO and even if you wanted to do it it's still easily possible.

For 99% of users it makes a lot more sense to fail the build with a helpful error message if you forgot to specify a specific device.

henrikssn:
> @ianrrees, since you added a compiler error in the absence of the `device` feature, do you also need the error for the `usb` feature? Doesn't the former ensure a device is specified?
> 
> I'm inclined to say that you **must** specify a chip when compiling the HAL. The HAL should be for _hardware abstraction_. If there are pieces that are independent of hardware, then we should split them out into separate crates. I think the biggest example of that right now is the `time` module. But we should be aiming to replace that with `embedded-time`. I'm not sure if there is anything else.

I completely agree with what's stated above. The problem arises when the HAL not only implements an abstraction (which is naturally device independent), but also provides that abstraction (which is typically device independent), as libraries which needs the abstraction but not the implementation then unnecessarily need to be tied to a specific device.

The ideal solution is to migrate out everything which is device independent into its own crate, either an existing such as embedded-hal / embedded-time or possibly a new crate for traits shared across atsamd devices. This is a lot of work, so #394 solved this using the path of least resistance by allowing the HAL to compile without features to expose the abstractions.

ianrrees:
> @ianrrees, since you added a compiler error in the absence of the `device` feature, do you also need the error for the `usb` feature? Doesn't the former ensure a device is specified?
> 
The USB error is a bit more subtle, and I think it's still needed.  The issue is that the USB feature could be enabled, along with a device that either doesn't have USB hardware, or lacks HAL support for USB (SAMD11).

Sympatron:
> as libraries which needs the abstraction but not the implementation then unnecessarily need to be tied to a specific device.

Are there libraries depending on atsamd-hal? On crates.io are only BSPs that depend on it, which use one specific device.

henrikssn:
> > as libraries which needs the abstraction but not the implementation then unnecessarily need to be tied to a specific device.
> 
> ***Are there libraries depending on atsamd-hal?*** On crates.io are only BSPs that depend on it, which use one specific device.
> Even if there are, it's still not a big deal, because it still compiles just fine if you activate the `device` feature manually instead of specifying a specific device.

The answer is yes :) Specifying `device` feature for these explicitly sounds like a good workaround to me.

hal/src/lib.rs
```diff
@@ -6,6 +6,9 @@ pub use paste;
 
 pub mod typelevel;
 
+#[cfg(not(feature = "device"))]
+compile_error!("The HAL requires a target device selection via a Cargo feature");
```
henrikssn, line 10: Please add to the error message: "or if this is a library, disable target device selection by specifying the `device` feature"

hal/src/lib.rs
```diff
@@ -98,6 +101,10 @@ pub mod timer_traits;
 #[cfg(all(feature = "unproven", feature = "dma"))]
 pub mod dmac;
 
+// The SAMD11 hardware supports USB too, but the SAMD11 HAL currently doesn't
+#[cfg(all(feature = "usb", not(feature = "samd21"), not(feature = "min-samd51g")))]
+compile_error!("The 'usb' feature is enabled, but not a specific chip that supports USB");
```
henrikssn, line 106: Doesnt this do the wrong thing (crash) when 
 `device` and `usb` features are enabled?

Sympatron: This would not compile with only `usb` and `device`, but I doubt there is any device-independent usb-specific code...

Sympatron:
Alternatively we could a a `no-device` or `library` feature that can be specified to make it compile without a device. This would probably be a bit more self-explanatory and less hacky.

ianrrees:
> Alternatively we could a a `no-device` or `library` feature that can be specified to make it compile without a device. This would probably be a bit more self-explanatory and less hacky.

I came to the same conclusion - we need to add a new feature which would only be used in:
```Rust
#[cfg(all(not(feature = "library"), not(feature = "device")))]
compile_error!("The HAL requires a target device selection via a Cargo feature, or if this is a library, disable target device selection by specifying the `library` feature");
```

hal/src/lib.rs
ianrrees: I think we'll need to modify the condition to consider `library`:
```Rust
#[cfg(all(feature = "usb", not(feature = "samd21"), not(feature = "min-samd51g"), not(feature = "library")))]
```

ianrrees:
I've rebased on to current master and made another attempt.  Any suggestions to improve wording would be appreciated!

jbeaurivage:
@ianrrees is this ready? We could try to squeeze it into the v0.13 release.

ianrrees:
Yep, I think so

PR #470, opened 2021-07-14
Support enabling clock mode RTC without resetting

theatrus:
If you're trying to use the HAL RTC and a VBat backup power source, it
makes no sense to fully reset the RTC on every initialization.

This patch adds new methods (_noreset) which do not reset the RTC. The
existing methods were kept for API compatibility.


bradleyharden:
@henrikssn, I don't have any experience with RTC. Any chance you could look at this?

bradleyharden:
@theatrus, I haven't used the RTC before, but why do you need to call these functions more than once? Are you switching back and forth between count mode and clock mode? Beyond resetting the peripheral, it's also enabling the APB clock on each call, which is also useless if it has already been enabled.

If you truly need this functionality, it seems like a better solution might be to rethink the API more generally, to piece out which functionality should be grouped into methods.

@vccggorski, I can't remember where we landed on the RTC. Were we planning to include it in the update to the `clock` module? I think so. I need to catch up on it.

AfoHT:
The RTC implementation would regardless need some touchup to play nice with `clocking::v2`. We also have an older in-house implementation written before the atsamd-hal RTC parts were added. Ideally a 2-way merge of the best from both would result in the third `clocking::v2` compatible version.

But I believe that should come after clocking is mainlined, just to not have even more mid-air dependencies than we already have :smile:

theatrus:
> @theatrus, I haven't used the RTC before, but why do you need to call these functions more than once? Are you switching back and forth between count mode and clock mode? Beyond resetting the peripheral, it's also enabling the APB clock on each call, which is also useless if it has already been enabled.
> 
> If you truly need this functionality, it seems like a better solution might be to rethink the API more generally, to piece out which functionality should be grouped into methods.
> 
> @vccggorski, I can't remember where we landed on the RTC. Were we planning to include it in the update to the `clock` module? I think so. I need to catch up on it.

The functions are called at POR and when the device performs a switch from backup domain power. It’s never called twice out of init code. 

In backup domain the RTC registers remain, as does the 32k OSC, and some RAM ( device dependent), but all other peripherals are reset state, so it doesn’t hurt to reset the clock domains.

PR #472, opened 2021-07-20
Fix typo in SPI interrupt register field

ianrrees:
Should be fairly self-explanatory ;)


jbeaurivage:
Awesome! 

Looks like CI is failing because the BSP examples haven't been updated to the new RTIC alpha.

bradleyharden:
@ianrrees, could you fix thumbv7em as well? And could you hardcode the RTIC version to alpha4? That should fix the tests. We can upgrade them later.

bradleyharden:
@jacobrosenthal, this is probably the problem with #455

ianrrees:
I've updated this PR for v7, but per matrix chat will make a new PR once we've figured out a fix for the RTIC example failures.

PR #473, opened 2021-07-21
Fix CI for RTIC 0.6.0-alpha.5

jbeaurivage:
This is a quick fix to get CI working again. There is currently exacly 1 BSP example using RTIC `0.6.0-alpha.5`: `feather_m0/blinky_rtic`.

### DIFF
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -52,7 +52,7 @@ version = "0.3"
 optional = true
 
 [dependencies.rtic-monotonic]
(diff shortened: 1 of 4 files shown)


LongLiveCHIEF:
I'm attempting to update my github.com/LongLiveCHIEF/led-backlighting-controller from rtic alpha.2 to alpha.5, and will try to provide some examples when I get that working.

At the moment having some issues with the monotonic that weren't happening in 0.6.0-alpha.2.

It would _really_ help if patches were published to crates.io so i'm not constantly trying to merge things into the correct state on my local machine.

PR #474, opened 2021-07-24
boards/qt_py_m0: Add support for Adafruit QT Py

gkelly:
Adds support for the QT Py board, using the `bsp_pins!` approach.


bradleyharden:
@gkelly, thanks for this. Can I assume you've tested the neopixel example? Any chance you could add a USB example as well? I think there should be one you can copy from pygamer or feather_m0.

boards/qt_py_m0/examples/usb_echo.rs
```diff
@@ -0,0 +1,103 @@
...
+        NVIC::unmask(interrupt::USB);
+    }
+
+    loop {
+        block!(timer.wait()).ok();
+        unsafe {
+            USB_SERIAL.as_mut().map(|serial| {
+                serial.write("tick!\r\n".as_bytes()).ok();
+            });
+        }
```
bradleyharden, line 77: I haven't actually looked at USB examples before now. Did you write this? Or is this copied from another example?

This seems like UB waiting to happen, because you access `USB_SERIAL` with an `&mut` reference from both `main` and from `poll_usb`. If this were in RTIC, it would require a `Mutex`.

If all the examples are like this, then we can merge it for now, but we should try to get rid of this, I think.

bradleyharden: Looks like the metro_m0 and feather_m0 examples are different and don't have this problem.

If you want to do something like this, I would **highly** recommend using RTIC. `static mut` variables are *notoriously* difficult to implement soundly. Why try to do it yourself when you can let a well-reviewed code-gen tool do it for you? It's also much easier to understand and gets rid of a lot of the boilerplate you have here.

gkelly: They're not like this, I added this for something for the main loop to do. It's most definitely UB, I was going to ask if anyone knew a nice ergonomic way of doing what I'm doing here. I was considering just wrapping this `free`. I can just drop this have throw a `wfi` in here instead though if you'd prefer.

gkelly:
> @gkelly, thanks for this. Can I assume you've tested the neopixel example? Any chance you could add a USB example as well? I think there should be one you can copy from pygamer or feather_m0.

Yep, the neopixel example runs on the hardware:

small

I added the USB echo example from `feather_m0` with some small changes.

gkelly:
Drop the main-loop serial write in favor of just a `wfi`.

TDHolmes:
Noticed this was never published to crates.io and it seems the example doesn't build on TOT. fixup in #563

PR #478, opened 2021-08-22
Update Metro M4 BSP and examples for v2/tier1

sajattack:


jbeaurivage:
Should we merge this even though the SAMD51 still seems to be buggy?

boards/metro_m4/examples/adc.rs
```diff
@@ -27,35 +32,14 @@ fn main() -> ! {
...
-        .expect("Could not configure sercom3 clock");
-
-    let mut uart = UART3::new(
-        &uart_clk,
-        9600.hz(),
-        peripherals.SERCOM3,
-        &mut peripherals.MCLK,
-        (rx, tx),
-    );
+    let mut a0 = pins.a0.into_alternate::<B>();
```
TDHolmes, line 38: Could use the `aliases` feature for this one:
```rust
let mut a0: bsp::Analog0 = pins.a0.into();
```

and `lib.rs` would have something like:
```rust
    PA02 {
         /// Analog pin 0.  Can act as a true analog output
         /// as it has a DAC (which is not currently supported
         /// by this hal) as well as input.
         name: a0,
        aliases: {
            AlternateB: Analog0
     }
```

boards/metro_m4/examples/clock_out.rs
```diff
@@ -23,11 +29,12 @@ fn main() -> ! {
...
-    let mut pins = hal::Pins::new(peripherals.PORT);
+    let pins = bsp::Pins::new(peripherals.PORT);
 
-    let _gclk2 = clocks
+    // assuming 120MHz main clock, output a 10MHz clock on d7
+    let _ = clocks
         .configure_gclk_divider_and_source(GCLK2, 12, DPLL0, false)
         .unwrap();
-    pins.d13.into_function_m(&mut pins.port);
+    let _ = pins.d7.into_alternate::<M>();
```
TDHolmes, line 38: similar `into()` comment here

boards/metro_m4/examples/pwm.rs
```diff
@@ -25,16 +28,16 @@ fn main() -> ! {
         &mut peripherals.NVMCTRL,
     );
     let mut delay = Delay::new(core.SYST, &mut clocks);
-    let mut pins = hal::Pins::new(peripherals.PORT);
+    let pins = bsp::Pins::new(peripherals.PORT);
 
-    let d12 = pins.d12.into_function_e(&mut pins.port);
+    let d12 = pins.d12.into_alternate::<E>();
```
TDHolmes, line 33: similar `into()` comment here

boards/metro_m4/src/lib.rs
```diff
@@ -1,213 +1,345 @@
...
-        scl.into_pad(port),
-    )
+    let clock = &clocks.sercom5_core(&gclk0).unwrap();
+    let baud = baud.into();
+    let sda = sda.into();
+    let scl = scl.into();
+    I2CMaster5::new(clock, baud, sercom5, mclk, sda, scl)
 }
 
+pub type UartPads = uart::Pads<Sercom3, IoSet1, UartRx, UartTx>;
```
TDHolmes, line 309: since its public, should we doc?

TDHolmes (approved):
looks good overall to me!

sajattack:
@TDHolmes I'm not totally sold on the need for all the different aliases, so I just added the doc comment you recommended. Maybe we can have a discussion about it on matrix at a later time.

PR #480, opened 2021-08-28
ICM - Integrity Check Monitor

vcchtjader:
HAL interface for the ICM, useful for calculating SHA1, SHA224, SHA256 checksums and monitoring of
memory regions

As a demo of functionality the main application and two examples found in the `icm` branch of my [atsamd-demo](https://github.com/vcchtjader/atsamd-demo/tree/icm) repo:
* `example/icm-clockv1` uses the current v1 clocking and not much more to compute SHA sums
* `example/icm-clockv2` uses v2 clocking and UART0 and pretty prints output instead
* `src/main` uses v2 clocking and just like `icm-clockv2` it computes the SHA sums, but then switches over to memory monitoring mode.

All of the examples are inspired by the demo software provided by Atmel/Microchip, see [source][source].
They also have an application note [here][application-note].

[source]: https://github.com/atmelcorp/atmel-software-package/blob/master/examples/crypto_icm/main.c
[application-note]: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-42703-ICM-Usage-on-SAM-S7-E7-V7-Microcontrollers_ApplicationNote_AT12869.pdf

These demos might be useful as part of the examples for some of the boards (thumbv7). I've been testing on a custom board `atsamd51n` so cannot verify on hardware such as the `feather_m4`, but it should be fairly simple to copy if somebody can test/verify.

This should not conflict with anything else in the repo. For now it assumes the user is able to enable the `APB` clock for `ICM`, to not incur any clocking version dependency.

`cargo test --doc --features unproven,same51n-rt -- icm` passes

Looking forward to feedback and suggestions :smile: 

Uart output from running `src/main`:

```
Booted RTIC.
ICM Interrupt!
Region 0: Expected,  Received - SHA1 (Should not match)
mismatch! 0x363e99a9 0x91e54adc
mismatch! 0x6a810647 0x31c6b14e
mismatch! 0x71253eba 0xd029c237
mismatch! 0x6cc25078 0xd75a8ad7
mismatch! 0x9dd8d09c 0xd92b0463
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
Region 1: Expected,  Received - SHA1 (should match)
   Match! 0x363e99a9 0x363e99a9
   Match! 0x6a810647 0x6a810647
   Match! 0x71253eba 0x71253eba
   Match! 0x6cc25078 0x6cc25078
   Match! 0x9dd8d09c 0x9dd8d09c
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
Region 2: Expected,  Received - SHA224 (should match)
   Match! 0x227d0923 0x227d0923
   Match! 0x22d80534 0x22d80534
   Match! 0x77a44286 0x77a44286
   Match! 0xb355a2bd 0xb355a2bd
   Match! 0xe4bcad2a 0xe4bcad2a
   Match! 0xf7b3a0bd 0xf7b3a0bd
   Match! 0xa79d6ce3 0xa79d6ce3
   Match! 0x00000000 0x00000000
Region 3: Expected,  Received - SHA256 (should match)
   Match! 0xbf1678ba 0xbf1678ba
   Match! 0xeacf018f 0xeacf018f
   Match! 0xde404141 0xde404141
   Match! 0x2322ae5d 0x2322ae5d
   Match! 0xa36103b0 0xa36103b0
   Match! 0x9c7a1796 0x9c7a1796
   Match! 0x61ff10b4 0x61ff10b4
   Match! 0xad1500f2 0xad1500f2
Done!
 Switch to region monitoring mode
Manually modify region1
Manually modify region1
Manually modify region2
Manually modify region3
ICM Interrupt!
Region0 digest mismatch!
Region 0: Expected,  Actual
   Match! 0x11111111 0x11111111
   Match! 0x22222222 0x22222222
   Match! 0x33333333 0x33333333
mismatch! 0x44444444 0xdeadbeef
   Match! 0x55555555 0x55555555
   Match! 0x66666666 0x66666666
   Match! 0x77777777 0x77777777
   Match! 0x88888888 0x88888888
   Match! 0x99999999 0x99999999
   Match! 0xaaaaaaaa 0xaaaaaaaa
   Match! 0xbbbbbbbb 0xbbbbbbbb
   Match! 0xcccccccc 0xcccccccc
   Match! 0xdddddddd 0xdddddddd
   Match! 0xeeeeeeee 0xeeeeeeee
   Match! 0xffffffff 0xffffffff
   Match! 0x00000000 0x00000000
Region1 digest mismatch!
Region 1: Expected,  Actual
   Match! 0x80636261 0x80636261
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
mismatch! 0x00000000 0xdeadbeef
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x18000000 0x18000000
Region2 digest mismatch!
Region 2: Expected,  Actual
   Match! 0x80636261 0x80636261
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
mismatch! 0x00000000 0xdeadbeef
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x18000000 0x18000000
Region3 digest mismatch!
Region 3: Expected,  Actual
   Match! 0x80636261 0x80636261
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
mismatch! 0x00000000 0xdeadbeef
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x00000000 0x00000000
   Match! 0x18000000 0x18000000
RDM interrupts 1111
ICM Interrupt!
```


hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1530 @@
...
+
+use crate::pac::icm::*;
+use crate::typelevel::Sealed;
+use core::marker::PhantomData;
+
+/// Reexport the User SHA Algorithm
+pub use crate::icm::cfg::UALGO_A as icm_algorithm;
+
+/// IcmToken represents the peripheral and it encapsulates the hardware register
+pub struct IcmToken {
```
twitchyliquid64, line 363: If `IcmToken` is an internal member of `Icm`, can it be made private?

AfoHT:
> Awesome stuff! I'm not familiar with ICM specifically but from a quick glance your code looks good, my only suggestions would be to try and limit the public methods / structs to just whats needed, i think `IcmToken` can be private?

You are correct, `IcmToken` should be internal to `Icm` and not public: see https://github.com/atsamd-rs/atsamd/pull/480/commits/3d65376a0479968e1d721000f8e600a1094c3cf8  :heavy_check_mark:

@vccggorski also had a look and provided some insights, for example instead of storing state just return an object providing the queryable interface.

Most importantly, since this is a hardware peripheral dealing with raw memory addresses and with the ability to write and reading memory, the location and persistence of these memory addresses used becomes paramount.

In the demo app linked above using RTIC I've changed those to be static, which required some changes.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+use crate::typelevel::Sealed;
+use core::marker::PhantomData;
+
+/// Reexport the User SHA Algorithm
+pub use crate::icm::cfg::UALGO_A as icm_algorithm;
+
+/// IcmToken represents the peripheral and it encapsulates the hardware register
+struct IcmToken {
+    icm: crate::pac::ICM,
+}
```
bradleyharden, line 374: What exactly is the purpose of this struct? It's similar to the `Registers` structs in `v2::spi` and `v2::uart`, but there's a key difference. The `Registers` structs act to *completely* encapsulate interaction with the PAC, which means calls to PAC functions are contained to a single type. But this struct only forwards along references to the particular registers.

I also don't know that `Token` is the right name. Intuitively, I would expect a `Token` type to have no real utility on its own. It can only be exchanged for something that does have utility. I always imagine it like the tokens at a video game arcade. We use a lot of `Token` types in the `clock::v2` module, but there it makes more sense. The PAC does not have any dedicated types for the individual clocks, so we create various `Token`s to represent uninitialized clocks.

In this case, I think the `pac::ICM` type already acts as the token. Users could use the PAC struct directly, or they can exchange it for a `hal::Icm` struct.

If it were me, I would either: 1) remove this type and access the `pac::ICM` struct from the `hal::Icm` type directly, or 2) turn this into a dedicated register interface struct, and remove all PAC register writing from other types.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+
+/// Struct useful for returning the interrupt status
+/// of the ICM. Provides methods for easy parsing of
+/// all the regions or via the `bitmask` argument
+/// narrow it down to the specific set of [`IcmRegionNum`]
+/// of interest.
+pub struct IcmInterrupt {
+    interrupt_vector: u32,
+}
```
bradleyharden, line 456: You might consider `bitflags` or some other crate for this type. I think it would end up being simpler for you and more useful for users. @jbeaurivage and I found it useful for a similar purpose in `sercom::v2`.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+    }
+}
+
+/// Struct useful for returning the interrupt status
+/// of the ICM. Provides methods for easy parsing of
+/// the region specific [`IcmRegionNum`]
+pub struct IcmRegionInterrupt<I: IcmRegionNum> {
+    region: PhantomData<I>,
+    interrupt_vector: u32,
+}
```
bradleyharden, line 509: The type parameter makes it more difficult, but I think this also might be another candidate for a `bitflags` type, or something similar.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+    pub fn get_rhc_int(&self) -> bool {
+        matches!((self.interrupt_vector & 0x0f) as u8 & self.mask(), 1)
+    }
+}
+
+/// IcmRegionToken provides access to region-specific
+/// settings like interrupts and status
+pub struct IcmRegionToken<I: IcmRegionNum> {
+    region: PhantomData<I>,
+}
```
bradleyharden, line 559: I wouldn't call this a `Token` type. I represents the public interface for an ICM region, so I would call it `IcmRegion` directly. I think you use that name further down, for the collection of four regions, but I would rename that to `IcmRegions`, with an `s`.

Separately, you might consider removing the `Icm` prefix from all your types. If users what to disambiguate, they can always use the `icm::` namespace. And removing the `Icm` prefix gives users the option to be more terse when context tells you that `Region` means `icm::Region`. It's up to you though.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+    // Region specifics
+
+    /// Enable region0
+    ///
+    /// Creates an [`IcmRegionToken`] which provides region specific
+    /// settings
+    #[inline]
+    pub fn enable_region0(&mut self) -> IcmRegionToken<Region0> {
+        IcmRegionToken::new()
+    }
```
bradleyharden, line 899: Maybe consider also adding a generic version as well, i.e. `fn enable_region<R: RegionNum>(...)`? It might be useful to let the type be inferred sometimes.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+    const OFFSET: u32;
+}
+
+/// ICM Region 0
+pub enum Region0 {}
+impl Sealed for Region0 {}
+impl IcmRegionNum for Region0 {
+    const NUM: usize = 0;
+    const OFFSET: u32 = 0;
+}
```
bradleyharden, line 1126: You could save some typing by turning this into a `seq!` macro call. We already use it throughout the HAL.

```rust
seq!(for N in 0..=3 {
    paste! {
        #[doc = "ICM Region " N]
        pub enum Region#N {}
        impl Sealed for Region#N {}
        impl IcmRegionNum for Region#N {
            const NUM: usize = N;
            const OFFSET: u32 = 0x10 * N;
        }
    }
});
```

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+impl IcmRegionNum for Region3 {
+    const NUM: usize = 3;
+    const OFFSET: u32 = 0x10 * 3;
+}
+
+/// Functions required by [`IcmRegionDesc`]
+///
+/// Both Main List descriptors and Secondary List descriptors
+pub trait IcmRegionDescT {
+    /// Set the [`Raddr`] start of the region memory region
```
bradleyharden, line 1156: I don't really like adding a `T` suffix to break name collisions between a type and a trait. I think it indicates that something isn't named as clearly as it could be.

In this case, it looks like you have `Main` descriptors and `Secondary` descriptors. Why not change `IcmRegionDesc` to `icm::MainRegionDesc`. I think that would match well with the `icm::SecondaryRegionDesc`.

Also, does the datasheet use Main & Secondary, not Primary & Secondary? If so, that's unfortunate. The pairing of Main & Secondary is a bit odd.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+///
+/// >**Important**
+/// >
+/// >ICM engine will **read** wherever this
+/// >is instantiated in memory, based on the [`Icm::set_dscr_addr()`]
+/// >so the user must ensure that this variable lives long enough or is
+/// >static
+#[repr(C)]
+#[repr(align(64))]
+pub struct IcmRegion {
```
bradleyharden, line 1183: As mentioned above, I would rename this to `icm::Regions`, with an `s`.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+        let region0 = IcmRegionDesc::new_region0();
+        let region1 = IcmRegionDesc::new_region1();
+        let region2 = IcmRegionDesc::new_region2();
+        let region3 = IcmRegionDesc::new_region3();
+        Self {
+            region0,
+            region1,
+            region2,
+            region3,
+        }
```
bradleyharden, line 1205: Why not implement `Default` here?

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+
+    const fn new_region0() -> Self {
+        IcmRegionDesc {
+            num: PhantomData,
+            raddr: Raddr::default(),
+            rcfg: Rcfg::default(),
+            rctrl: Rctrl::default(),
+            rnext: Rnext::default(),
+        }
+    }
```
bradleyharden, line 1249: Why not pull this out into a generic version?

```rust
impl<R: RegionNum> MainRegionDesc<R> {
    const fn new_region() -> Self {
        // ...
    }
}
```

If you really want separate names, you can then use a `seq!` macro to create the four different versions.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+    pub fn region1(&self) -> [u32; 8] {
+        self.region1
+    }
+    pub fn region2(&self) -> [u32; 8] {
+        self.region2
+    }
+    pub fn region3(&self) -> [u32; 8] {
+        self.region3
+    }
```
bradleyharden, line 1399: Why do you need these if the `region` fields are `pub`?

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+    pub const fn default() -> Self {
+        IcmHashArea {
+            region0: [0; 8],
+            region1: [0; 8],
+            region2: [0; 8],
+            region3: [0; 8],
+        }
+    }
```
bradleyharden, line 1407: Again, why not implement `Default`?

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+
+/// Region Start Address Structure
+///
+/// Follows C-structure conventions
+#[derive(Clone, Copy)]
+#[repr(C)]
+pub struct Raddr {
+    pub raddr: u32,
+}
+
```
bradleyharden, line 1418: I think you're matching the datasheet here, but I don't like `Raddr` as a name. It's very terse. Without a lot of context, I wouldn't know what it represents.

That being said, if you rename it here, you probably ought to be more verbose throughout the module, e.g. in function names. Personally, I would probably prefer `RegionAddress`, `get_region_address`, etc. But I can see counter arguments.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+    pub fn set_raddr<T>(&mut self, raddr: *const T) {
+        self.raddr = raddr as u32;
+    }
+    const fn default() -> Self {
+        // Unsure what a good Raddr default should be,
+        // this is at least easily spotted when debugging...
+        Raddr {
+            raddr: 0xDEADBEEF_u32,
+        }
+    }
```
bradleyharden, line 1429: Again with `Default`

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+#[repr(C)]
+pub struct Rcfg {
+    rcfg: u32,
+}
+
+impl Rcfg {
+    const fn default() -> Self {
+        Rcfg { rcfg: 0x3F0 }
+    }
+}
```
bradleyharden, line 1445: The comments from `Raddr` apply here and to all the types below. The names are quite terse and cryptic unless you're already familiar with the peripheral.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1612 @@
...
+        self.rcfg = (self.rcfg & !0x01) | (value as u32);
+    }
+    /// Wrap Command
+    ///
+    /// * true: next region descriptor address loaded is `DSCR`.
+    /// * false: the next region descriptor address is `current + 0x10`.
+    #[inline]
+    pub fn set_wrap(&mut self, value: bool) {
+        self.rcfg = (self.rcfg & !(0x01 << 1)) | ((value as u32) << 1);
+    }
```
bradleyharden, line 1463: This looks like a lot of manual work to maintain a set of bit fields. I think you should consider using `modular_bitfields` or a similar crate. I know @jbeaurivage made good use of it in the DMA module for his descriptors.

hal/src/thumbv7em/icm.rs
bradleyharden: Also, why `new_region` instead of just `new`? Doesn't the type name already imply it's a region?

hal/src/thumbv7em/icm.rs
AfoHT: I agree it was somewhere in-between, not ideal. Opted for alternative 1), since `Region` also needs access to `ICM` and it isn't easy (at least to my knowledge) to refer back to this single copy it would be a bit clunky (imho) passing around this dedicated register interface.

tl;dr: `pac::ICM` is the "token", no need for extra "token"

hal/src/thumbv7em/icm.rs
AfoHT: Great idea, adopted!

hal/src/thumbv7em/icm.rs
AfoHT: Not as straight-forward, but combining the previous style and bitflags became much nicer than it was

hal/src/thumbv7em/icm.rs
AfoHT: Adopted
* `IcmRegionToken` -> `Region`,
*  `IcmRegion` -> `Regions`

hal/src/thumbv7em/icm.rs
AfoHT: Added :+1:

hal/src/thumbv7em/icm.rs
AfoHT: Very kind of you to provide a copy-paste solution, just added some `#[allow(clippy::...)]` to make Clippy happy.

hal/src/thumbv7em/icm.rs
AfoHT: Datasheet indeed uses Main and Secondary, example page 628 figure 26-2 and 26-3.

It became nicer with:

* `MainRegionDesc`
* `SecondaryRegionDesc`

which also frees up `RegionDesc` for the trait

hal/src/thumbv7em/icm.rs
AfoHT: Adopted

hal/src/thumbv7em/icm.rs
AfoHT: Previously it was using standard `impl Default for ...`, but since I want to have the `HashArea` and `Regions` `'static` I ran into error [E0015][E0015] which isn't surprising, trying to make all the relevant parts `const` didn't play well with `impl Default`, resulting in [E0379][E0379]. Maybe one day, but it seems not yet available.

This is what prompted these 'odd' workarounds, including the less than pleasant naming `new_region*`, for some reason all `new()` were picked up, prompting "multiple applicable items in scope", even though `impl MainRegionDesc<Region*>` was repeated for each number, explicitly.  Can't be generic due to `const fn` -.-

Opted for `seq!` macro for now

[E0015]: https://doc.rust-lang.org/error-index.html#E0015
[E0379]: https://doc.rust-lang.org/error-index.html#E0379

hal/src/thumbv7em/icm.rs
AfoHT: Good point, removed these methods

hal/src/thumbv7em/icm.rs
AfoHT: Adopted `RegionAddress`, similarly for `Rctrl` etc.

hal/src/thumbv7em/icm.rs
AfoHT: Once again, great idea. Simplified it quite a bit.

hal/src/thumbv7em/icm.rs
bradleyharden: 📎😃

hal/src/thumbv7em/icm.rs
bradleyharden: Hmm. Yeah, I guess that sounds like the best option.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1640 @@
...
+    #[inline]
+    pub fn reset_region_configuration_to_default(&mut self) {
+        self.0 = 0x3F0_u32;
+    }
+}
+
+/// Processing Delay
+///
+/// For a given SHA algorithm, the runtime period has two possible lengths
+pub enum Procdly {
```
bradleyharden, line 1587: This name feels a bit out of place now. If `ProcessingDelay` is too long, I would at least capitalize the `D` to `ProcDly`.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1640 @@
...
+    #[inline]
+    fn uasr(&self) -> &UASR {
+        &self.icm().uasr
+    }
+
+    // User interface for ICM
+
+    /// Enable the ICM peripheral
+    #[inline]
+    pub fn enable(&mut self) {
```
bradleyharden, line 937: I don't quite understand this peripheral, but it seems like there are some things you can/can't do before/after enabling it. Would it be worth adding type safety, like we have for `v2::spi`, `v2::uart` and `clock::v2`? I'm not sure if it would make sense or how difficult it would be.

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1640 @@
...
+    /// Disable Region Hash Completed interrupt enable
+    #[inline]
+    pub fn disable_rhc_int(&self) {
+        self.idr().write(|w| unsafe { w.rhc().bits(self.mask()) });
+    }
+
+    /// Get Region Status Updated interrupt enable mask
+    #[inline]
+    pub fn get_rsu_int_mask(&self) -> bool {
+        matches!(self.imr().read().rsu().bits() & self.mask(), 1)
```
bradleyharden, line 753: You must have 30 copies of this pattern throughout the module. Would it be worth adding a function or macro for it?

hal/src/thumbv7em/icm.rs
```diff
@@ -0,0 +1,1640 @@
...
+//! ];
+//!
+//! static MESSAGE_SHA224_RES: [u32; 8] = [
+//!     0x227d0923, 0x22d80534, 0x77a44286, 0xb355a2bd, 0xe4bcad2a, 0xf7b3a0bd, 0xa79d6ce3, 0x00000000,
+//! ];
+//! static MESSAGE_SHA256_RES: [u32; 8] = [
+//!     0xbf1678ba, 0xeacf018f, 0xde404141, 0x2322ae5d, 0xa36103b0, 0x9c7a1796, 0x61ff10b4, 0xad1500f2,
+//! ];
+//! static mut HASH: HashArea = HashArea::default();
+//! static mut ICM_REGION_DESC: Regions = Regions::default();
```
bradleyharden, line 136: Would it be worth using [`cortex_m::singleton`](https://docs.rs/cortex-m/0.7.3/cortex_m/macro.singleton.html) for these `static mut`s? I think it's a bit safer. I've heard `static mut` values are notoriously UB-prone. If nothing else, it might set a good example for people referring to your documentation.

vcchtjader: Sounds like a good idea. I weren't able to use it in the RTIC demo-app, but I added it to documentation for all other use-cases, see https://github.com/atsamd-rs/atsamd/pull/480/commits/2675d6211566c7ef8b3e5c0dd21e221d5d06a014

hal/src/thumbv7em/icm.rs
vcchtjader: Good point, made it cleaner with some macros in https://github.com/atsamd-rs/atsamd/pull/480/commits/912991108dae800bc11dd5aadbff641a6e075ffc

hal/src/thumbv7em/icm.rs
vcchtjader: Adopted :+1: `Procdly` -> `ProcessingDelay`

hal/src/thumbv7em/icm.rs
vcchtjader: I did have a go at getting some basic type safety, but it became rather cumbersome to use as a RTIC resource, which most likely will be my use case. It is quite possible I got it all wrong, but I get stuck on either requiring dynamic Trait objects, which the compiler isn't happy about since now Icm with the `Is` trait refer to `Self`, or just being unable to lock the RTIC resource and modify it in place since I cannot move out of a reference.

My experimentation can be found here: [HAL icm-typestate][haltypestate] and [demo icm-typestate][demotypestate]

[haltypestate]: https://github.com/vcchtjader/atsamd/tree/icm-typestate
[demotypestate]: https://github.com/vcchtjader/atsamd-demo/tree/icm-typestate

For the bigger picture, looking at the [atmel-software-demo][atmeldemo] they use it as a proper "enable" after configuring the ICM. When ICM has completed its work, they turn it off, then software reset (`swrst()`) and then configure ICM anew before enabling.

Turning it off explicitly is interesting since `swrst()` clears the `enable` bit, maybe they just wanted to be extra clear.

Some observations after testing:

If one set up the ICM for memory monitoring and set the `wrap` bit, causing it to loop back to the start through memory and monitor it continuously then it is easy to observe the `enable` bit, since it is continuously operating.

On the other hand, if there is no wrap bit set, then the `enable` bit acts as a "run once" push-button. Observable not through the `enable` bit (my timing and GDB overhead and all that makes it hard I imagine), but rather through the continuous stream of interrupts if you set the `enable` bit in the interrupt handler, causing yet another interrupt, creating an infinite loop.

With this rather loose usage of the enable bit and the general complexity of the peripheral I think it might be quite tricky to achieve something fully safe while also having the full flexibility if taking the typestate path, especially since it is not one of those "everybody will need this" peripherals. However, if that becomes the case we could always revisit this :smile: 

Thoughts?

[atmeldemo]: https://github.com/atmelcorp/atmel-software-package/blob/master/examples/crypto_icm/main.c

jbeaurivage:
@vcchtjader could you please open up a quick new PR to add your changes to the HAL's `CHANGELOG.md`?

PR #485, opened 2021-09-15
eic fix unused borrow that must be used

jacobrosenthal:
Been getting these with 1.55.0

```
warning: 	
   --> /home/jacob/Downloads/atsamd/hal/src/thumbv7em/eic/pin.rs:103:17
    |
103 | /                 &(*target_device::EIC::ptr()).intflag.write(|w| {
104 | |                     w.bits(1 << $num)
105 | |                 });
    | |__________________^
...
388 | / ei!(ExtInt[15] {
389 | |     Pa15,
390 | |     Pa31,
391 | |     #[cfg(feature = "min-samd51j")]
...   |
400 | |     Pc31,
401 | | });
    | |___- in this macro invocation
    |
    = note: this warning originates in the macro `ei` (in Nightly builds, run with -Z macro-backtrace for more info)
 ```

### DIFF
--- a/hal/src/thumbv6m/eic/pin.rs
+++ b/hal/src/thumbv6m/eic/pin.rs
@@ -95,11 +95,11 @@ crate::paste::item! {
         }
 
         pub fn is_interrupt(&mut self) -> bool {
-            unsafe { (*target_device::EIC::ptr()) }.intflag.read().[<extint $num>]().bit_is_set()
+            unsafe { &(*target_device::EIC::ptr()) }.intflag.read().[<extint $num>]().bit_is_set()
         }
 
         pub fn clear_interrupt(&mut self) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by AfoHT** [comment]
I also ran into this issue, it seems the changes made in `thumbv6` aren't needed, because the reference are used outside of the `unsafe` block.

Changing pin.rs for thumbv7 to 

```
        pub fn clear_interrupt(&mut self) {
            unsafe {
                {&(*target_device::EIC::ptr())}.intflag.write(|w| { w.bits(1 << $num) });
            }
        }
```
solves it for me because now with the extra block the reference is "used", while just removing the reference errors with trying to move a non-Copy value, full error available in CI error logs.

Hope this can be of some use

TDHolmes (approved):

PR #486, opened 2021-09-17
Pukcc: GF(p)'s ECDSA signature generation/validation + full C-style ABI

vccggorski:
This PR provides easy to use, high-level API for a small subset of PUKCC capabilities. Namely, ECDSA signature generation and validation over prime fields.


hal/src/thumbv7em/pukcc/c_abi.rs
```diff
@@ -0,0 +1,822 @@
...
+
+#[repr(C)]
+#[derive(Copy, Clone, Default, Debug)]
+pub struct ZpEccQuickDualMulFast {
+    pub pu1ModCnsBase: pu1,
+    pub pu1PointABase: pu1,
+    pub pu1PointBBase: pu1,
+    pub pu1KABBase: pu1,
+    pub pu1AWorkBase: pu1,
+    // pu1       pu1ABase,
```
vcchtjader, line 507: Not needed?

vcchtjader (commented):
I tested this on hardware with [atsamd-demo pukcc-experimentation](https://github.com/vcchtjader/atsamd-demo/tree/pukcc-experimentation) and it produces valid results.

I do not see any obvious flaws. The C-ABI naming conventions draws directly from the Atmel/Microchip implementation used inside the ROM if I understood it correctly.

Some minor spelling nitpicking

hal/src/thumbv7em/pukcc.rs
```diff
@@ -0,0 +1,579 @@
...
+        unsafe { c_abi::ZpEcDsaGenerateFast::call(&mut pukcl_params) };
+
+        match pukcl_params.header.u2Status.into() {
+            PukclReturnCode::Ok => {}
+            error_code => return Err(EcdsaSignFailure::ServiceFailure(error_code)),
+        };
+
+        // Generated signature R part is written to base point X coordinate memory.
+        // Generated signature S part is written to base point Y coordinate memory.
+        // Base point Z coordnate should be zero.
```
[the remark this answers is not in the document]
vccggorski: Fixed

hal/src/thumbv7em/pukcc.rs
```diff
@@ -0,0 +1,579 @@
...
+            Ok => 0x0000,
+            Severe(PukclReturnCodeSevere::ComputationNotStarted) => 0xC001,
+            Severe(PukclReturnCodeSevere::UnknownService) => 0xC002,
+            Severe(PukclReturnCodeSevere::UnexploitableOptions) => 0xC003,
+            Severe(PukclReturnCodeSevere::HardwareIssue) => 0xC004,
+            Severe(PukclReturnCodeSevere::WrongHardware) => 0xC005,
+            Severe(PukclReturnCodeSevere::LibraryMalformed) => 0xC006,
+            Severe(PukclReturnCodeSevere::Error) => 0xC007,
+            Severe(PukclReturnCodeSevere::UnknownSubservice) => 0xC008,
+            Severe(PukclReturnCodeSevere::OverlapNotAllowed) => 0xC010,
```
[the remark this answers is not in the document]
vccggorski: Resolved

hal/src/thumbv7em/pukcc/c_abi.rs
```diff
@@ -0,0 +1,822 @@
+//! Module that defines low-level constructs required for an interaction with
```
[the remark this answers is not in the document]
vccggorski: Done

hal/src/thumbv7em/pukcc/c_abi.rs
vccggorski: Removed all comments and weird whitespace from `c_abi` module

hal/src/thumbv7em/pukcc.rs
```diff
@@ -0,0 +1,579 @@
...
+    /// - `hash`: `&[u8]` of length [`Curve::SCALAR_LENGTH`]
+    ///     - Hash of a message that is signed.
+    /// - `public_key`: `&[u8]` of length [`Curve::SCALAR_LENGTH`]
+    ///     - Public key used for a signature verification.
+    ///
+    /// Return value:
+    /// - `Result::Ok`
+    ///     - Signature is valid against chosen `hash` and `public_key`
+    /// - `Result::Err`
+    ///     - Possible failure scenarios are encapsulated in a
```
twitchyliquid64, line 288: Is it possible that this could be a padding/error oracle? In other words, could different kinds of failures indicate information about the key to an attacker?

If so, it might be worth adding a disclaimer to not expose the details from `EcdsaSignatureVerificationFailure` outside of debugging.

hal/src/thumbv7em/pukcc.rs
```diff
@@ -0,0 +1,579 @@
...
+    __: (),
+}
+
+impl Pukcc {
+    /// Constructor.
+    ///
+    /// Waits for a CryptoRAM readiness, enables a synchronous PUKCC clock and
+    /// performs a self test. In case a self test fails it returns an error
+    pub fn enable(mclk: &mut MCLK) -> Result<Self, SelfTestFailure> {
+        // TODO: Alleged RNG initialization. Passthrough constructor?
```
twitchyliquid64, line 68: How do you mean? Is the RNG initialized? is there a way to check?

The module should fail to operate if it cannot verify a RNG is operating, either because it initialized it itself, or it checked some status registers.

Another way to do it is require some struct be passed in that can only be obtained by initializing the RNG elsewhere.

hal/src/thumbv7em/pukcc.rs
```diff
@@ -0,0 +1,579 @@
...
+    }
+
+    /// Service generating an ECDSA signature.
+    ///
+    /// GF(p) service. GF(2^n) variant is not implemented -- use low-level API.
+    ///
+    /// Input parameters:
+    /// - `hash`: `&[u8]` of length [`Curve::SCALAR_LENGTH`]
+    ///     - Hash of a message that is supposed to be signed.
+    /// - `private_key`: `&[u8]` of length [`Curve::SCALAR_LENGTH`]
```
twitchyliquid64, line 119: Aside from the length property, is it possible to pass in a private key that is invalid? If so, can we create a new type that represents a private key with valid parameters, and only have safe ways of creating that type?

hal/src/thumbv7em/pukcc.rs
```diff
@@ -0,0 +1,579 @@
...
+    ///     - Private key used for signing.
+    /// - `k`: `&[u8]` of length [`Curve::SCALAR_LENGTH`]
```
twitchyliquid64, line 121: It would be my personal preference if this wasnt a parameter and instead the module obtained entropy from some hardware source every time.

Other ideas:
 - Make a newtype `EntropySeed`, which is passed by value (so it is consumed). This newtype can be an enum with variants `FromHardware` (in which case the module obtains the entropy for you) or `InsecureFromSeed` which has the entropy contained within.
 - Separate `zp_ecdsa_sign` into two functions, one where the user provides the entropy (ie: `insecure_zp_ecdsa_sign_with_seed`) and the regular one which obtains the entropy for you (`zp_ecdsa_sign`)

twitchyliquid64 (commented):
Thanks for implementing this! PUKCC is not an easy peripheral.

Overall this looks good, though I think it would be great if we could tighten up the API so its harder to mis-use.

In particular:
 - Making it hard to use this without real entropy
 - Making it hard (or with hard-to-miss disclaimers) to expose error codes back to the user except for debugging
 - Making it hard to use invalid curves (though you may have already done that with `verify_curve()`

Its also worth adding a strongly-worded disclaimer to the module saying this has not been evaluated for correctness / side channels / oracle scenarios etc.

If you have an example we can whack in a BSP that evaluates test vectors, that would be great too.

hal/src/thumbv7em/pukcc.rs
vccggorski: No. According to 43.3.6.12.7 in datasheet it is either "ok" or "wrong signature". No specifics. For signature generation (43.3.6.11.7) it is either "ok" or "wrong select number" which is `k`.

hal/src/thumbv7em/pukcc.rs
vccggorski: Documentation puts a lot of stress on "having an RNG module enabled compulsorily". Like for example on a page 1382. I don't know if these hidden functions discreetly use RNG related peripherals and this is why you are expected to make sure they are enabled or not. 

ATM I think they are just meant to vaguely remind a programmer to use RNG when necessary while generating input parameters for these services. Like `k` for instance. So probably I should remove this TODO.

hal/src/thumbv7em/pukcc.rs
vccggorski: I don't know. I'm not even sure if these lengths for slices are correct as I relied on a documentation and its max lengths fields from these tables. My cryptographical expertise is also limited, to say the least. Plus documentation is full of typos and errors tbf. Private key is just a [u8;32] for this implemented curve. I don't think we can prevent the user from passing slice of zeroes for example :shrug:

hal/src/thumbv7em/pukcc.rs
vccggorski: What does it mean it "obtains entropy for you"? I guess you mean you pass `impl RngCore` into the safe variant which then is used to populate a `k` slice locally.

vccggorski:
@twitchyliquid64 
There's also a question of how to deal with CNS. Documentation, as far as I can read, *does not require* CNS parameter filled in for these specific services. Yet, without it, it produces bogus results: signatures are scrambled; all signatures are invalid. Probably it is preferable to generate this constant everytime? As far as I understand, it should be constant against a specified P modulus, so having it as a part of curve parametrization should be fine (I guess). Also I'm pretty convinced I messed up endianness in CNS but it gives correct results so I left it there as it is :D I'll create a test vector in my spare time.

twitchyliquid64:
> There's also a question of how to deal with CNS. Documentation, as far as I can read, does not require CNS parameter filled in for these specific services. Yet, without it, it produces bogus results

From what I can tell CNS is a fixed parameter used for doing modulus arithmetic in fast mode, but only for some services. Other services seem to specify the CNS parameter as working memory for other computations?

Maybe a safe starting point would be to always initialize CNS to stack memory with the correct value for the current curve?

PR #487, opened 2021-09-19
update feather_m4 bsp and examples to v2 as much as possible

TDHolmes:
Wasn't too bad, mostly used `feather_m0`'s transformation as a guide. I was able to test:
  - blinky_basic
  - dmac
  - neopixel_rainbow
  - pwm
  - sleeping_timer_rtc
  - timers
  - trng
  - usb_echo

I couldn't validate:
  - uart
  - uart_poll_echo
  - serial

Unfortunately, `sleeping_timer_rtc` seems to not work. As that is my example, I'm going to take a look at this and see if I can fix it and/or if it's limited to M4


TDHolmes:
`sleeping_timer_rtc` also fails on `feather_m0`, so doesn't seem related to this refactor (that would've been shocking...)

boards/feather_m4/src/lib.rs
```diff
@@ -1,103 +1,156 @@
...
+        /// The USB D- pad
+        name: usb_dm,
+    }
+    PA25 {
+        /// The USB D+ pad
+        name: usb_dp,
+    }
 );
 
 /// Convenience for setting up the labelled SPI peripheral.
```
jbeaurivage, line 156: @TDHolmes, thanks for working on this. I think the only thing I see missing is switching to `spi::v2` in `lib.rs`.

jbeaurivage:
@TDHolmes, thanks for working on this. I think the only thing I see missing is switching to `spi::v2` in `lib.rs`.

TDHolmes:
I just updated to v2 SPI, but I am not confident in the change I made to be honest. Looking at the data sheet, it seems like the SPI hookup used by the feather board should work. The IoSet doesn't seem to match up even though they all are on Sercom1.
```
PA17 Sclk - IoSet1
PB23 Mosi - IoSet3
PB22 Miso - IoSet3
```

I added IoSet3 as another option for PA17 on Sercom1. @jbeaurivage / @bradleyharden is it possible the data sheet just omitted this?

image

bradleyharden:
I won't try auto merge for this. I expect there will be a conflict with #506. But it should be ready for merge

TDHolmes:
I'll need to edit this a bit after rebasing. I'll ping ya when I do it later tonight

bradleyharden:
Have you seen #445? Could you add that pin as well?

TDHolmes:
@bradleyharden ah let me sneak that in

TDHolmes:
@bradleyharden should be good to go!

PR #489, opened 2021-09-24
Remove unused panic-abort and unnecessary panic-halt feature/dependency

TDHolmes:
Some boards have default panic-halt feature and dependency, but this doesn't make sense for users of the BSPs. This dependency is a dev-dependency for the examples. Also, consistency around examples using halt or semihosting panic option was varried on the boards that had this issue, so tried to clean that up as well.

### DIFF
--- a/boards/arduino_mkr1000/Cargo.toml
+++ b/boards/arduino_mkr1000/Cargo.toml
@@ -7,6 +7,7 @@ keywords = ["no-std", "arm", "cortex-m", "embedded-hal"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/atsamd-rs/atsamd"
 readme = "README.md"
+edition = "2018"
 
 [dependencies]
 cortex-m = "0.6"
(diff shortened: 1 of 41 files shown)


TDHolmes:
I found this when trying to trim unnecessary dependencies from one of my projects

jbeaurivage:
@TDHolmes, are these the only 3 BSPs that needed cleanup?

TDHolmes:
Looks like I might have missed a few:
- [ ] arduino_mkrvidor4000
- [ ] arduino_mkrzero

TDHolmes:
alright @jbeaurivage should be good now. Lots of little changes in this one

jbeaurivage:
@TDHolmes this might be a nit, so please take it as a suggestion as we will do the cleanup in another PR, but we're trying to harmonize the BSP examples and the crates they import. I think we settled on using `as hal` only for `atsamd-hal`, and import the relevant BSP `as bsp`.

TDHolmes:
@jbeaurivage yeah I've seen that and it definitely makes it more clear, but as the scope of this PR is creeping already a bit another PR seems like a good idea to me - which I'm happy to do

jbeaurivage:
@TDHolmes sounds good; merging now then

PR #491, opened 2021-09-25
Update to latest method of bsp/hal import style and rust edition

TDHolmes:
As promised @jbeaurivage, here's my crack at updating everything to our latest way of doing things.


bradleyharden:
Did you do all of this by hand?!

TDHolmes:
Haha, no most of it was regex find and replace

jbeaurivage:
While we're at it, how do you feel about trying to get rid of all references to `target_device`? The HAL already aliases it as `pac`, and we're eventually (soon) phase out `target_device` in favour of `pac`.

TDHolmes:
If the PR wasn't massive enough yet haha. Was pretty easy find/replace though.

PR #495, opened 2021-09-27
release-crates workflow: Add sleep between steps to let index update

twitchyliquid64:


TDHolmes (changes requested):
this just feels like a hack, maybe we could instead poll for it to exist or something?

jbeaurivage:
Superseded by #762.

PR #497, opened 2021-09-29
feat: Add support for Adafruit Neokey Trinkey board

BroderickCarlin:


boards/neokey_trinkey/Cargo.toml
```diff
@@ -0,0 +1,74 @@
...
+keywords = ["no-std", "arm", "cortex-m", "embedded-hal"]
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/atsamd-rs/atsamd"
+readme = "README.md"
+edition = "2018"
+
+[dependencies]
+cortex-m = "0.6"
+embedded-hal = "0.2.3"
+nb = "0.1"
```
TDHolmes, line 15: what is using `nb`?

boards/neokey_trinkey/Cargo.toml
```diff
@@ -0,0 +1,74 @@
...
+version = "0.2"
+optional = true
+
+[dependencies.usbd-serial]
+version = "0.1"
+optional = true
+
+[dependencies.panic-halt]
+version = "0.2"
+optional = true
```
TDHolmes, line 36: no need for this, just do dev-dependency for the examples

boards/neokey_trinkey/Cargo.toml
```diff
@@ -0,0 +1,74 @@
...
+
+[dev-dependencies]
+panic-halt = "0.2"
+
+[features]
+# ask the HAL to enable atsamd21e support
+default = ["rt", "atsamd-hal/samd21e"]
+leds = ["ws2812-timer-delay", "smart-leds"]
+panic_halt = ["panic-halt"]
```
TDHolmes, line 54: no need for this feature

boards/neokey_trinkey/examples/usb_echo.rs
```diff
@@ -0,0 +1,95 @@
...
+                };
+            });
+        });
+    };
+}
+
+#[interrupt]
+fn USB() {
+    poll_usb();
+}
```
TDHolmes, line 95: is this example significantly different than usb_ack? any reason to have both?

boards/neokey_trinkey/Cargo.toml
```diff
@@ -0,0 +1,74 @@
...
+cortex-m = "0.6"
+embedded-hal = "0.2.3"
+nb = "0.1"
+
+[dependencies.cortex-m-rt]
+version = "0.6.12"
+optional = true
+
+[dependencies.atsamd-hal]
+path = "../../hal"
```
TDHolmes, line 22: This would make it a tier 1 board - not sure what the guidelines will be around new boards, if they should be tier 1 or tier 2 automatically

boards/neokey_trinkey/Cargo.toml
```diff
@@ -0,0 +1,74 @@
...
+panic-halt = "0.2"
+
+[features]
+# ask the HAL to enable atsamd21e support
+default = ["rt", "atsamd-hal/samd21e"]
+leds = ["ws2812-timer-delay", "smart-leds"]
+panic_halt = ["panic-halt"]
+rt = ["cortex-m-rt", "atsamd-hal/samd21e-rt"]
+unproven = ["atsamd-hal/unproven"]
+use_semihosting = []
```
TDHolmes, line 57: Could we update the examples to key off of this feature to do different panic behavior like [other examples](https://github.com/atsamd-rs/atsamd/blob/1f7517177beb8ccd08748a19b12bf1cbf683fe23/boards/feather_m4/examples/neopixel_rainbow.rs#L16)

I'm also open to just removing this feature

boards/neokey_trinkey/Cargo.toml
jbeaurivage: The guidelines dictate that boards are Tier 2 by default. @BroderickCarlin, this means that the BSP should reference a crates.io version of `atsamd-hal` rather than a path dependency.

jbeaurivage:
You would also need to add your board to `crates.json` so it gets picked up by CI

TDHolmes:
👀

Issue #481, opened 2021-09-06
More Frequent Releases | Moving towards CI/CD


LongLiveCHIEF:
I was away from my work that was using atsamd crates for a few months, and came back to find many of the things I was waiting on that were in the works previously, have now been merged!

Unfortunately though, no new BSP/PAC/HAL's have been released.  This has made it difficult to use these features and continue to patch in features from other contributors for things in the works, as I have to do a lot of merging and cherry-picking locally against the upstream master.

If there were more frequent releases (say for example a minor/patch ), it would be easier to patch in prospective features as I develop laterally on things like BSP's.

I've taken a look at existing actions, and think they are a good start to this.  I have some questions, and based on those answers, have some suggestions.

I'd also be willing to contribute much of the work I propose.  This issue will serve as a central issue to some of the smaller things that are capable of being isolated into scoped issues/PR's.

### Questions

- Am I right in seeing that releasing a PAC or BSP releases _all_ of them? There's no workflow that allows to release a BSP independent of the other BSP's?  
  - Ditto for bumping versions of the BSP/PAC?
  
  
  ### Suggestions

**Filtered Build/Release workflows** It looks like each PR builds and tests all BSP's, regardless of which BSP's (if any) are changed.  We could make it possible to build and release individual BSP's based on the file changes detected in the push/PR. For example, I've done this in the [Octoprint/octoprint-docker project](https://github.com/OctoPrint/octoprint-docker/blob/master/.github/workflows/build.yml), where I selectively build and release the various image variants based on detected file changes.

**Add input to manual release workflow to allow individual crates to be bump-released**.  We can use a free-form input for the Release crates workflow to allow a privileged maintainer to enter the name of the crate to be released

**Add a step to aggregate/update changelogs and release notes**  We can grab a list of all PR's since the last release of an affected crate, and automatically generate a changelog and release notes based on these changes. This is something I did back when I maintained some official puppet modules, and it worked really well.  [Here's how that worked](https://voxpupuli.org/docs/releasing_version/).  Ignore the Ruby pieces, and just pay attention to their groups similarity to our project structure and the logic behind their releases and changelog generation.  That group serves as the "official" modules of the puppet forge ecosystem, in much the same way the atsamd project is a monorepo with many packages.

We could do the same thing, but with github actions.

### Closing Thoughts

[...]

Issue #503, opened 2021-10-04
Using [package.metadata] in each crate instead of maintaining a Crates.json file


LongLiveCHIEF:
what do you guys think of using the `[package.metadata.tier]` format in the `Cargo.toml` so we can get rid of having to maintain a `Crates.json` file?

It would make things much simpler, especially as we move into auto-generating `CHANGELOG` and even auto-updating `README`with list of tiered crates

likewise for `[package.metadata.build_invocation]`
each crate would add those two metadata fields, and specify their tier and build invocation, and then I can use that in the build matrix to modify the README accordingly as well as invoke any build instructions
https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table

This would also easily allow the build stage to check to ensure any new crates have the appropriate information required in the Cargo file, enabling us to dynamically add crates (especially BSP's) very easily.

LongLiveCHIEF:
maybe just a single `[package.metadata.atsamd_metadata]`:

```
[package.metadata.atsamd_metadata]
crate_type = BSP # BSP | PAC | HAL
tier = 2 # only applicable for 
build_invocation = "cargo build --examples --features=unproven,usb"
```

jbeaurivage:
This makes perfect sense to me!

LongLiveCHIEF:
related #481

LongLiveCHIEF:
Proposal spec:

---

### Adding a new BSP/PAC

The build and release workflows will automatically recognize any new crate added to the repo,
but you will need to add some metadata to the crates's `Cargo.toml`.

```toml
# example of metadata for a BSP
[package.metadata.atsamd]
# also_build (optional): indicate if any additional crates (HAL | PAC | BSP's) should be built
# to properly test this crate
# also_build = ["crate", "crate2"]
# room for other generic high-level hooks for pipeline/build customization

[package.metadata.atsamd.bsp]

# tier (required)
# see: https://github.com/atsamd-rs/atsamd#how-to-use-a-bsp-ie-getting-started-writing-your-own-code
# for determining what tier your BSP is
tier = 2

# build_command (required)
# used to invoke the appropriate build command for the bsp
build_command = "cargo build --examples --features=unproven"

[package.metadata.atsamd.pac]
# target (required): the build target, used to configure the runner properly
target = "thumbv6m-none-eabi"
# features to build the pac with
features = ["samd21g", "unproven", "usb"]
```

sajattack:
[...]

PR #504, opened 2021-10-04
Filter workflows to affected crates

LongLiveCHIEF:
This is first in a series of PR's meant to implement packaging, deployment and release changes discussed in #481.  This will be merged to the `feature/ci-cd-dev` branch, instead of `master`, so that this effort can have smaller PR's that can be reviewed individually for their scope of change, before making any alterations to the 'production' pipeline.

# Summary
This PR implements a `build-test` workflow, which automatically determines crates affected by the change, and creates a matrix consisting of the affected crate and the `nightly` and `stable` toolchains.

Example:  if a change was made to any file in both `/pac/atsamd11c`, and `boards/circuit_playground_express`.  This action would run a `generate-matrix` action, then determine that 4 build jobs need to be run:

- pac/atsamd11c, stable
- pac/atsamd11c, nightly
- boards/circuit_playground_express, stable
- boards/circuit_playground_express, static

### Other Notes

- matrix will include a crate that has _any_ file added/deleted/modified/renamed within it's file tree
- matrix will not run a `build` if none of the files affected are a child of a crate directory 

### In-Scope for this PR

- creating a filtered matrix of crates affected by changes, to be used by jobs and workflows in the pipeline

### Out-of-scope for this PR

- changes to build, test, release, or package versioning

### Example runs:

- [test of deleting a file in a BSP](https://github.com/LongLiveCHIEF/atsamd/actions/runs/1304428221)
- [test of adding a file to a BSP](https://github.com/LongLiveCHIEF/atsamd/actions/runs/1304392848)


LongLiveCHIEF:
for now, I went with js based action, since I'm not as proficient with python. I wanted to start getting the concept out there so we could see what other parts of the pipeline this concept would affect.  We can always come back and make it python later/before we adopt this to master.

For any reviewers, please leave a note if your approval is contigent upon this being done in python prior to the new ci/cd being adopted into production

TDHolmes:
If `atsamd11c` was touched, shouldn't the BSC for `SAMD11_bare` also be triggered since it depends on it?

jbeaurivage:
# Usage of JS

Personally I don't really mind going with js, as long as we don't end up with a disparate collection of scripts in different languages. The only caveat I would see is that to me python is more readable to someone unfamiliar with the language, for example someone trying to understand why the CI is failing.

# BSP triggered by PAC
@TDHolmes has a good point, especially for PAC builds, would the dependency in the BSP and HAL crates be updated automatically - and their version bumped? Or is that a manual step?

Similarly for tier 1 bsps, is there a mechanism we can put in place to automatically sync the hal version in their dependencies, and perhaps bump the (T1) bsp versions if the dependency is updated?

LongLiveCHIEF:
> If `atsamd11c` was touched, shouldn't the BSC for `SAMD11_bare` also be triggered since it depends on it?

I wasn't aware of the BSC's in the `boards` directory, so this got by me.  However, I _did_ propose an [`also_build`](https://github.com/atsamd-rs/atsamd/issues/503#issuecomment-933827073) for the `[package.metadata.atsamd]` in #503 just in case something like this came up.

That being said... it does feel like a code smell that there's a downstream crate that introduces build dependencies to an upstream crate.

> Similarly for tier 1 bsps, is there a mechanism we can put in place to automatically sync the hal version in their dependencies, and perhaps bump the (T1) bsp versions if the dependency is updated?

Yes, in fact, the [sync-pac-versions](https://github.com/atsamd-rs/atsamd/tree/master/.github/actions/sync-pac-versions) action in the release workflow does almost exactly this already (it just does it to _every_ pac crate, instead of just _affected_ pac/bsc crates).

LongLiveCHIEF:
> The only caveat I would see is that to me python is more readable to someone unfamiliar with the language, for example someone trying to understand why the CI is failing.

Agreed... python is _much_ easier to reason through at a glance.  So far, all the maintainers have basically said js is ok with them, and if we get through all this, i'll likely be the primary maintainer, and js is a language I can do in my sleep.

My primary reason for using js here is that I wanted to quickly see what the affects of this filtering would be, for which this has definitely succeeded.

> as long as we don't end up with a disparate collection of scripts in different languages.

Agreed, I think we should stick to one language (either python or js) in addition to any bash needed.  That is one of my acceptance requirements for when we merge the `feature/ci-cd-dev` branch to master.

So, I'll either redo this in python... or remove any python through the remaining work.

I'm leaning more towards redoing this in python, as I think the users of these crates are more likely to have python experience/understand python... than they are js.

LongLiveCHIEF:
> @TDHolmes has a good point, especially for PAC builds, would the dependency in the BSP and HAL crates be updated automatically - and their version bumped? Or is that a manual step?

I think with the discussion in #481, the implication is that we wouldn't automatically bump the dependencies in a crate when the dependency updates.  Reasons for this:

- There aren't any tests for most crates, and just running a build does not guarantee functionality, 
- Not updating dependencies in a dependent crate decouples the PAC/HAL from their dependents, which makes it easier to implement changes to the codebase without maintaining several versions of an interface in an active codebase.
- We can update rules for Tier 1 crates that state they will maintain forward compatibility with HAL/PAC, up to breaking changes. (will be compatible with Y or Z changes).
- We can create an action that automatically determines if a crate is currently meeting Tier 1 rules, and adjust the listings in README and cargo.toml files accordingly automatically on each merge to master.

LongLiveCHIEF:
Unless there are no other concerns, let's merge for the moment so I can use the output of this in the other pieces of work for this feature.  I can come back and re-do in python prior to releasing into the wild

TDHolmes:
I kinda agreed with the code smell bit… but also agreed that this can be
iterated on and tweaked later

PR #505, opened 2021-10-05
Fix cortex-m/cortex-m-rt version selection issues

TDHolmes:
# Summary
CI started failing recently, not being able to select the right version of `cortex-m`/`cortex-m-rt`. Update some of our dependencies to fix this. Unfortunately updating this everywhere resulted in too many errors. I think the embedded ecosystem needs some time to upgrade.

# Checklist

### DIFF
--- a/boards/feather_m0/CHANGELOG.md
+++ b/boards/feather_m0/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Bump `cortex-m`/`cortex-m-rt` dependencies to fix a build issue
+
 # 10.0.1
 
 * Bump dependencies `rtic-monotonic` to `0.1.0-rc.1` and `cortex-m-rtic` to `0.6.0-rc.2`.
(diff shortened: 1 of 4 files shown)


jbeaurivage:
At this point shouldn't we just bump all PAC/BSP/HAL to the latest cortex-m and cortex-m-rt?

TDHolmes:
Like I said in the summary, I tried but hit errors in some BSC's. Feel free to try though, maybe you can figure them out

jbeaurivage:
Ah whoops I read the description too fast. Are the errors something about duplicate `INTERRUPT` symbols in the linker or something?

TDHolmes:
It’s the same error as the one I’m fixing here, but for only some BSCs and
presumably like the opposite - some other dependencies outside of our
family of crates is dependent on an old version. I think.

bradleyharden:
Would changing the feature resolver affect this at all?

PR #538, opened 2021-11-05
Update xiao m0 library to use v2 gpio and sercom

avan1235:
# Summary
Updated xiao m0 library to use up to date gpio and sercom api and produce no deprecation warnings when compiling. Updated examples to use new code and provide new examples tested with physical screen and accelerometer.

# Checklist

### DIFF
--- a/boards/xiao_m0/CHANGELOG.md
+++ b/boards/xiao_m0/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 0.11.0
+
+- update gpio and sercom dependencies to v2
+- create shared_i2c example with I2C bus used for SSD1306 OLED screen and accelerometer
+- create ssd1306_i2c example with basic usage of OLED screen for animation with I2C
+- update blink example to be more readable for newcomers
+- clean up usb_echo example (and extend of blinking on data transfer)
+
 # Unreleased
 
 - remove extraneous `embedded-hal` dependencies from BSPs
--- a/boards/xiao_m0/CHANGELOG.md
+++ b/boards/xiao_m0/CHANGELOG.md
@@ -5,9 +5,6 @@
 - create ssd1306_i2c example with basic usage of OLED screen for animation with I2C
 - update blink example to be more readable for newcomers
 - clean up usb_echo example (and extend of blinking on data transfer)
-
-# Unreleased
-
 - remove extraneous `embedded-hal` dependencies from BSPs
 - cleanup `cortex_m` dependency
 - move `usbd-x` crates used only in examples to `[dev-dependencies]`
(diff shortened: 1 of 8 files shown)


bradleyharden:
Gave this a quick once over and it looks good to me

boards/xiao_m0/src/lib.rs
```diff
@@ -1,160 +1,204 @@
...
-    pin a5 = a9,
-    /// Pin A6/D6/TX
-    pin a6 = b8,
-    /// Pin A7/D7/RX
-    pin a7 = b9,
-    /// Pin A8/D8/SCK
-    pin a8 = a7,
-    /// Pin A9/D9/MISO
-    pin a9 = a5,
-    /// Pin A10/D10/MOSI
```
TDHolmes, line 52: could we keep these doc comments on the pins?

boards/xiao_m0/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
...
+
+- update gpio and sercom dependencies to v2
+- create shared_i2c example with I2C bus used for SSD1306 OLED screen and accelerometer
+- create ssd1306_i2c example with basic usage of OLED screen for animation with I2C
+- update blink example to be more readable for newcomers
+- clean up usb_echo example (and extend of blinking on data transfer)
+
 # Unreleased
 
 - remove extraneous `embedded-hal` dependencies from BSPs
```
TDHolmes, line 11: If we're doing a release, move all items in `# Unreleased` to your release and put an empty `# Unreleased` at the top.

TDHolmes (changes requested):
`CHANGELOG.md` needs tweaks

TDHolmes:
thanks for updating this board!! I'm looking forward to not seeing deprecation warnings someday in the future 😅

boards/xiao_m0/src/lib.rs
avan1235: I reverted these names - could you explain to e where they are actually used/visible? I don't see these macros expanded in my IDE so they weren't useful for me

avan1235:
Thanks for fast response 😃 I just started playing with rust and this small xiao board so hope to provide more examples if it's convenient to put them here

boards/xiao_m0/src/lib.rs
TDHolmes: Having the docstrings allows documentation like this to be produced: https://docs.rs/feather_m4/0.8.0/feather_m4/struct.Pins.html#structfield.a0

TDHolmes:
I personally use VS Code + `rust-analyzer`, which works pretty well, but getting it to properly enable the right features for this library can be tricky, since it supports so many different targets

TDHolmes (approved):

TDHolmes:
Thanks @avan1235 , published the new version.

PR #542, opened 2021-11-11
Update `wio-terminal` to use v2 APIs

jbeaurivage:
# Summary

* Update `wio-terminal` to remove all uses of the `v1` APIs, thus completing its T1 status
* Remove a bunch of clippy warnings
* Update `cortex-m` and `cortex-m-rt` to `0.7`, `bbqueue` to `0.5`

# Checklist
  - [ ] ~~All new or modified code is well documented, especially public items~~

## If Adding a new Board
  - [ ] ~~Board CI added to `crates.json`~~

### DIFF
--- a/boards/wio_terminal/src/display.rs
+++ b/boards/wio_terminal/src/display.rs
@@ -1,9 +1,8 @@
 use atsamd_hal::clock::GenericClockController;
-use atsamd_hal::gpio::v2::*;
+use atsamd_hal::ehal::digital::v2::OutputPin;
 use atsamd_hal::hal::blocking::delay::DelayMs;
 use atsamd_hal::hal::spi::{Phase, Polarity};
 use atsamd_hal::pac::{MCLK, SERCOM7};
-use atsamd_hal::prelude::*;
 use atsamd_hal::sercom::v2::spi;
 use atsamd_hal::sercom::v2::{IoSet4, Sercom7};
 use atsamd_hal::time::Hertz;
@@ -11,7 +10,7 @@ use atsamd_hal::typelevel::NoneT;
 use display_interface_spi::SPIInterface;
 use ili9341::{DisplaySize240x320, Ili9341, Orientation};
 
-use super::pins::*;
+use super::pins::aliases::*;
 
 /// ILI9341 LCD display pins (uses `SERCOM7`)
 pub struct Display {
--- a/boards/wio_terminal/src/display.rs
+++ b/boards/wio_terminal/src/display.rs
@@ -1,6 +1,5 @@
 use atsamd_hal::clock::GenericClockController;
-use atsamd_hal::gpio::v2::{PB19, PB20};
-use atsamd_hal::gpio::*;
+use atsamd_hal::gpio::v2::*;
 use atsamd_hal::hal::blocking::delay::DelayMs;
 use atsamd_hal::hal::spi::{Phase, Polarity};
 use atsamd_hal::pac::{MCLK, SERCOM7};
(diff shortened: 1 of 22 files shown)


boards/wio_terminal/src/sensors.rs
```diff
@@ -9,24 +10,31 @@ use atsamd_hal::sercom::{I2CMaster4, PadPin, Sercom4Pad0, Sercom4Pad1};
...
+    D: Into<I2c0Sda>,
+{
     /// `I2C0` bus clock pin
-    pub scl: Pa12<Input<Floating>>,
+    pub scl: C,
 
     /// `I2C0` bus data pin
-    pub sda: Pa13<Input<Floating>>,
+    pub sda: D,
 }
```
bradleyharden, line 23: Shouldn't the `struct` hold `I2c0Scl` and `I2cSda` directly, instead of being generic? You only need `Accelerometer::init` to be generic

boards/wio_terminal/src/sensors.rs
```diff
@@ -47,24 +55,28 @@ impl Accelerometer {
...
 /// Analog Light Sensor
-pub struct LightSensor {
+pub struct LightSensor<P>
+where
+    P: Into<LightSensorAdc>,
+{
     /// Analog Light Sensor input pin
-    pub pd1: Pd1<Input<Floating>>,
+    pub pd1: P,
 }
```
bradleyharden, line 64: Same comment as above. I think the struct should hold a concrete type, not a generic type. The `init` function should be generic.

boards/wio_terminal/src/sensors.rs
bradleyharden: Oh wait, this is the whole "Sets" approach. In that case, it still shouldn't be generic, but instead should hold `Pin<PA12, Disabled<Floating>>`, otherwise known as `Pin<PA12, Reset>`. Or you could do like I did in the `pygamer` PR and add an additional alias as `I2cSclReset` or something.

boards/wio_terminal/src/wifi.rs
```diff
@@ -23,77 +20,107 @@ use seeed_erpc as erpc;
...
+{
+    pub pwr: Pwr,
+    pub rxd: Rxd,
+    pub txd: Txd,
+    pub mosi: Mosi,
+    pub clk: Clk,
+    pub miso: Miso,
+    pub cs: Cs,
+    pub ready: Ready,
+    pub dir: Dir,
```
bradleyharden, line 43: This would be much simpler with the `Pin<PXXX, Reset>` approach.

boards/wio_terminal/src/display.rs
```diff
@@ -91,8 +87,8 @@ impl Display {
         // does not appear to support PWM.
         //   HIGH - backlight enabled
         //   LOW  - backlight disabled
-        let mut backlight = self.backlight.into_push_pull_output(port);
-        backlight.set_high()?;
+        let mut backlight = self.backlight.into_push_pull_output();
+        backlight.set_high().ok();
```
TDHolmes, line 91: any reason not to still return the error like before? I know it's impossible, but 🤷🏼

boards/wio_terminal/src/wifi.rs
```diff
@@ -1,99 +1,106 @@
...
         delay.delay_ms(10u8);
 
         // Reset the RTL8720 MCU.
-        let mut pwr = pins.pwr.into_push_pull_output(port);
-        pwr.set_low()?;
+        let mut pwr: WifiPwr = pins.pwr.into();
+        pwr.set_low().ok();
         delay.delay_ms(100u8);
-        pwr.set_high()?;
+        pwr.set_high().ok();
```
TDHolmes, line 87: Same Q as earlier

hal/src/thumbv7em/usb/mod.rs
```diff
@@ -11,8 +11,11 @@ mod devicedesc;
 use self::devicedesc::Descriptors;
 
 /// Default SOF pad
+#[allow(deprecated)]
 pub type SofPad = gpio::v1::Pa23<gpio::v1::PfH>;
 /// Default USB D- pad
+#[allow(deprecated)]
 pub type DmPad = gpio::v1::Pa24<gpio::v1::PfH>;
 /// Default USB D+ pad
+#[allow(deprecated)]
```
TDHolmes, line 20: <3

TDHolmes (approved):

boards/wio_terminal/src/display.rs
bradleyharden: I think that's a question for @jbeaurivage. I didn't make that change.

boards/wio_terminal/src/pins.rs
```diff
@@ -1,160 +1,497 @@
...
-    /// XIN/XOUT
-    pin xin = b22,
-    pin xout = b23,
-
-    /// MISCELLANEOUS
-    pin swo = b30,
-    pin ir_ctl = b31,
-    pin output_ctr_5v = c14,
-    pin output_ctr_3v3 = c15,
+hal::bsp_pins!(
```
bradleyharden, line 11: Note to self. Put this in its own module, to help keep the documentation manageable.

boards/wio_terminal/src/sound.rs
```diff
@@ -1,15 +1,18 @@
+use crate::pins::MicOutput;
 use atsamd_hal::adc::Adc;
 use atsamd_hal::clock::GenericClockController;
-use atsamd_hal::gpio::{self, Floating, Input, Pc30, Pd11, PfB, Port};
+use atsamd_hal::gpio::v2::*;
```
bradleyharden, line 4: Other note to self. Try to get rid of some of these wildcard imports.

boards/wio_terminal/src/wifi.rs
```diff
@@ -1,99 +1,106 @@
...
-            sercom0,
+
+        let pads = uart::Pads::default().rx(pins.miso).tx(pins.mosi);
+        let uart = uart::Config::new(
             mclk,
-            (rx, tx),
-        );
+            sercom0,
+            pads,
+            clocks.sercom2_core(&gclk0).unwrap().freq(),
```
bradleyharden, line 73: Uh, isn't this a bug when considering two lines above?

bradleyharden: By the way, something like this won't be possible with the new clock API.

bradleyharden:
I'll make another pass just to catch a few more things. It would be really nice if someone could test it though. Who has a wio_terminal?

boards/wio_terminal/src/wifi.rs
bradleyharden: In this instance, @jbeaurivage changed the return type, so `?` is not possible. It looks like `.unwrap()` is used everywhere else in this function (presumably for things that won't actually fail), so I don't see the need to return a `Result` just for setting the pins.

boards/wio_terminal/src/display.rs
bradleyharden: This case is less clear. There are already some errors that are passed back to the user, but they are completely erased and converted to `()`, so I don't really know what a user is supposed to do with them other than `panic!`. Seems like someone might want to rethink this approach.

In any event, I think that's a problem for another day. I vote to leave the current `.ok()` approach, because it signals that setting the pin is actually `Infallible`.

bradleyharden:
Other than testing, this should be good to go.

bradleyharden:
Untested YOLO commit. Sorry to future developers

PR #543, opened 2021-11-15
detailed EIC example to XIAO_m0 board

Alexander89:
# Summary

I added a detailed example, how to use the `EIC` with a XIAO board. It would be obvious for senior users, but I think this makes it clear for newbies like me :-) 

Other existing examples don't have this ExtInt9 for Pin 5 issue and don't show how to figure out, which ExtInt was triggered.

BTW: It took me days to figure it out :exploding_head: :sob: 

# Checklist

## If Adding a new Board

### DIFF
--- a/boards/xiao_m0/examples/eic.rs
+++ b/boards/xiao_m0/examples/eic.rs
@@ -8,6 +8,7 @@
 #![no_std]
 #![no_main]
 
+use bsp::{hal, Led0, Led1};
 use core::cell::RefCell;
 use cortex_m::{
     interrupt::{free, Mutex},
@@ -24,7 +25,7 @@ use hal::{
     prelude::*,
 };
 use panic_halt as _;
-use xiao_m0::{hal, Led0, Led1};
+use xiao_m0 as bsp;
 
 static LED_1: Mutex<RefCell<Option<Led1>>> = Mutex::new(RefCell::new(None));
 
--- a/boards/xiao_m0/examples/eic.rs
+++ b/boards/xiao_m0/examples/eic.rs
@@ -1,0 +1,112 @@
+//! It uses an external interrupt to toggle an LED.
+//!
+//! You need to connect a button between Pin a9 and ground. Each time the button
+//! is pressed, the LED will toggle his state.
+//!
+//! To select the correct ExtInt
+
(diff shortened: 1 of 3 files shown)


boards/xiao_m0/examples/eic.rs
```diff
@@ -0,0 +1,112 @@
...
+    eic::{
+        pin::{ExtInt5, Sense},
+        EIC,
+    },
+    gpio::v2::{Pin, PullUpInterrupt},
+    pac::{self, interrupt, CorePeripherals, Peripherals},
+    prelude::*,
+};
+use panic_halt as _;
+use xiao_m0::{hal, Led0, Led1};
```
TDHolmes, line 27: Can you use the pattern we use elsewhere?
```suggestion
use xiao_m0 as bsp;
use bsp::{hal, Led0, Led1};
```

bradleyharden:
Thanks for the contribution. It's nice to have a fully bare metal example like this. At the same time, if I were to implement this, I would have used RTIC to handle the EIC and LED resources. Are you familiar with it? If not, you should check it out. It would save you some `unsafe` and manual management of `static mut` resources.

Also, in the future, feel free to open an issue or join us on Matrix chat if you're stuck on something.

Alexander89:
Hi @bradleyharden 

Yes, I found RTIC. It looks very cool and simplifies a lot of things. 

On the other side, it has a lot of magic :smile: I personally try to start with the minimal abstraction I could handle (or not :seedling: ). My current project is that small, that I _think_ I can handle it but moving forward. RTIC will be the way to go.

 > if you think the mapping of "Analog pin 9" to "Pin PA05 = ExtInt5" was bad, wait til you find out that not all pins PAXY map to ExtIntXY. That ends up being really confusing.
 
 I found this in the docs of my chip. Somebody really likes to confuse poor little developers.

jbeaurivage:
@Alexander89, if you add an entry to the BSP's changelog, I'd be happy to merge.

Alexander89:
sorry, didn't know where I should add it. 

should I open a new section "Version 0.11.1"?

TDHolmes:
@Alexander89 you can put it under a new `# Unreleased` section

PR #546, opened 2021-12-10
fix pwm divisor in get_period

Alexander89:
# Summary

I'm currently trying to use PWM. I'm kind of stuck here but I found this bug in the code.

Regarding the ctrla::PRESCALER_A. 
- DIV1 = 0 
- DIV2 = 1 
- DIV3 = 2

Actually, this code is only working for small prescalers (up to 16). 

```
match params.divider {
    1 => w.prescaler().div1(),
    2 => w.prescaler().div2(),
    4 => w.prescaler().div4(),
    8 => w.prescaler().div8(),
    16 => w.prescaler().div16(),
    64 => w.prescaler().div64(),
    256 => w.prescaler().div256(),
    1024 => w.prescaler().div1024(),
    _ => unreachable!(),
}
```

The only way I saw with my limited experience here, is to match the other way around, (feels a bit hacky).

Any better idea?

---

# Checklist

## If Adding a new Board

### DIFF
--- a/hal/src/thumbv6m/pwm.rs
+++ b/hal/src/thumbv6m/pwm.rs
@@ -93,7 +93,7 @@ impl $TYPE {
         let count = self.tc.count16();
         let divisor = count.ctrla.read().prescaler().bits();
         let top = count.cc[0].read().cc().bits();
-        Hertz(self.clock_freq.0 / divisor as u32 / (top + 1) as u32)
+        Hertz(self.clock_freq.0 / (1u32 << divisor) / (top + 1) as u32)
     }
 }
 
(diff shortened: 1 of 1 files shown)


Alexander89:
btw, the same issue exists at the v7. 
If there is a nice fix, I will add the v7, too.

TDHolmes:
From offline matrix convo:

> maybe impl Into/From<u32> for PRESCALER_A to put the conversion in one place? we have code like this in RTC, timer and PWM code. Definitely could use some cleaning up.
> 
> I don't think it's per chip, but per family. You could probably place it in thumbv6/thumbv7em folders but you'd need to double check the data sheets

bradleyharden:
Pinging @Alexander89, did you want to follow up on this?

Alexander89:
I will follow up, as soon as I'm back at this topic. 

But I will restart this, because the code base enhanced at the meantime.

PR #550, opened 2021-12-15
AES: Advanced Encryption Standard HAL interface

vcchtjader:
# Summary
Adds AES peripheral support including RustCrypto BlockCipher implementations using AES hardware.

Demo-application showing usage of RustCrypto https://github.com/vcchtjader/atsamd-demo/tree/aes

# Checklist


vcchtjader:
There is still an open question regarding the potential risk of multiple simultaneous uses of the AES peripheral.

For now I tried to be explicit in documentation and also provide a "holding area" for AES peripheral without any methods acting on AES, if the AES peripheral is needed at some later stage etc.

Is there some smart way of achieving this?

Sympatron:
I think `unsafe impl !Sync for Aes {}` would be enough to ensure no concurrent access. But I am not entirely sure.

Sympatron:
In your [example ](https://github.com/vcchtjader/atsamd-demo/blob/d478ff8ce57618f73602ae73189d068be9d773c1/src/main.rs#L138-L144)you initialize `AesRustCrypto` like this:
```rust
// Atsamd peripheral init
let aes = Aes::new(device.AES);
let aes_reg = aes.destroy();

// Store AES hardware peripheral in AesRustCrypto to ensure
// no uses of the AES peripheral outside RustCrypto
let aes_rust_crypto = Aes::new_rustcrypto_backend(aes_reg);
```

This is quite unintuitive. I think it would be better to either add `AesRustCrypto`'s impls to `Aes` directly or have as `From` impl to allow `let aes_rust_crypto: AesRustCrypto = aes.into();`.

I don't really get why there is a distinction between `Aes` and `AesRustCrypto`. I don't think they are mutually exclusive.

vcchtjader:
Thank you for looking into this, as you've noted there are some open questions :)

> In your [example ](https://github.com/vcchtjader/atsamd-demo/blob/d478ff8ce57618f73602ae73189d068be9d773c1/src/main.rs#L138-L144)you initialize `AesRustCrypto` like this:
> 
> ```rust
> // Atsamd peripheral init
> let aes = Aes::new(device.AES);
> let aes_reg = aes.destroy();
> ```
^ This is just to test that the `destroy()` method does what is intended, return the AES register.
If somebody would not plan to use any RustCrypto then that would be it, start configuring AES through the regular interface.

> ```rust
> // Store AES hardware peripheral in AesRustCrypto to ensure
> // no uses of the AES peripheral outside RustCrypto
> let aes_rust_crypto = Aes::new_rustcrypto_backend(aes_reg);
> ```
>
This is the big question, the type, `AesRustCrypto` is not needed, but I added it as a "storage"-wrapper for the `AES` register and the difference between `Aes` and `AesRustCrypto` is that `AesRustCrypto` does not expose any of the AES peripheral registers, preventing modifications to AES.

If the user is diligent and simply take care of keeping uses of RustCrypto-backed AES computations and "regular"
AES peripheral use apart then this is completely superfluous.

I weren't able to prevent people from just constructing the `Aes128`/`Aes192`/`Aes256` types conditionally on having `AesRustCrypto`, maybe it is doable somehow? 
 
In order to use AES with RustCrypto there is no need to even deal with the AES peripheral register, only bus clock needs to be enabled and then having the `Aes128`/`Aes192`/`Aes256` type from this crate as a starting point for RustCrypto setup.

> This is quite unintuitive. I think it would be better to either add `AesRustCrypto`'s impls to `Aes` directly or have as `From` impl to allow `let aes_rust_crypto: AesRustCrypto = aes.into();`.
> 
> I don't really get why there is a distinction between `Aes` and `AesRustCrypto`. I don't think they are mutually exclusive.

I agree with the above statement, they are not exclusive, but any use of `encrypt()` or `decrypt()` from RustCrypto using the AES peripheral will reset the peripheral, reconfigure and use it and leave it configured for either encryption or decryption, in Electronic Codebook (ECB) mode for one of the following keysizes: 128/192/256.

Therefore I created this type to try to artificially create some distinction between operation modes. 

If we think documentation informing/warning users of potential pitfalls/race-conditions if using both modes are sufficient, I don't see any reason keeping this `AesRustCrypto` type around

Sympatron:
Good points. Maybe it's clearer if `AesRustCrypto` wraps `Aes` and not `pac::AES` since `AesRustCrypto` a higher lavel wrapper around `Aes`? On the other hand it wouldn't really matter if you could change settings, because `AesRustCrypto` resets everything every time anyway. I not sure.

> I weren't able to prevent people from just constructing the Aes128/Aes192/Aes256 types conditionally on having AesRustCrypto, maybe it is doable somehow?

I just looked deeper into it and the unsafety issue seems to be worse than I thought. Currently you can just create `AesXxx` instances from thin air, send them to different threads and use them concurrently. It wouldn't even help to make them `!Send`.

The `NewBlockCipher::new()` signature is quite unfortunate. The only other embedded project using `cipher` I could find is [`lpc55-hal`](https://crates.io/crates/lpc55-hal) and they didn't implement `NewBlockCipher` at all. I don't have any experience with Rust's crypto libraries so I don't know hoe important `NewBlockCipher` is. But not implementing it and instead adding a constructor that requires `Aes`, `AesRustCrypto` or `AES` and then making `AesXxx` and similar `!Sync` would solve the problem IMO.

vcchtjader:
I moved `Aes` into `AesRustCrypto` as a minimal step for now, I agree this is much cleaner.

This is not a solution to the real problem though, and I tried the general approach lpc55-hal took,
and even though HAL builds just fine and the `BlockCipher` is available neatly gated in my demo application, I can't use [CMAC](https://crates.io/crates/cmac) or [CTR](https://crates.io/crates/ctr) crates which builds on top of `BlockCipher` by requiring `NewBlockCipher`.

I find this really unfortunate :( Maybe there is some way to opt in: "yes let me play with footguns" kind of thing, not sure which way is best though...

Sympatron:
Maybe a scary sounding feature gate to enable `NewBlockCipher`?

Or some kind of singleton that ensures correct usage? For example a mutex you have to provide before calling `NewBlockCipher::new()` that will be stored globally and passed to all `BlockCipher` instances internally.
This wouldn't be the prettiest or most intuitive solution, but at least it would be safer.

Something like this:
```rust
static mut CRYPTO_MUTEX: Option<Mutex> = None;

pub fn set_mutex(mutex: Mutex) {
  unsafe { CRYPTO_MUTEX = Some(mutex) };
}

impl NewBlockCipher for $name {
    fn new(key: &GenericArray<u8, $key_size>) -> Self {
        if { unsafe { CRYPTO_MUTEX.is_none() } } {
          panic!("Call `set_mutex()` before this!");
        }
        ...
    }
}
```

How to implement this mutex in a way that's safe, efficient and doesn't easily cause deadlocks won't be easy though...

bradleyharden:
Where does this stand? I thought I remember it being pretty close to ready

vcchtjader:
> Where does this stand? I thought I remember it being pretty close to ready

Been preempted at work upgrading our application to RTIC 1, but with that done I hope to finish this and get back to #450 :tada: 

Apart from the `NewBlockCipher` issue I would consider it ready for testing

vcchtjader:
Runnable with updated [atsamd-demo](https://github.com/vcchtjader/atsamd-demo) (assuming atsame51 board etc)

```
cargo run --bin aes --features hal-aes
```

Is the feature gate scary enough? Documentation maybe could be even more elaborate, looking forward to hear your thoughts.

vcchtjader:
Anything keeping this from being merged? :)

bradleyharden:
No, I think you've waited long enough. Sorry for the delay. I had hoped to get more eyes on this, but I don't think that will happen.

For my own curiosity, why do you use both `GenericArray` and `const` generics? As far as I could tell, the `GenericArray`s were used inside of a macro, which would seem to lose most of the benefits of `typenum`. At that point, why not just use `const` generics?

vcchtjader:
> For my own curiosity, why do you use both `GenericArray` and `const` generics? As far as I could tell, the `GenericArray`s were used inside of a macro, which would seem to lose most of the benefits of `typenum`. At that point, why not just use `const` generics?

Without going back in there and verifying (*hazy memory access incoming*) I think it were the RustCrypto that relied/used GenericArrays, not sure if that has changed lately though, cipher 0.4 has been released since I noted. 

https://docs.rs/cipher/0.3.0/cipher/trait.NewBlockCipher.html where `type BlockCipherKey<B> = GenericArray<u8, <B as NewBlockCipher>::KeySize>;`

Sympatron:
I tried to make `cipher` use const-generics a while ago. It was not possible, because const-generics are not complete enough on stable currently. That's why they still use `GenericArray`.

The current version of `cipher` does not contain `NewBlockCipher`. I don't how that's supposed to be handled now. Depending on how they replaced it, it might actually solve the problem we discussed above.

PR #555, opened 2021-12-24
Add SAMD20J18A support

didjcodt:
# Summary

This Pull Request integrates the SAMD20J18A microcontroller:
- It adds the SVD file and its XSL patches in the svd directory
- It generates the corresponding pac
- It updates the HAL to have support for it
- It adds the SAMD20 Xplained Pro board in the BSP with a bliny_basic example

This is only a proof of concept, not all features have been tested in real life even though everything compiles fine :)

# Checklist

## If Adding a new Board


boards/samd20_xplained_pro/.cargo/config
```diff
@@ -0,0 +1,18 @@
...
+runner = 'arm-none-eabi-gdb -q -x openocd.gdb'
+#runner = 'probe-run --chip ATSAMD21G18A'
+
+rustflags = [
+
+  # This is needed if your flash or ram addresses are not aligned to 0x10000 in memory.x
+  # See https://github.com/rust-embedded/cortex-m-quickstart/pull/95
+  # "-C", "link-arg=--nmagic",
+
+  "-C", "linker=lld",
```
TDHolmes, line 16: Don't include this, and uncomment the nmagic line above to align with the rest of the configs. Also nit pick, can you name it config.toml? you don't need to leave off the .toml anymore.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -271,8 +290,16 @@ impl<S: Sercom> Registers<S> {
         use BaudMode::*;
         use Oversampling::*;
 
+        #[cfg(feature = "samd20")]
```
TDHolmes, line 293: since this is so littered with `#[cfg]`s, I would consider separate functions. Or maybe config blocks, like
```rust
#[cfg(feature = "samd20")]
{
    // all of the samd20 logic
}

#[cfg(feature = "samd21")]
{
    // all of the samd21 logic
}
```

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -409,12 +438,18 @@ impl<S: Sercom> Registers<S> {
         // Enable RX
         if rxen {
             usart.ctrlb.modify(|_, w| w.rxen().set_bit());
+            #[cfg(feature = "samd20")]
+            while usart.status.read().syncbusy().bit_is_set() {}
+            #[cfg(not(feature = "samd20"))]
             while usart.syncbusy.read().ctrlb().bit_is_set() {}
```
TDHolmes, line 444: same sync busy comment as the other peripheral

hal/src/thumbv6m/clock.rs
```diff
@@ -180,23 +180,35 @@ impl GenericClockController {
         }
 
         // Feed 32khz into the DFLL48
-        state.enable_clock_generator(DFLL48, GCLK1);
+        #[cfg(not(feature = "samd20"))]
+        state.enable_clock_generator(ClockId::DFLL48, GCLK1);
```
TDHolmes, line 184: why `DFLL48` here but `DFLL48M in `state.set_gclk_divider_and_source`?

hal/src/thumbv6m/clock.rs
```diff
@@ -321,7 +338,7 @@ impl GenericClockController {
             XOSC32K | OSC32K | OSCULP32K => OSC32K_FREQ,
             GCLKGEN1 => self.gclks[1],
             OSC8M => OSC8M_FREQ,
-            DFLL48M => OSC48M_FREQ,
+            ClockSource::DFLL48M => OSC48M_FREQ,
```
TDHolmes, line 341: why did you add this prefix?

hal/src/thumbv6m/sercom/v1/i2c.rs
```diff
@@ -89,13 +89,21 @@ where
             // reset the sercom instance
             sercom.i2cm().ctrla.modify(|_, w| w.swrst().set_bit());
             // wait for reset to complete
+            #[cfg(feature = "samd20")]
+            while sercom.i2cm().status.read().syncbusy().bit_is_set()
+                || sercom.i2cm().ctrla.read().swrst().bit_is_set()
+            {}
```
TDHolmes, line 95: same comment as past peripherals around syncbusy and such

hal/src/thumbv6m/sercom/v1/spi.rs
```diff
@@ -168,13 +168,21 @@ macro_rules! spi_master {
                 // reset the sercom instance
                 sercom.spi().ctrla.modify(|_, w| w.swrst().set_bit());
                 // wait for reset to complete
+                #[cfg(feature = "samd20")]
+                while sercom.spi().status.read().syncbusy().bit_is_set()
+                    || sercom.spi().ctrla.read().swrst().bit_is_set()
+                {}
```
TDHolmes, line 174: same syncbusy peripheral comment

pac/atsamd20j/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "atsamd20j"
+description = "Peripheral access API for ATSAMD20J microcontrollers (generated using svd2rust)"
+version = "0.11.0"
```
TDHolmes, line 4: maybe 0.1.0

hal/src/dmac/channel/mod.rs
```diff
@@ -8,7 +8,7 @@
 //! requires setting a priority level, as well as enabling or disabling
 //! interrupt requests (only for the specific channel being initialized).
 #![cfg_attr(
-    not(any(feature = "samd11", feature = "samd21")),
+    not(any(feature = "samd11", feature = "samd20", feature = "samd21")),
```
Sympatron, line 11: I would prefer `feature = "min-samd51g"` in this case.

hal/src/dmac/channel/mod.rs
```diff
@@ -145,7 +145,7 @@ impl<Id: ChId, S: Status> Channel<Id, S> {
         // Software reset the channel for good measure
         self._reset_private();
 
-        #[cfg(any(feature = "samd11", feature = "samd21"))]
+        #[cfg(any(feature = "samd11", feature = "samd20", feature = "samd21"))]
```
Sympatron, line 148: To better match the `#[cfg(feature = "min-samd51g")]` below, I would prefer `#[cfg(not(feature = "min-samd51g"))]` here.

hal/src/dmac/channel/mod.rs
```diff
@@ -253,7 +253,7 @@ impl<Id: ChId> Channel<Id, Ready> {
         // SAFETY: This is actually safe because we are writing the correct enum value
         // (imported from the PAC) into the register
         unsafe {
-            #[cfg(any(feature = "samd11", feature = "samd21"))]
+            #[cfg(any(feature = "samd11", feature = "samd20", feature = "samd21"))]
```
Sympatron, line 256: Same here.

Sympatron:
I'm not that familiar with SAMD20. But I think it's mostly SAMD21 without USB. So it's probably useful to have a common feature for them (`samd2x`?) to simplify the `cfg`s.

boards/samd20_xplained_pro/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+repository = "https://github.com/atsamd-rs/atsamd"
+readme = "README.md"
+edition = "2018"
+
+[dependencies.cortex-m-rt]
+version = "0.7"
+optional = true
+
+[dependencies.atsamd-hal]
+path = "../../hal"
```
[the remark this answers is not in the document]
didjcodt: If I remove the path dependency, it will use the published crate which does not contains the current updates, how should I replace it?
Btw why does including this path makes it a Tier 1 dependency?

hal/src/thumbv6m/clock.rs
didjcodt: The svd file defines it that way. Maybe I can try adding a replacement rule in the xslt, I'll check that

hal/src/thumbv6m/clock.rs
didjcodt: In the SAMD20, the PAC has `pac::gclk::clkctrl::ID_A::DFLL48M` and `pac::gclk::genctrl::SRC_A::DFLL48M`. As ID_A and SRC_A are both glob-imported, that was making a ambiguous reference error

boards/samd20_xplained_pro/.cargo/config
didjcodt: Removed and renamed :)

boards/samd20_xplained_pro/.cargo/config
```diff
@@ -0,0 +1,18 @@
+# samd21 is a Cortex-M0 and thus thumbv6m
+
+[build]
+target = "thumbv6m-none-eabi"
+
+[target.thumbv6m-none-eabi]
+runner = 'arm-none-eabi-gdb -q -x openocd.gdb'
+#runner = 'probe-run --chip ATSAMD21G18A'
```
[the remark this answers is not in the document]
didjcodt: Done

boards/samd20_xplained_pro/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "samd20_xplained_pro"
+version = "0.9.0"
```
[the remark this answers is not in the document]
didjcodt: Updated :)

hal/Cargo.toml
```diff
@@ -17,7 +17,7 @@ edition = "2021"
 rust-version = "1.56"
 
 [package.metadata.docs.rs]
-features = ["samd21g", "samd21g-rt", "unproven", "usb"]
+features = ["device", "samd20", "samd20j-rt", "samd21g", "samd21g-rt", "unproven", "usb"]
```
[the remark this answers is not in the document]
didjcodt: No problem, I was struggling to get the documentation but I finally got it from the board crate :)
Done

pac/atsamd20j/Cargo.toml
didjcodt: Done

hal/src/dmac/channel/mod.rs
didjcodt: Perfect, I didn't know that was equivalent :)
Updated

hal/src/dmac/channel/mod.rs
didjcodt: Same, updated

hal/src/dmac/channel/mod.rs
didjcodt: Updated too

didjcodt:
> I'm not that familiar with SAMD20. But I think it's mostly SAMD21 without USB. So it's probably useful to have a common feature for them (`samd2x`?) to simplify the `#[cfg]`s.

Yes, good idea! I'll try to make a `samd2x` feature which is activated from any of the samd20/samd21 features. I never wrote anything with features yet so that's going to be something new to learn :)

hal/src/sercom/v2/pad/impl_pad_thumbv6m.rs
```diff
@@ -86,11 +86,11 @@ macro_rules! pad_table {
 }
 
 pad_table!(
-    #[cfg(feature = "samd21")]
+    #[cfg(any(feature = "samd20", feature = "samd21"))]
```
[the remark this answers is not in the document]
didjcodt: I tried making one, pushed the new version. What do you think about it? I was thinking maybe I should do the same for the min-samd*

hal/src/thumbv6m/clock.rs
didjcodt: I just made a replacement in the xslt so the naming is now identical :)

hal/src/sercom/v2/uart/reg.rs
didjcodt: Oh, perfect, I thank putting them in blocks would have made the variables definitions in another scope. That's fixed :)

boards/samd20_xplained_pro/.cargo/config.toml
```diff
@@ -0,0 +1,15 @@
+[build]
+target = "thumbv6m-none-eabi"
+
+[target.thumbv6m-none-eabi]
+runner = 'arm-none-eabi-gdb -q -x openocd.gdb'
+#runner = 'probe-run --chip ATSAMD21G18A'
```
TDHolmes, line 6: nit: even this is commented out, it's the wrong chip name

boards/samd20_xplained_pro/Cargo.toml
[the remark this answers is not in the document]
TDHolmes: ah, good point... I think we'll need to leave it as a path dependency for now since it's new.

Tier one BSPs use path dependencies on the HAL, which means for any HAL change affecting BSPs, all Tier one BSPs must be upgraded in the same PR for CI to pass. Tier two BSPs only rely on released versions of the HAL/PACs, so they can be locked to an older version and updated if someone is interested in doing so. This lowers the burden on maintainers as well as folks looking to change the HAL. See [the README](https://github.com/atsamd-rs/atsamd#how-to-use-a-bsp-ie-getting-started-writing-your-own-code) for more details.

boards/samd20_xplained_pro/SAMD20.yaml
```diff
@@ -0,0 +1,1329 @@
+---
```
TDHolmes, line 1: what is this file for?

hal/src/gpio/v1.rs
```diff
@@ -643,45 +643,45 @@ port!([
     (PA25, Pa25),
     #[cfg(not(feature = "samd11"))]
     (PA27, Pa27),
-    #[cfg(any(feature = "samd11", feature = "samd21"))]
+    #[cfg(any(feature = "samd11", feature = "samd2x"))]
     (PA28, Pa28),
     (PA30, Pa30),
     (PA31, Pa31),
-    #[cfg(any(feature = "min-samd21j", feature = "min-samd51j"))]
+    #[cfg(any(feature = "min-samd2x", feature = "min-samd51g"))]
```
TDHolmes, line 650: Are you sure about all of these `samd51j` -> `samd51g` changes?

hal/src/sercom/v2.rs
```diff
@@ -40,6 +40,14 @@ pub mod uart;
 #[cfg(feature = "dma")]
 pub mod dma;
 
+/// Register synchronization flags
+pub enum reg_sync {
```
TDHolmes, line 44: I think enums are usually camel case `RegSync`

hal/src/spi_common.rs
```diff
@@ -26,18 +26,27 @@ pub trait CommonSpi {
     /// Helper for accessing the spi member of the sercom instance
     fn spi_mut(&mut self) -> &SPI;
 
+    /// Wait for the registers synchronization
+    #[inline]
+    fn is_register_synced(&mut self) -> bool {
```
TDHolmes, line 31: Since this is just a read, I don't think this needs to take a mutable ref

hal/src/spi_common.rs
```diff
@@ -26,18 +26,27 @@ pub trait CommonSpi {
...
+        #[cfg(feature = "samd20")]
+        return !self.spi().status.read().syncbusy().bit_is_set();
+        #[cfg(not(feature = "samd20"))]
+        !self.spi().syncbusy.read().enable().bit_is_set()
```
TDHolmes, line 35: nit: I'd prefer either both using `return` or both not just for consistency

hal/src/thumbv6m/sercom/v1/i2c.rs
```diff
@@ -19,6 +19,12 @@ const BUS_STATE_BUSY: u8 = 3;
 const MASTER_ACT_READ: u8 = 2;
 const MASTER_ACT_STOP: u8 = 3;
 
+enum i2c_reg_sync {
```
TDHolmes, line 22: same camel case comment here

hal/src/thumbv6m/sercom/v1/i2c.rs
```diff
@@ -85,38 +91,53 @@ where
...
+            while !s.is_register_synced(i2c_reg_sync::SysOp) {}
         }
 
-        Self { sda, scl, sercom }
+        s
+    }
+
+    /// Wait for the registers synchronization
+    #[inline]
+    fn is_register_synced(&mut self, synced_reg: i2c_reg_sync) -> bool {
```
TDHolmes, line 132: not sure `&mut self` is needed for just a read

hal/src/thumbv6m/sercom/v1/spi.rs
TDHolmes: I think you missed SPI in your reg sync cleanup

hal/src/thumbv6m/sercom/v1/uart.rs
```diff
@@ -253,10 +259,17 @@ macro_rules! uart {
...
+                            w.txpo().bit(txpo != 0);
+                            #[cfg(not(feature = "samd20"))]
                             w.txpo().bits(txpo);
 
                             w.form().bits(0x00);
-                            w.sampr().bits(0x00); // 16x oversample fractional
+
+                            // 16x oversample fractional
+                            #[cfg(not(feature = "samd20"))]
+                            w.sampr().bits(0x00);
```
TDHolmes, line 271: does SAMD20 not have oversampling setting? If it does and it's just different, make sure to still set it

TDHolmes (commented):
Looking better! the `samd2x` feature helps a lot I think

hal/src/sercom/v2/pad/impl_pad_thumbv6m.rs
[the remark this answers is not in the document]
TDHolmes: love it

hal/src/thumbv6m/sercom/v1/spi.rs
didjcodt: Yes, sorry I just pushed my work of tonight and I will continue tomorrow, there are still a few things I need to clean :)

hal/src/gpio/v1.rs
bradleyharden: I don't think these changes are correct. I was pretty confident about these attributes when I wrote the module.

hal/src/sercom/v1/pads.rs
```diff
@@ -125,9 +125,9 @@ macro_rules! pad_alias {
 }
 
 pad_alias!(Sercom0, Sercom1);
-#[cfg(any(feature = "samd21", feature = "min-samd51g"))]
+#[cfg(any(feature = "samd2x", feature = "min-samd51g"))]
 pad_alias!(Sercom2, Sercom3);
-#[cfg(any(feature = "min-samd21g", feature = "min-samd51g"))]
+#[cfg(any(feature = "min-samd2x", feature = "min-samd51g"))]
```
bradleyharden, line 130: What's the difference between `min-samd2x` and `samd2x`? That seems redundant. `min-samd21g` is intended to exclude `samd21e`.

hal/src/sercom/v2/spi/reg.rs
```diff
@@ -71,6 +71,20 @@ impl<S: Sercom> Registers<S> {
...
+        return !self.spi().status.read().syncbusy().bit_is_set();
+        #[cfg(not(feature = "samd20"))]
+        match synced_reg {
+            reg_sync::CtrlB => !self.spi().syncbusy.read().ctrlb().bit_is_set(),
+            reg_sync::Length => !self.spi().syncbusy.read().length().bit_is_set(),
+            reg_sync::Enable => !self.spi().syncbusy.read().enable().bit_is_set(),
+            reg_sync::SwRst => !self.spi().syncbusy.read().swrst().bit_is_set(),
+        }
+    }
+
```
bradleyharden, line 87: I would like to look through the SAMD20 datasheet myself before we move forward with this.

hal/src/sercom/v2/spi/reg.rs
```diff
@@ -80,12 +83,16 @@ impl<S: Sercom> Registers<S> {
     #[inline]
     pub fn set_op_mode(&mut self, mode: MODE_A, mssen: bool) -> () {
         self.spi().ctrla.modify(|_, w| w.mode().variant(mode));
+        #[cfg(not(feature = "samd20"))]
         self.spi().ctrlb.modify(|_, w| w.mssen().bit(mssen));
```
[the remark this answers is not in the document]
bradleyharden: Strangely, `MSSEN` can be found in the datasheet, but only in the text, not in the register maps. I don't know what's going on there.

hal/src/sercom/v2/uart/reg.rs
```diff
@@ -74,8 +89,11 @@ impl<S: Sercom> Registers<S> {
     #[inline]
     pub(super) fn configure_pads(&mut self, rxpo: u8, txpo: u8) {
         self.usart().ctrla.modify(|_, w| unsafe {
-            w.rxpo().bits(rxpo);
-            w.txpo().bits(txpo)
+            #[cfg(feature = "samd20")]
+            w.txpo().bit(txpo != 0);
+            #[cfg(not(feature = "samd20"))]
+            w.txpo().bits(txpo);
+            w.rxpo().bits(rxpo)
```
bradleyharden, line 96: We should review the differences in SERCOM carefully. @jbeaurivage, have you looked at this yet?

hal/src/sercom/v2.rs
bradleyharden: I would like to look at this and verify it's the best approach to integrating with the rest of the `sercom` code.

boards/samd20_xplained_pro/Cargo.toml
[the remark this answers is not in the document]
didjcodt: Ok, I understand now. Maybe we can remove the board implementation from this PR, and I open a new PR with it after this one is merged? (If yes, I would probably keep it until the end of the PR for practical reasons and remove it at the end :))

hal/src/sercom/v2/uart/reg.rs
jbeaurivage: I haven't yet, but I'd definitely like to comb through this finely with the datasheet next to me. Annoyingly enough, it seems the SAMD20 isn't exactly 'just a SAMD21 minus the USB'.

hal/src/sercom/v2/spi/reg.rs
jbeaurivage: I would like to propose an alternate approach. Instead of declaring a function that takes an enum, we could do something like:
``` rust
#[cfg(not(feature = "samd20"))]
#[inline]
fn sync_ctrlb(&self) {
    while self.spi().syncbusy.read().ctrlb().bit_is_set() {}
}

#[cfg(feature = "samd20")]
#[inline]
fn sync_ctrlb(&self) {
    while self.spi().status.read().syncbusy().bit_is_set() {}
}
```

This has the advantage of not having any overhead of passing enums around, and reduces chances of making mistakes. At any rate, a shared reference to `self` is sufficient.

Also, `SYNCBUSY.LENGTH` seems to be exclusive to SAMx5x targets.

hal/src/sercom/v2/uart/config.rs
```diff
@@ -249,12 +250,14 @@ where
...
     /// corresponding flag in the STATUS register.
     #[inline]
+    #[cfg(not(feature = "samd20"))]
     pub fn set_collision_detection(&mut self, enabled: bool) {
         self.registers.set_collision_detection(enabled);
     }
 
     /// Get the current collision detector setting
     #[inline]
+    #[cfg(not(feature = "samd20"))]
```
jbeaurivage, line 260: Maybe it would be a good idea to separate out all the functions unavailable to the SAMD20 out in their own `impl` block to avoid having to repeat the `#[cfg]s` at every function definition.

hal/src/sercom/v2/uart/pads_thumbv6m.rs
```diff
@@ -264,7 +264,7 @@ where
     }
 }
 
-#[cfg(feature = "samd21")]
+#[cfg(feature = "samd2x")]
```
jbeaurivage, line 267: RTS/CTS pads don't exist for SAMD20. I think this chip should get special cases for the implementation of `ValidPads`, and disable the current implementations, so that `RTS` and `CTS` are both `NoneT`. This would be simpler that changing the `impl_rxpotxpo` macro, though it could be confusing the the end user when creating a new `Pads` type.

There are also slight changes int `USART.CTRLA.TXPO` field. We need to make sure we never write `0x2` to this field for SAMD20 targets.

 In any case, whatever differences between the SAMD20 and the other chips definitely need to be documented in the relevant module-level documentations.

Paging @bradleyharden - What do you think of all this?

hal/src/sercom/v2/uart/reg.rs
jbeaurivage: @didjcodt, can you please explain what's the reasoning here? This kind of thing should probably be handled at the `impl_rxpotxpo` level rather than inside the functions.

hal/src/sercom/v2/uart/reg.rs
jbeaurivage: As I said in an earlier comment, it would be better to put all the SAMD20-specific functions in their own `impl` block, as well as all the functions disabled on SAMD20 in their own impl. That means you would only need two `#[cfg]s` and it would make the whole thing much more readable :) There would be two separate `set_baud` and `get_baud` functions in this case.

hal/src/thumbv6m/adc.rs
```diff
@@ -201,6 +201,30 @@ adc_pins! {
...
+    PB02: 10,
+    PB03: 11,
+    PB04: 12,
+    PB05: 13,
+    PB06: 14,
+    PB07: 15,
+    PB08: 2,
+    PB09: 3
+}
+
```
jbeaurivage, line 227: Consider separating this out in common/SAMD20G/SAMD20J blocks, like for the SAMD21. Or at least,
```diff
-#[cfg(feature = "samd20")]
+#[cfg(feature = "samd20j")]
```

So someone doesn't accidentally use a pin that doesn't exist on, say, an SAMD20E.

hal/src/thumbv6m/eic/pin.rs
```diff
@@ -237,143 +237,143 @@ ei!(ExtInt[7] {
...
-#[cfg(feature = "samd21")]
+#[cfg(feature = "samd2x")]
 ei!(ExtInt[0] {
     Pa0,
     Pa16,
-    #[cfg(feature = "min-samd21j")]
+    #[cfg(feature = "min-samd2x")]
     Pb0,
-    #[cfg(feature = "min-samd21j")]
+    #[cfg(feature = "min-samd2x")]
```
jbeaurivage, line 246: The `min-samd21j` -> `min-samd2x` changes will need to be reviewed.

hal/src/thumbv6m/pwm.rs
```diff
@@ -1,15 +1,21 @@
...
-use crate::pac::{PM, TCC0};
+use crate::pac::PM;
+#[cfg(not(feature = "samd20"))]
+use crate::pac::TCC0;
 #[cfg(feature = "samd11")]
 use crate::pac::{TC1, TC2};
-#[cfg(feature = "samd21")]
-use crate::pac::{TC3, TC4, TC5, TCC1, TCC2};
-#[cfg(feature = "samd21j")]
+#[cfg(feature = "samd2x")]
```
jbeaurivage, line 13: Does this mean that SAMD20s don't have TC1 and TC2?

hal/src/thumbv6m/sercom/v1/i2c.rs
jbeaurivage: Same suggestion as for the SPI and UART modules: consider using the approach that doesn't require maintaining a bespoke enum.

hal/src/thumbv6m/sercom/v1/i2c.rs
```diff
@@ -136,6 +157,11 @@ where
 
         // Signal start and transmit encoded address.
         unsafe {
+            #[cfg(feature = "samd20")]
+            self.i2cm()
+                .addr
+                .write(|w| w.addr().bits(addr));
+            #[cfg(not(feature = "samd20"))]
```
jbeaurivage, line 164: Why is this different?

hal/src/thumbv6m/sercom/v1/i2c.rs
```diff
@@ -158,6 +184,12 @@ where
...
         }
+
+        #[cfg(feature = "samd20")]
+        if status.lowtout().bit_is_set() {
+            return Err(I2CError::Timeout);
+        }
+        #[cfg(not(feature = "samd20"))]
         if status.lowtout().bit_is_set() || status.sexttout().bit_is_set()
             || status.mexttout().bit_is_set()
         {
```
jbeaurivage, line 195: You might want to break this up into separate SAMD20/non-SAMD20 functions to avoid so many `#[cfg]`s, or at least into config blocks.

hal/src/thumbv6m/sercom/v1/i2c.rs
```diff
@@ -379,7 +424,7 @@ i2c!([
         ),
 ]);
 
-#[cfg(feature = "min-samd21g")]
+#[cfg(feature = "min-samd2x")]
```
jbeaurivage, line 427: `min-samd21g` -> `min-samd2x`

boards/samd20_xplained_pro/SAMD20.yaml
didjcodt: This is the chip description file to include when using cargo flash using `--chip-description-path` argument

boards/samd20_xplained_pro/Cargo.toml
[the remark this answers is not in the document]
ianrrees: FWIW, to facilitate development, one option is to use a [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) in Cargo.toml .  The idea would be to PR the changes that you want to eventually be published, but don't include the patch section which just refers to your local copy of atsamd-rs instead of the one on crates.io .

That said, it probably still makes sense to publish the PAC and HAL changes, before a separate PR to add the new boards.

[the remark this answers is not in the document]
ianrrees: FWIW, to facilitate development, one option is to use a [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) in Cargo.toml .  The idea would be to PR the changes that you want to eventually be published, but don't include the patch section which just refers to your local copy of atsamd-rs instead of the one on crates.io .

That said, it probably still makes sense to publish the PAC and HAL changes, before a separate PR to add the new boards.

Magkoooh (approved):

PR #559, opened 2021-12-28
General update itsybitsy m0

javiser:
# Summary
The goal of this PR is to bring the ItsyBitsy M0 (which is a Tier 2 BSP) to the same "maturity" level of a Tier 1 BSP. Instead of just bumping a couple of dependencies as done with other boards (i.e. for the ItsyBitsy M4 in #547, which by the way was quite useful as reference for this PR), my proposal is to take advantage of the fact that the ItsyBitsy M0 is internally pretty similar to the Feather M0, which is Tier 1. So I copied the current code of the Feather M0 and made the changes necessary to reflect the HW differences. I also copied the examples from the Feather M0 and adjusted them as necessary, and kept the existing examples where necessary or meaningful. By doing do I hope that maintaining this board as Tier 2 BSP will much easier, as keeping an eye on the Feather M0 and implementing deltas will be much easier. The main and relevant differences between the boards are:
- ItsyBitsy M0 does not have an external crystal clock, only internal
- ItsyBitsy M0 has a built-in DOTStar LED, not a neopixel (which is only available in the Feather M0 Express variant)
- ItsyBitsy M0 has a built-in SPI flash as the Feather M0 Express variant, but on other internal pins
- There are not any variantes on the ItsyBitsy M0 (i.e. RMF or Adalogger)
- Different number of pins and partly different capabilities (see comments in pins structure for details)

What has been done in this PR:
- Copied feather_m0 as itsybitsy_m0 as starting point
- Adjusted pins, renamed, removed and added. Also adding hints and documentation from the official Adafruit learn guide (https://learn.adafruit.com/introducing-itsy-bitsy-m0/pinouts)
- Adjusted existing examples to work, or in some cases replaced to be equivalent to the feather m0 examples
- Dotstar example renamed to dotstar_rainbow and added a new dotstar_flashing similar to the one from itsybitsy_m4 board
- Adjusted Cargo.toml correspondingly

:bulb: Hint: It might be useful to compare the contents of the `feather_m0` folder with this new version of `itsybitsy_m0` as well, as this might show the changes of this PR better than the before-after comparison of the `itsybitsy_m0` files.

What should be done yet:
  - [x] ~~Somebody help me test all ssd1306 examples with a real display~~
  - [x] ~~Somebody help me test the new `adc.rs`, `clock.rs`, `dmac.rs` and `uart.rs` examples copied from `feather_m0` (even if I am pretty sure that they should work just as fine as for the Feather M0)~~

# Checklist


bradleyharden:
If this is mostly just a copy and rework of `feather_m0`, then I'm good with it. We just need to test it on hardware to be sure there are no lingering bugs.

javiser:
> If this is mostly just a copy and rework of `feather_m0`, then I'm good with it. We just need to test it on hardware to be sure there are no lingering bugs.

That's right @bradleyharden, this is a copy and rework of `feather_m0`. I tested many of the examples, but for some I was missing either the know-how or the equipment, that's why I left two open tasks in this PR. I do know that the SPI is working because I got a display with a ST7789 to work, as well as a rotary encoder and some RGB LEDs on other pins. So I definitely would appreciate some help here to confirm that I did not break anything here.

boards/itsybitsy_m0/Cargo.toml
```diff
@@ -6,43 +6,70 @@ description = "Board Support crate for the Adafruit ItsyBitsy M0"
 keywords = ["no-std", "arm", "cortex-m", "embedded-hal"]
 categories = ["embedded", "hardware-support", "no-std"]
 license = "MIT OR Apache-2.0"
-edition = "2021"
 repository = "https://github.com/atsamd-rs/atsamd"
 readme = "README.md"
+edition = "2021"
+resolver = "2"
```
jbeaurivage, line 12: This is probably a nit, but `edition = "2021"` implies `resolver = "2" `, so it's not actually needed here.

boards/itsybitsy_m0/Cargo.toml
```diff
@@ -51,19 +78,57 @@ chip = "ATSAMD21G18A"
...
 [[example]]
-name = "spi_ssd1306"
+name = "ssd1306_terminalmode_128x64_spi"
 
 [[example]]
 name = "usb_echo"
 required-features = ["usb"]
 
 [[example]]
 name = "twitching_usb_mouse"
```
jbeaurivage, line 106: Does this example need the `usb` feature?

boards/itsybitsy_m0/Cargo.toml
javiser: I was not aware of that thanks. This PR has been lying around for a while now and it is not aligned with the current master anymore, also mostly because I switched from rust/itsibitsy_m0 to C++/ESP32-C3 for my project (blame on me). I am also not sure anymore how well accepted my proposal was and if somebody is interested at all yet in this change. If not, I don't take it personally and I can close this PR (I had forgotten about this actually...). If yes, I can try to realign it again, but then it needs to be merged before the differences get too big again.

javiser: Thanks! I will see if I find some time in the couple of days to get it done.

jbeaurivage:
@javiser, I just confirmed that the pin/sercom definitions are correct to the hardware.

boards/itsybitsy_m0/Cargo.toml
javiser: which one do you mean?

boards/itsybitsy_m0/Cargo.toml
javiser: I removed that `resolver = "2"` line. But it is there because it is in the feather_m0 code as well, from which it is copied ;)

javiser:
Ok, @jbeaurivage, I finally found some time and resynchronized everything with the current master. I still would like to hook up the board and do a couple of basic tests to see if the things are working yet fine, but in the mean time, if you are interested, you can finish the review. We will also need to decide what to do with the above open tasks. Either somebody helps me there with that or we need to "merge at risk and hope for the best".

boards/itsybitsy_m0/Cargo.toml
jbeaurivage: The `twitching_usb_mouse` example. Sorry for the confusion!

boards/itsybitsy_m0/Cargo.toml
jbeaurivage: Heh thanks, like I said it's really a nit but thanks for cleaning it up nonetheless!

jbeaurivage:
Thanks @javiser.

I'm pretty confident the `adc`, `dmac`, `clocks` and `uart` examples work as is, since they're copied from `feather_m0`. The `clock` and `dmac` examples are purely "internal" (as in they don't depend on external comms or peripherals apart from USB). 

I think you'll have a hard time finding someone here with both an itsybitsy m0 and an ssd1306 to test.

Overall I'm comfortable merging whenever you complete your tests.

boards/itsybitsy_m0/Cargo.toml
javiser: Ok thanks, done in fd8f90e08942185cd4f4f6f1f50cf9ac5e5dddc2

javiser:
So I finally managed to flash some of the tests, but none of the ones with external periphery and I could not test any of the tests I requested help for anyways. Based on your previous comment, I have decided to remove the corresponding tasks. I think that this PR can be merged now, it is up to you now @jbeaurivage . Thanks again for your help!

javiser:
Thanks!

PR #561, opened 2022-01-04
I2C v2 API

jbeaurivage:
# Summary
In anticipation  of removing the `sercom::v1` module, here's a `sercom::v2` implementation of the I2C module.

Currently, only master mode is supported. Also, given that DMA transfers are set-and-forget, I don't think a Future-like API is necessary.

# To Do

- [x] Implement support for DMA transfers
- [x] Fix the `Pads` API for SAMD11
- [x] Verify `pads_thumbv7em` is correct

# Checklist
  - [x] No new warnings or clippy suggestions have been introduced (see CI or check locally) -- I fixed some documentation in other sercom modules.


jbeaurivage:
@bradleyharden, if you don't mind, could you do a quick sanity check, especially for the typelevel stuff? I think `sercom::v2::i2c::pads` should be correct, but I'd like a second pair of eyes on it.

hal/src/sercom/v2/i2c.rs
```diff
@@ -0,0 +1,137 @@
...
+    Us105 = 0x2,
+    /// 20-21 SCL cycles (200-210 us @ 100 kHz)
+    Us205 = 0x3,
+}
+
+pub struct I2c<C: AnyConfig> {
+    config: C,
+}
+
+impl<C: AnyConfig> I2c<C> {
```
bradleyharden, line 43: Do you want some way to `unsafe`ly write the DATA register directly? It also might help to expose a way to get a pointer to the DATA register, for setting up DMA with `samd-dma` rather than the `dmac` module.

hal/src/sercom/v2/i2c.rs
```diff
@@ -0,0 +1,137 @@
...
+mod reg;
+use reg::Registers;
+
+mod flags;
+pub use flags::*;
+
+mod config;
+pub use config::*;
+
+pub mod impl_ehal;
```
bradleyharden, line 15: If you make this module `pub`, then I would add documentation there on the various `embedded-hal` traits and their implementations. Otherwise, make it private.

boards/feather_m0/examples/i2c.rs
```diff
@@ -0,0 +1,68 @@
...
+    let channels = dmac.split();
+
+    let chan0 = channels.0.init(PriorityLevel::LVL0);
+    let chan1 = channels.1.init(PriorityLevel::LVL0);
+
+    type Pads = i2c::Pads<Sercom3, bsp::Sda, bsp::Scl>;
+
+    let gclk0 = clocks.gclk0();
+    let sercom3_clock = &clocks.sercom3_core(&gclk0).unwrap();
+    let pads: Pads = i2c::Pads::new(sda, scl);
```
bradleyharden, line 54: With the suggested changes to the `pads` module, I don't think you'll need this type annotation.

hal/src/sercom/v2/i2c/pads.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use crate::sercom::v2::*;
+use crate::typelevel::{NoneT, Sealed};
+use core::marker::PhantomData;
+
+pub struct Pads<S, SDA, SCL>
+where
+    S: Sercom,
+    SDA: IsPad<PadNum = Pad0, Sercom = S> + IsI2cPad,
+    SCL: IsPad<PadNum = Pad1, Sercom = S> + IsI2cPad,
```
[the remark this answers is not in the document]
bradleyharden: Actually, I think there might be some room for improvement of the existing structure. I think both `IsPad` and `GetPad` could have the `IoSet` as an associated type. I might not have realized that before. I'll play with it. That would clean up a few trait bounds.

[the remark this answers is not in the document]
bradleyharden: Had another realization just now. For I2C on `thumbv7em,` `Sercom` and `IoSet` is enough information to specify the `Pads`. That's not true for SPI, though, because you can mix and match MISO and MOSI to different `PadNum`s.

I think you'll need a different implementation anyway, so maybe it's worth trying to implement that. I think it would require a new trait though, to look up `PinId` by `Sercom` and `IoSet`.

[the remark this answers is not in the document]
bradleyharden: Ah, now I remember. `IoSet` can't be an associated type of `IsPad`, because the same `PinId` and `PinMode` can map to *multiple* `IoSet`s for the same `Sercom`.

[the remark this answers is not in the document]
bradleyharden: So I think it would be possible to implement `i2s::Pads<S, I> where S: Sercom, I: IoSet`. Knowing the `Sercom` and `IoSet` is definitely enough information, because you know to use `Pad0` and `Pad1`. *However*, I think we may not want to actually do that.

Number one, it would be an outlier among the different `Pads` structs, so it could add confusion. But secondly, I think it might make type inference worse. Right now, the compiler is able to infer **all** of the types when creating `uart::Pads` or `spi::Pads` directly in `main`. It can derive the `Sercom` type from the PAC `SERCOM` type. It can get the `PinId`s from the arguments to `.mosi()`, etc. It then looks up the correct `PinMode`. And finally, it can compare all of those to discover that only **one** `IoSet` type is common across all the pins.

If we switch to `Pads<S, I>`, the user **must** specify the `IoSet`, even in `main`, because it can never be inferred. At least it can't without analyzing multiple `PinId` types simultaneously.

It would also add additional type-level complexity to something that is already fairly complex. I think it's probably not worth it. What do you think?

hal/src/sercom/v2/i2c/pads.rs
```diff
@@ -0,0 +1,128 @@
...
+impl<S, SDA, SCL> Pads<S, SDA, SCL>
+where
+    S: Sercom,
+    SDA: IsPad<PadNum = Pad0, Sercom = S> + IsI2cPad,
+    SCL: IsPad<PadNum = Pad1, Sercom = S> + IsI2cPad,
+{
+    /// Create a new [`Pads`] struct. `SDA` must always be SERCOM pad 0, and
+    /// `SCL` SERCOM pad 1.
+    #[inline]
+    pub fn new(sda: impl Into<SDA>, scl: impl Into<SCL>) -> Self {
```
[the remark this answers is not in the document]
bradleyharden: I just confirmed that this works. The only missing piece is that you need to add the bound `Pad<S, DI>: IsI2cPad<...>`, etc.

hal/src/sercom/v2/i2c/pads.rs
```diff
@@ -0,0 +1,128 @@
...
+///     i2c::Pads::default().sda(pins.pa08).scl(pins.pa09)
+/// }
+/// ```
+///
+/// [`Pin`]: crate::gpio::v2::Pin
+/// [`PinId`]: crate::gpio::v2::PinId
+/// [`OptionalPinId`]: crate::gpio::v2::OptionalPinId
+#[cfg(feature = "samd21")]
+pub type PadsFromIds<S, SDA = NoneT, SCL = NoneT> =
+    Pads<S, <SDA as GetOptionalPad<S>>::Pad, <SCL as GetOptionalPad<S>>::Pad>;
```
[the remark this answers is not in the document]
bradleyharden: Confirmed that this can be
```rust
pub type PadsFromIds<S, SDA, SCL> = Pads<S, Pad<S, SDA>, Pad<S, SCL>>;
```

hal/src/sercom/v2/i2c/pads.rs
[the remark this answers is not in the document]
jbeaurivage: I'd rather use the approach we used in other sercom modules, I agree that it's not worth it. The regular approach seems to work quite well from the BSP examples I've included!

hal/src/dmac/transfer.rs
```diff
@@ -548,6 +548,18 @@ where
         self.chan.as_mut().software_trigger();
     }
 
+    /// Unsafely and mutably borrow the source buffer
```
TDHolmes, line 551: Do these need  safety comments per the clippy PR?

hal/src/sercom/v2/dma.rs
```diff
@@ -11,12 +11,151 @@ use crate::{
...
+
+impl<C: i2c::AnyConfig> I2c<C> {
+    /// Initialize the bus to start receiving with DMA.
+    ///
+    /// In reality, this function only checks whether or not the I2C bus is
+    /// ready to accept a new transaction. A call to [`init_dma_transfer`]
+    /// should immediately be followed by a call to [`send_with_dma`] or
+    /// [`receive_with_dma`].
+    ///
+    /// ```
```
TDHolmes, line 58: Does this need a `no_run` comment? Compile only doc test

hal/src/sercom/v2/dma.rs
```diff
@@ -11,12 +11,151 @@ use crate::{
...
+
+        #[cfg(feature = "min-samd51g")]
+        let trigger_action = TriggerAction::BURST;
+
+        #[cfg(any(feature = "samd11", feature = "samd21"))]
+        let trigger_action = TriggerAction::BEAT;
+
+        // SAFETY: We use new_unchecked to avoid having to pass a 'static self as the
+        // destination buffer. This is safe as long as we guarantee the destination
+        // buffer is static.
```
TDHolmes, line 103: since the incoming buffer is user provided, does that require this function be `unsafe` and declare this safety contract?

hal/src/sercom/v2/dma.rs
```diff
@@ -11,12 +11,151 @@ use crate::{
...
+        let xfer = unsafe { dmac::Transfer::new_unchecked(channel, self, buf, false) };
+        let mut xfer = xfer
+            .with_waker(waker)
+            .begin(C::Sercom::DMA_RX_TRIGGER, trigger_action);
+
+        unsafe { xfer.borrow_source().start_dma_read(address, len as u8) };
```
TDHolmes, line 109: safety comment?

hal/src/sercom/v2/i2c.rs
```diff
@@ -0,0 +1,426 @@
...
+//!
+//! // Disable I2C peripheral
+//! let config = i2c.disable();
+//! ```
+//!
+//! # Non-supported features
+//!
+//! * Slave mode is not supported at this time.
+//! * High-speed mode is not supported.
+//! * 4-wire mode is not supported.
```
TDHolmes, line 190: 4 wire I2C?? 😮

hal/src/sercom/v2/i2c/config.rs
```diff
@@ -0,0 +1,283 @@
...
+    ) -> Self {
+        sercom.enable_apb_clock(apb_clk_ctrl);
+        Self::default(sercom, pads, freq)
+    }
+}
+
+impl<P: PadSet> Config<P> {
+    /// Obtain a reference to the PAC `SERCOM` struct
+    ///
+    /// Directly accessing the `SERCOM` could break the invariants of the
```
TDHolmes, line 78: Tag this with `# Safety`?

hal/src/dmac/transfer.rs
jbeaurivage: Indeed, fixed!

hal/src/sercom/v2/i2c.rs
jbeaurivage: Haha, I don't know what the hell it's supposed to be used for, but the datasheet says it's supported!

hal/src/sercom/v2/dma.rs
jbeaurivage: No, this is actually safe because of the bounds check in the function signature:
```rust
B: Buffer<Beat = C::Word> + 'static
```

We just check the invariants one level higher. We need to call `new_unchecked` because `self` isn't `'static`, even though the peripheral itself is obviously static. I've edited the comments to make it a bit clearer.

hal/src/sercom/v2/dma.rs
jbeaurivage: I added `no_run` pretty much everywhere, but to be honest our doctests pretty much all fail. This could be an action item to fix, but to be fair the setup required to make every example snippet compile is pretty large.

TDHolmes: yeah, that would be a large undertaking

hal/src/sercom/v2/spi.rs
```diff
@@ -45,9 +45,9 @@
 //! type Pads = spi::Pads<Sercom0, IoSet1, Miso, NoneT, Sclk>;
 //! ```
 //!
-//! Alternatively, you can use the [`PadsFromIds`] alias to define a set of
-//! `Pads` in terms of [`PinId`]s instead of `Pin`s. This is useful when you
-//! don't have `Pin` aliases pre-defined.
+//! Alternatively, you can use the `PadsFromIds` alias (not available for
```
bradleyharden, line 48: Could you add a `cfg`-based link to `PadsFromIds`? I don't like breaking the link for the narrow SAMD11 use case.

hal/src/sercom/v2/pad/impl_pad_thumbv7em.rs
```diff
@@ -191,10 +215,11 @@ pad_table!(
         D: (Sercom5, Pad3, IoSet5, IoSet6),
     }
     PB02 {
-        D: (Sercom5, Pad0, IoSet6),
+        // According to Metro M4, PB02 is I2C-capable. This disagrees with datasheet table 6-8.
+        D: (Sercom5, Pad0, IoSet6) + I2C,
     }
-    PB03 {
-        D: (Sercom5, Pad1, IoSet6),
+    PB03 {// According to Metro M4, PB03 is I2C-capable. This disagrees with datasheet table 6-8.
```
bradleyharden, line 221: Is there somewhere we could document this? I don't have any immediate ideas, but it would be nice to keep track of which pins have undocumented functions.

hal/src/sercom/v2/i2c/pads_thumbv7em.rs
```diff
@@ -0,0 +1,143 @@
...
+    SDA: IsI2cPad<PadNum = Pad0, Sercom = S> + InIoSet<I>,
+    SCL: IsI2cPad<PadNum = Pad1, Sercom = S> + InIoSet<I>,
+{
+    sercom: PhantomData<S>,
+    ioset: PhantomData<I>,
+    sda: SDA,
+    scl: SCL,
+}
+
+impl<S, I, DI, CI> Pads<S, I, Pad<S, DI>, Pad<S, CI>>
```
bradleyharden, line 26: I realized that this type is exactly `PadsFromIds<S, I, DI, CI>`. Do you think it would make sense to use that alias here?

hal/src/sercom/v2/i2c/pads_thumbv6m.rs
```diff
@@ -0,0 +1,163 @@
...
+    sda: SDA,
+    scl: SCL,
+}
+
+/// Container for a set of SERCOM [`Pad`]s
+///
+/// See the [module-level](crate::sercom::v2::i2c) documentation for more
+/// details on specifying a `Pads` type and creating instances.
+#[cfg(feature = "samd11")]
+impl<S, DI, CI> Pads<S, Pad<S, Pad0, DI>, Pad<S, Pad1, CI>>
```
bradleyharden, line 29: Same comment about `PadsFromIds`

hal/src/sercom/v2/i2c/flags.rs
```diff
@@ -0,0 +1,87 @@
...
+#[derive(Debug, Clone, Copy)]
+pub enum Error {
+    BusError,
+    ArbitrationLost,
+    LengthError,
+    Nack,
+    Timeout,
+}
+
+impl TryFrom<Status> for () {
```
bradleyharden, line 70: I think I use this approach in the `spi` module as well, but the more I see it the more I think it's awkward and not a good solution.

I think I do like the `Result<(), Error>` pattern, but not as a `TryFrom` implementation. I think an inherent method probably works better. Thoughts?

hal/src/sercom/v2/i2c.rs
```diff
@@ -0,0 +1,431 @@
...
+"
+)]
+
+#[cfg(any(feature = "samd11", feature = "samd21"))]
+#[path = "i2c/pads_thumbv6m.rs"]
+mod pads;
+
+#[cfg(feature = "min-samd51g")]
+#[path = "i2c/pads_thumbv7em.rs"]
+mod pads;
```
bradleyharden, line 272: I realized that the `path` specifier may not be necessary. Wouldn't this do the same thing?

```rust
mod pads_thumbv7em;

use pads_thumbv7em as pads;

pub use pads::*;
```

bradleyharden:
I only reviewed the HAL bits. I'll leave it to you on the BSPs. Looks good. Should be good to go soon.

hal/src/sercom/v2/pad/impl_pad_thumbv7em.rs
jbeaurivage: I documented it directly in the `sercom::v2` module-level documentation. Seems like a good place to put that stuff.

hal/src/sercom/v2/i2c/pads_thumbv6m.rs
jbeaurivage: I did the same for the samd21 impl, but `PadsFromIds` doesn't exist for samd11 targets.

hal/src/sercom/v2/i2c.rs
jbeaurivage: It is, but it takes a `#[cfg]` for both the `mod` and the `use` statements. I guess it's a matter of preference, but to me the single `#[cfg]` version seems a little less error-prone.

hal/src/sercom/v2/i2c/flags.rs
jbeaurivage: I agree, the impl is a little awkward. I made it an inherent method named `try_error`, feel free to suggest another name if you prefer.

hal/src/sercom/v2/i2c/pads_thumbv6m.rs
bradleyharden: I think it actually could for I2C. The problem with SAMD11 is that knowing the `Sercom` and the `PinId` is not enough, because there is ambiguity in the `PadNum`. But for I2C, you know the `PadNum` for both pins. I think you could do

```rust
pub type PadsFromIds<S, SDA, SCL> = Pads<S, Pad<S, Pad0, SDA>, Pad<S, Pad1, SCL>>;
```

hal/src/sercom/v2/i2c/flags.rs
bradleyharden: I don't love `try_error`. The `try_` prefix doesn't feel quite right, but I can't put my finger on why. Maybe `check_error` or `check_status`? It would be like to `check_bus_status`, which has a similar signature.

hal/src/sercom/v2/i2c/pads_thumbv6m.rs
jbeaurivage: Huh indeed you're right. Just implemented it.

hal/src/sercom/v2/i2c/flags.rs
jbeaurivage: Fair enough, I called it `check_bus_error`.

bradleyharden: Actually, what exactly is the difference between `check_bus_status` and `check_bus_error`?

jbeaurivage: `check_bus_status` is meant to be used before starting an I2C transfer, to check that the bus is ready for use. We don't want, for example, a device to be stalling the bus. `check_bus_error` is meant to be used after finishing a transfer, to check for example is a device has NACKed, or another error has occured.

sajattack:
image
:tada:

PR #567, opened 2022-01-15
circuit_playground_express: Add aliases for accel I2C pins

Rahix:
I missed these aliases when converting the board to the new pins macro...

### DIFF
--- a/boards/circuit_playground_express/src/lib.rs
+++ b/boards/circuit_playground_express/src/lib.rs
@@ -139,9 +139,15 @@ pub mod pins {
         },
         PA00 {
             name: accel_sda,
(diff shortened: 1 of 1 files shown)


bradleyharden:
I don't know the board at all, but would it also make sense to add another convenience function?

Rahix:
Yes, I actually have it on my todo list as well.  I just had this patch lying around for way too long so I wanted to at least have this part which already exists move upstream until I find time to take care of the rest.

Rahix:
Would you be okay with merging this PR as is (= just the aliases) and postponing further changes into the future (= a convenience function / tighter integration into the BSP)?

PR #569, opened 2022-02-09
Cleanup Clippy Lints

jbeaurivage:
# Summary
[describe your changes here]

# Checklist

This is a quick cleanup of most `clippy` warnings.

`clippy` also threw one error:

```diff
- while *PUKCCSR & BIT_PUKCCSR_CLRRAM_BUSY != 0 {}
+ while core::ptr::read_volatile(PUKCCSR) & BIT_PUKCCSR_CLRRAM_BUSY != 0 {}
```

### DIFF
--- a/hal/src/gpio/v2.rs
+++ b/hal/src/gpio/v2.rs
@@ -15,6 +15,8 @@
 //! between pins. However, by doing so, pins must now be tracked at run-time,
 //! and each pin has a non-zero memory footprint.
 
+#![allow(clippy::bool_comparison)]
+
 pub mod pin;
 pub use pin::*;
 
--- a/hal/src/sercom/v2/spi_future.rs
+++ b/hal/src/sercom/v2/spi_future.rs
@@ -204,6 +204,7 @@ type Data = u32;
 //=============================================================================
 
 /// Trait used to verify the [`SpiFuture`] buffer length
+#[allow(clippy::len_without_is_empty)]
 pub trait CheckBufLen: AnySpi {
     #[cfg(feature = "min-samd51g")]
     /// [`Spi`] transaction length
@@ -457,7 +458,7 @@ where
         let _ = self.spi.as_ref().read_flags_errors()?;
         let buf = self.buf.as_ref();
         if let Some(buf) = buf.get(self.sent..) {
-            let mut data = buf.into_iter();
+            let mut data = buf.iter();
             let mut bytes = [0; 4];
             let mut iter = bytes.iter_mut();
             for _ in 0..self.spi.step() {
@@ -487,7 +488,7 @@ where
         let buf = self.buf.as_mut();
         if self.rcvd < self.sent {
             let buf = unsafe { buf.get_unchecked_mut(self.rcvd..) };
-            let mut data = buf.into_iter();
+            let mut data = buf.iter_mut();
             let word = unsafe { self.spi.as_mut().read_data() as u32 };
             let bytes = word.to_le_bytes();
             let mut iter = bytes.iter();
@@ -549,6 +550,7 @@ where
     ///
     /// [`Spi`]: super::spi::Spi
     #[inline]
+    #[allow(clippy::missing_safety_doc)]
     pub unsafe fn free_unchecked(self) -> (S, B, SS) {
         (self.spi, self.buf, self.ss)
     }
--- a/hal/src/thumbv7em/clock.rs
+++ b/hal/src/thumbv7em/clock.rs
@@ -80,9 +80,9 @@ pub struct GClock {
     freq: Hertz,
 }
 
-impl Into<Hertz> for GClock {
-    fn into(self) -> Hertz {
-        self.freq
+impl From<GClock> for Hertz {
+    fn from(item: GClock) -> Hertz {
+        item.freq
     }
 }
 
@@ -114,10 +114,8 @@ impl State {
             if divider as u32 >= 2_u32.pow(16) {
                 divisor_invalid = true;
             }
-        } else {
-            if divider >= 2_u16.pow(8) {
-                divisor_invalid = true;
-            }
+        } else if divider >= 2_u16.pow(8) {
+            divisor_invalid = true;
         }
         if divisor_invalid {
             panic!("invalid divisor {} for GCLK {}", divider, gclk as u8);
@@ -364,9 +362,9 @@ impl $Type {
     }
 }
 $(#[$attr])*
-impl Into<Hertz> for $Type {
-    fn into(self) -> Hertz {
-        self.freq
+impl From<$Type> for Hertz {
+    fn from(item: $Type) -> Hertz {
+        item.freq
     }
 }
 )+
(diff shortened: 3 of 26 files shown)


TDHolmes:
I'm surprised a clippy error was allowed to fly by CI. We run clippy in CI but perhaps we don't ever fail?

hal/src/thumbv7em/clock.rs
```diff
@@ -80,9 +80,9 @@ pub struct GClock {
     freq: Hertz,
 }
 
-impl Into<Hertz> for GClock {
-    fn into(self) -> Hertz {
-        self.freq
+impl From<GClock> for Hertz {
+    fn from(item: GClock) -> Hertz {
+        item.freq
```
TDHolmes, line 85: I'm not sure if we actually want to implement `From` here. Doesn't that mean we can go both ways? `Hertz` -> `GClock` would be weird haha

hal/src/thumbv7em/pukcc.rs
```diff
@@ -567,6 +564,7 @@ impl Pukcc {
     ///
     /// All RSA variants up to **RSA4096** (included) will fit into CryptoRAM
     /// and therefore are supported.
+    #[allow(clippy::just_underscores_and_digits)]
```
TDHolmes, line 567: is the issue here the double underscores? If so could we just fix that?

hal/src/thumbv7em/nvm.rs
```diff
@@ -326,6 +326,7 @@ impl Nvm {
     ///
     /// If `destination_address` is not word-aligned, an error is returned.
     #[inline]
+    #[allow(clippy::missing_safety_doc)]
```
TDHolmes, line 329: I'd almost prefer not fixing these. It would be ideal if we could fill out `Safety` documentation.. I understand you can't fill all of them out as you didn't write this code

hal/src/thumbv7em/clock.rs
jbeaurivage: Haha yeah, you're right. I just blindly trusted clippy on that one. I'll just allow the lint instead.

hal/src/thumbv7em/pukcc.rs
jbeaurivage: Yeah, it complains about tbe double underscores. But also the module is full of super weird syntax [like this](https://github.com/atsamd-rs/atsamd/blob/62f5b959341545ff07b3b03e98b9249a3519ee0b/hal/src/thumbv7em/pukcc.rs#L617) which I don't understand, so I didn't want to touch it and break something.

hal/src/thumbv7em/nvm.rs
jbeaurivage: Agreed.

hal/src/thumbv7em/pukcc.rs
TDHolmes: That is really weird...

jbeaurivage: @vccggorski, would you mind explaining what the double underscore variables du throughout the module? If they serve a necessary practical purpose obviously I don't want to touch them, but it there's a way to refactor the module to avoid having to explicitly `#[allow]` the lints that would be great.

hal/src/thumbv7em/nvm.rs
jbeaurivage: Awesome, thank you very much!

hal/src/thumbv7em/pukcc.rs
jbeaurivage: Sounds good, think I'll just `#[allow]` the lint for the entire module then!

vccggorski: Nice :ok_hand: We went through these lints today with @vcchtjader and experimented with alternatives like `_padding` but it doesn't really express every use case of `__` in there plus it obstructs the code more than simple `__`.

Really nice job @jbeaurivage, quite mundane task but it'll help with keeping the repo clean for sure :)

jbeaurivage:
The only remaining warnings are `missing_safety_docs` in the `spi` and `spi_future` modules. @bradleyharden, do you want to take a stab at them? @TDHolmes, otherwise this'll be ready to merge. The deprecation warnings thrown in `thumbv6m::usb` will be fixed by #570.

bradleyharden:
Yeah, I can try to get rid of any warnings

TDHolmes (approved):
Everything looks good to me! thanks for this cleanup. Won't merge until @bradleyharden takes a look and/or takes a stab at the missing `Safety` comments

hal/src/sercom/v2/spi_future.rs
```diff
@@ -204,6 +204,7 @@ type Data = u32;
 //=============================================================================
 
 /// Trait used to verify the [`SpiFuture`] buffer length
+#[allow(clippy::len_without_is_empty)]
```
bradleyharden, line 207: I'm due to overhaul `SpiFuture` after using it a good amount. A lot will change.

hal/src/gpio/v2.rs
```diff
@@ -15,6 +15,8 @@
 //! between pins. However, by doing so, pins must now be tracked at run-time,
 //! and each pin has a non-zero memory footprint.
 
+#![allow(clippy::bool_comparison)]
```
bradleyharden, line 18: We can merge this, but I think I'm going to change this.

@vccggorski convinced me that `bool` is often used where there are better options, merely because there are only two choices. I think I will add a `LogicLevel` `enum` to represent `High` and `Low`. I think that will convert the `bool` comparisons and give clearer code.

bradleyharden:
@TDHolmes, I only looked through "my" modules. I didn't see any missing safety comments. What were you referring to?

hal/src/gpio/v2.rs
jbeaurivage: > @TDHolmes, I only looked through "my" modules. I didn't see any missing safety comments. What were you referring to?

@bradleyharden, specifically: `SpiFuture::free_unchecked`, `v2::spi::Config::sercom`, `v2::spi::Spi::read_data` and `v2::spi::Spi::write_data`.

jbeaurivage:
> @TDHolmes, I only looked through "my" modules. I didn't see any missing safety comments. What were you referring to?

@bradleyharden, specifically: `SpiFuture::free_unchecked`, `v2::spi::Config::sercom`, `v2::spi::Spi::read_data` and `v2::spi::Spi::write_data`. It's more about a missing `# Safety` section in the docs of unsafe functions than it is about safety comments when using an unsafe block. I hope that makes sense.

TDHolmes:
What is completed here is already awesome, perhaps we merge what is completed now instead of waiting for 100% safety comment coverage?

jbeaurivage:
> What is completed here is already awesome, perhaps we merge what is completed now instead of waiting for 100% safety comment coverage?

@TDHolmes I agree, feel free to merge at your convenience!

TDHolmes:
I dunno how it broke but feather_m0 is having issues after I tried to get this branch up to date...

PR #570, opened 2022-02-09
Prepare for v0.15 release

jbeaurivage:
# Summary

Prepare for v0.15 release. These are major breaking changes:

* Strip `gpio::v1` and `sercom::v1`, promote `gpio::v2` to `gpio` and `sercom::v2` to `sercom`.
* Strip all the `#[deprecated]` items.
* Provide the necessary fixes to support those changes

# Version numbering

Currently there are two competing opinions on whether we should version this next release as `0.15.0` or as `1.0.0`. I'd like to hear your opinions on the matter!

# Dependencies

This PR depends on #561, and ideally (but not mandatory) on #569.

# Checklist


bradleyharden:
Once we get rid of `gpio::v1`, what do you think about adding aliases to replace the `Pa1` types?

```rust
type Pa1<M> = Pin<PA01, M>;
```

Does it hurt to have both `PA01` as a `PinId` and `Pa1` as a `Pin` alias? If not, should we be consistent with leading zeros? Are these aliases even worth it? Or does the `bsp_pins!` macro take care of most use cases?

jbeaurivage:
> Once we get rid of `gpio::v1`, what do you think about adding aliases to replace the `Pa1` types?
> 
> ```rust
> type Pa1<M> = Pin<PA01, M>;
> ```
> 
> Does it hurt to have both `PA01` as a `PinId` and `Pa1` as a `Pin` alias? If not, should we be consistent with leading zeros? Are these aliases even worth it? Or does the `bsp_pins!` macro take care of most use cases?

@bradleyharden, if I'm being honest, I don't think adding more pin aliases is beneficial. Most people should be using the `bsp_pins` macro, and seeing as this release will definitely be breaking, it's a good time to only have one way of referring to GPIO pins. IMO, having multiple aliases to the same pin would probably lead to confusion. I remember when the `gpio::v2` module came along, it took me a while to understand that `Pa1` was now supposed to be called `PA01`.

bradleyharden:
Sounds good. It was just an idea I had long ago, but it wasn't feasible then without breaking existing code.

hal/README.md
```diff
@@ -32,9 +32,9 @@ This crate can support other variants in a similar fashion; pull requests for th
 
 ## Examples?
 
-Check out the metro_m0 board support crate examples:
+Check out the `feather_m0` board support crate examples:
 
-https://github.com/atsamd-rs/atsamd/tree/master/boards/metro_m0/examples
+https://github.com/atsamd-rs/atsamd/tree/master/boards/feather_m0/examples
```
TDHolmes, line 37: perhaps we should more generally point folks at tier 1 BSPs?

boards/samd11_bare/Cargo.toml
```diff
@@ -35,6 +35,9 @@ use_semihosting = []
 [profile.release]
 debug = true
 
+[profile.dev]
+opt-level = "s"
```
TDHolmes, line 39: any particular reason for this? Not opposed to adding it, though.

TDHolmes (approved):
amazing 👏🏼 anything remaining to do here?

boards/samd11_bare/Cargo.toml
jbeaurivage: Yeah, some examples (looking at you, i2c) won't compiled unoptimized because they take too much flash space.

hal/README.md
jbeaurivage: I don't think this has to be in the scope of this PR, but yeah, nudging then towards T1 BSPs as examples for writing their own is definitely a good idea.

jbeaurivage:
> amazing 👏🏼 anything remaining to do here?

Nope I don't think so, I would merge #569 first though!

PR #576, opened 2022-02-21
Enable optional RTIC support for Xiao_m0

kadencartwright:
# Summary
It would be nice if the seeduino xiao_m0 supported RTIC. I'm a bit of a rust newbie, so maybe this is a naive fix¿? but it seems as simple as adding a feature flag to enable the `rtic` feature in the hal dependency.

# Checklist

## If Adding a new Board

### DIFF
--- a/boards/xiao_m0/Cargo.toml
+++ b/boards/xiao_m0/Cargo.toml
@@ -36,6 +36,7 @@ default = ["rt", "atsamd-hal/samd21g", "atsamd-hal/unproven"]
 rt = ["cortex-m-rt", "atsamd-hal/samd21g-rt"]
 unproven = ["atsamd-hal/unproven"]
 usb = ["atsamd-hal/usb", "usb-device"]
+rtic = [ "atsamd-hal/rtic"]
 
 [[example]]
 name = "blink"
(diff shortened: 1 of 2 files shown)


TDHolmes:
I think this is the right thing to do, but could you add an RTIC example as well? You should be able to copy/paste an existing M0 one. I think feather_m0 has one?

kadencartwright:
yea, I'll do that!

boards/xiao_m0/Cargo.toml
```diff
@@ -16,13 +16,15 @@ optional = true
 
 [dependencies.atsamd-hal]
 version = "0.14"
+path = "../../hal"
```
kadencartwright, line 19: I had to update the Hal to reference the repo version because a Monotonic related trait was not being satisfied. Feather M0 does the same in it's repo, I would assume for this same reason

kadencartwright:
I had to update the Hal to dependency in Cargo.toml to reference the repo version using `path = ...` because a Monotonic related trait bound was not being satisfied otherwise. Feather M0 does the same in it's repo, which I would assume for this same reason.

kadencartwright:
@bradleyharden Ah, I see. Makes sense to me. What's the best thing to do with this PR for the time being -- should I just leave it here as is or is there something else more helpful I can do with it?

TDHolmes:
I think we just need to wait on this one until v0.15 is released

TDHolmes (changes requested):
Make changes now that there has been a HAL release

bradleyharden:
Pinging @kadencartwright. Do you want to finish this up now that #570 has been merged?

Magkoooh (approved):

PR #579, opened 2022-03-12
Remove problematic shortcut in USB poll()

ianrrees:
I'm hoping to get some testing feedback; I only have a couple Linux machines to work with, and wonder if the code that this PR deletes might still address an issue on some other host/OS.

I believe this shortcut sometimes causes SETUP transactions to be dropped, which results in control transfers failing.  One way to see this on Linux is to run `# lsusb -v -d <VID>:<PID>` for a SAMD-based USB device (the usb-echo example that many boards provide, for instance) - sometimes (more than 1 in 10 on my machine) this will take a second or more to return, where it should only take a few milliseconds.  The mechanism is something like:

1. An IN transfer completes from EP0 (bank 1)
2. A SETUP transaction arrives on EP0 (bank 0)
3. The atsamd-rs `poll()` is called by [the usb-device `poll()`](https://github.com/mvirkkunen/usb-device/blob/b6037b0f98c933ff0a18a864fd34a42f89a6316a/src/device.rs#L162)
4. atsamd-rs `poll()` sees the complete IN flag on EP0 and doesn't set the set the bit in `ep_setup` for EP0 (due to the `continue;` this PR deletes).
5. usb-device `poll()` then [calls `ControlPipe::handle_out()`](https://github.com/mvirkkunen/usb-device/blob/b6037b0f98c933ff0a18a864fd34a42f89a6316a/src/device.rs#L209) which [calls the endpoint `read()`](https://github.com/mvirkkunen/usb-device/blob/b6037b0f98c933ff0a18a864fd34a42f89a6316a/src/control_pipe.rs#L151) on an empty buffer and thereby clears the RXSTP flag on the endpoint - dropping the SETUP packet.
6. Host eventually times out the transfer.

### DIFF
--- a/hal/src/thumbv6m/usb/bus.rs
+++ b/hal/src/thumbv6m/usb/bus.rs
@@ -906,31 +906,28 @@ impl Inner {
 
             let idx = ep as usize;
 
-            let bank1 = self
-                .bank1(EndpointAddress::from_parts(idx, UsbDirection::In))
-                .unwrap();
-            if bank1.is_transfer_complete() {
-                bank1.clear_transfer_complete();
-                dbgprint!("ep {} WRITE DONE\n", ep);
-                ep_in_complete |= mask;
-            }
-            drop(bank1);
-
-            let bank0 = self
-                .bank0(EndpointAddress::from_parts(idx, UsbDirection::Out))
-                .unwrap();
-            if bank0.received_setup_interrupt() {
-                dbgprint!("ep {} GOT SETUP\n", ep);
-                ep_setup |= mask;
-                // usb-device crate:
-                //  "This event should continue to be reported until the packet
-                // is read." So we don't clear the flag here,
-                // instead it is cleared in the read handler.
+            if let Ok(bank1) = self.bank1(EndpointAddress::from_parts(idx, UsbDirection::In)) {
+                if bank1.is_transfer_complete() {
+                    bank1.clear_transfer_complete();
+                    dbgprint!("ep {} WRITE DONE\n", ep);
+                    ep_in_complete |= mask;
+                }
             }
 
-            if bank0.is_transfer_complete() {
-                dbgprint!("ep {} READABLE\n", ep);
-                ep_out |= mask;
+            if let Ok(bank0) = self.bank0(EndpointAddress::from_parts(idx, UsbDirection::Out)) {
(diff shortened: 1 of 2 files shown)


ianrrees:
Hm, this is going to need some more work - the usb-echo example seems to hang the micro...  I'll look in to it more tomorrow.

ianrrees:
Somehow, "tomorrow" turned in to over a week...

Removing that shortcut made the usb_echo example reveal an issue with `poll()`, which the second commit in this PR fixes: if there was an endpoint index which had only an IN or an OUT endpoint (not both), and an interrupt fired for that index, then `poll()` would `unwrap()` a call to the accessors for each bank on that index.  In the case where only one was initialised, one of those accessors would return `Err(UsbError::InvalidEndpoint)` and so the `unwrap()` would panic.

In the usb_echo example, endpoint index 2 has an IN (bank1) endpoint but no OUT (bank0), which worked before because all interrupts were for that IN endpoint, and the `continue;` removed in the first commit of this PR would skip over the bank0 access.

cr1901:
I'm attempting to port Trinket M0 to v0.16, and the USB serial example stopped working on Windows 10. Based on `git bisect` can confirm that 14a19dd33de92766 (this PR) is causing timeouts like the following:

image.

It _is_ possible to get a successful enumeration, but about 5 seconds after, device manager errors the device and shows a "caution" symbol: 
image.

I have a USB analyzer, so I can test in more detail, but I'm wondering if anyone else can duplicate.

Despite me porting Trinket M0 to v0.16, _note that the extent of my changes to 14a19dd33de92766 to duplicate a malfunctioning device are this:_

```diff
$ git diff
diff --git a/boards/trinket_m0/Cargo.toml b/boards/trinket_m0/Cargo.toml
index 3e27b6c84..1444dab77 100644
--- a/boards/trinket_m0/Cargo.toml
+++ b/boards/trinket_m0/Cargo.toml
@@ -20,7 +20,7 @@ version = "0.7"
 optional = true

 [dependencies.atsamd-hal]
-version = "0.14"
+path = "../../hal"
 default-features = false

 [dependencies.usb-device]
```

jbeaurivage:
@cr1901, I'm slowly starting to look into this. I know close to nothing about USB unfortunately, but I'm trying to learn the basics as I'm running into enumeration issues as well.

PR #580, opened 2022-03-15
arduino_mkrzero: add blinky_rtic example

arkap:
# Summary
Add a blinky_rtic example for arduino_mkrzero.

I've adapted the example from the feather_m0 board for the arduino_mkrzero board. It's currently throwing a lot of warnings because of the use of `gpio::v1`.

### DIFF
--- a/boards/arduino_mkrzero/Cargo.toml
+++ b/boards/arduino_mkrzero/Cargo.toml
@@ -15,6 +15,7 @@ version = "0.7"
 optional = true
 
 [dependencies.atsamd-hal]
+path = "../../hal"
 version = "0.14"
 default-features = false
 
@@ -22,11 +23,16 @@ default-features = false
 version = "0.2"
 optional = true
 
+[dependencies.embedded-sdmmc]
+version = "0.3"
+optional = true
+
 [dev-dependencies]
 cortex-m = "0.7"
 usbd-serial = "0.1"
 panic-halt = "0.2"
 panic-semihosting = "0.5"
+cortex-m-rtic = "1.0"
 
 [features]
 # ask the HAL to enable atsamd21g support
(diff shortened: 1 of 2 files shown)


boards/arduino_mkrzero/Cargo.toml
```diff
@@ -15,18 +15,24 @@ version = "0.7"
 optional = true
 
 [dependencies.atsamd-hal]
+path = "../../hal"
```
jbeaurivage, line 18: This line should be removed. `arduino_mkrzero` is a Tier 2 board, so it should use a `crates.io` dependency to `atsamd-hal` rather than a path dependency.

arkap: If I remove that line, the build fails with
```
the trait `Monotonic` is not implemented for `arduino_mkrzero::rtc::Rtc<arduino_mkrzero::rtc::Count32Mode>`
```
Does that mean upstream `atsamd-hal` needs to catch up? I don't quite understand what the issue is but if it's a matter of implementing `Monotonic`, I'd be happy to give it a try.

jbeaurivage: That's because the latest `crates.io` `atsamd-hal` version still has one of the `0.6.0-rc.x` RTIC dependencies. `Monotonic` is implemented, but there's a version mismatch with you `cortex-m-rtic = "1.0"` dependency. You can either wait for atsamd-hal to catch up, which should hopefully happen soon, or match your RTIC version with the one from `atsamd-hal`.

arkap: Alright, let's give atsamd-hal some time and wait. Thanks for the insight.

TDHolmes:
Another one waiting on #570

TDHolmes (changes requested):
Make changes now that there has been a HAL release

bradleyharden:
@arkap, just pinging you about this, since #570 was merged.

sajattack:
Superceded by #697

PR #601, opened 2022-04-02
Fix .cargo/config for Feather M0

rtsuk:
Prior version didn't compile or run with a recent
Feather M0.

# Summary
Moved rustflags to the build section and changed the chip parameter
to probe-run

# Checklist


boards/feather_m0/.cargo/config
```diff
@@ -2,16 +2,13 @@
...
   # This is needed if your flash or ram addresses are not aligned to 0x10000 in memory.x
   # See https://github.com/rust-embedded/cortex-m-quickstart/pull/95
   "-C", "link-arg=--nmagic",
-  
   "-C", "link-arg=-Tlink.x",
 ]
+
+[target.thumbv6m-none-eabi]
+runner = 'arm-none-eabi-gdb'
+#runner = 'probe-run --chip ATSAMD21E17A'
```
TDHolmes, line 14: The feather M0 is a `ATSAMD21G18`, so this needs to change back

TDHolmes (changes requested):

boards/feather_m0/.cargo/config
rtsuk: Unfortunately, that value for chip doesn't work.

```
     Running `probe-run --chip ATSAMD21G18A target/thumbv6m-none-eabi/release/examples/blinky_basic`
(HOST) INFO  flashing program (19 pages / 1.19 KiB)
Error: Error while flashing

Caused by:
    0: Failed to erase flash sector at address 0x00000000.
    1: The execution of 'erase_sector' failed with code 1. This might indicate a problem with the flash algorithm.
feather_m0 $ 
```

rtsuk: But, I'm happy to leave it that way, I can always change it locally if I want to use the m0.

rtsuk: Changed.

TDHolmes: Weird... is it not properly supported on the probe-rs side?

rtsuk: That's a good guess. Perhaps the next time I want to use the m0 I'll spend some time debugging it.

Issue #603, opened 2022-04-13
Recent version bump broke essentially all text rendering


cbiffle:
PR #591 bumped versions of dependencies. While the PR description described the bump as "minor," many of the crates that were bumped are in the leading-0 phase of semver, meaning that minor version bumps are breaking.

One good example of breakage is the `embedded_graphics` crate, which appears to have switched the default baseline behavior for text rendering in (I'm guessing) 0.7. As a result, after #591, many of the examples render their text off the screen, and programs that attempt to clear the screen before writing text (the `clock` examples, say) instead clear a rectangle just beneath the text.

Fixing this generally requires diffs like the following:

```diff
diff --git a/boards/wio_terminal/examples/clock.rs b/boards/wio_terminal/examples/clock.rs
index 9e4567f29..2f0c02f72 100644
--- a/boards/wio_terminal/examples/clock.rs
+++ b/boards/wio_terminal/examples/clock.rs
@@ -15,7 +15,7 @@ use eg::mono_font::{ascii::FONT_10X20, MonoTextStyle};
 use eg::pixelcolor::Rgb565;
 use eg::prelude::*;
 use eg::primitives::{PrimitiveStyleBuilder, Rectangle};
-use eg::text::Text;
+use eg::text::{Baseline, Text};
 
 use cortex_m::interrupt::free as disable_interrupts;
 use cortex_m::peripheral::NVIC;
@@ -137,7 +137,7 @@ fn main() -> ! {
             .ok()
             .unwrap();
 
-        Text::new(data.as_str(), Point::new(55, 80), style)
+        Text::with_baseline(data.as_str(), Point::new(55, 80), style, Baseline::Top)
             .draw(&mut display)
             .ok()
             .unwrap();
```

Anyway, noticed this when I was using your code to get a shiny new Wio Terminal working before I write my own drivers. :-)

TDHolmes:
If you have the change to fix, please open a PR! Thanks for the report

cbiffle:
I have the fix for the one example I used to test flashing (above); I'm not equipped to test all the examples. I also can't promise this is the only breaking change in embedded-graphics 0.7.

TDHolmes:
As you can tell by us breaking the examples, we don't necessarily have all of the hardware on hand at all times either 🙂 even a partial fix PR is appreciated

bradleyharden:
[...]

PR #604, opened 2022-04-14
Fix text baseline broken in recent version bump

cbiffle:
As I noted in #603, PR #591 bumped a bunch of crates across what are
effectively major version boundaries, because the crates are in the
leading-zero phase of semver -- so an 0.6 to 0.7 update is breaking.
This has broken _at least_ text rendering.

This commit fixes the text baseline calculations to use the method
expected by the embedded-graphics 0.7 crate, which has (quite
reasonably) switched to using baselines instead of bounding box top to
place text, but opted to leave the old implicit-baseline `new` operation
in place and change its behavior, which allowed all of this to break
without any compile errors.

NOTE THAT WITH THIS COMMIT NEARLY ALL OF THE WIO TERMINAL DEMOS ARE
STILL BADLY BROKEN. I'm new here and I'm not sure if these worked before
the version bump, or if they were also broken by the merge of this
untested PR, but at this point, only clock and usb_serial_display work.
Everything else displays blinking garbage on the display and hangs.

# Summary
[describe your changes here]

# Checklist

## If Adding a new Board

### DIFF
--- a/boards/wio_terminal/examples/clock.rs
+++ b/boards/wio_terminal/examples/clock.rs
@@ -15,7 +15,7 @@ use eg::mono_font::{ascii::FONT_10X20, MonoTextStyle};
 use eg::pixelcolor::Rgb565;
 use eg::prelude::*;
 use eg::primitives::{PrimitiveStyleBuilder, Rectangle};
-use eg::text::Text;
+use eg::text::{Baseline, Text};
 
 use cortex_m::interrupt::free as disable_interrupts;
 use cortex_m::peripheral::NVIC;
@@ -137,7 +137,7 @@ fn main() -> ! {
             .ok()
             .unwrap();
 
-        Text::new(data.as_str(), Point::new(55, 80), style)
+        Text::with_baseline(data.as_str(), Point::new(55, 80), style, Baseline::Top)
             .draw(&mut display)
             .ok()
             .unwrap();
--- a/boards/wio_terminal/examples/microphone.rs
+++ b/boards/wio_terminal/examples/microphone.rs
@@ -19,7 +19,7 @@ use eg::mono_font::{ascii::FONT_6X12, MonoTextStyle};
 use eg::pixelcolor::Rgb565;
 use eg::prelude::*;
 use eg::primitives::{PrimitiveStyleBuilder, Rectangle};
-use eg::text::Text;
+use eg::text::{Baseline, Text};
 use heapless::consts::*;
 use heapless::String;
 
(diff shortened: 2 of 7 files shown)


cbiffle:
To partially answer my own question: I've rewound past the version bump commit to the commits preparing for the 0.15 release, and to the 0.14 release, and the Wio Terminal examples were already broken, though the screen garbage looks different than after the version bump. So, that commit is off the hook for part of the breakage.

sajattack:
@jessebraham @twitchyliquid64 this is what happens when you leave your boards unattended :stuck_out_tongue:

cbiffle:
FWIW it turns out some of the display corruption is because the graphics examples have, for some time, been attempting to clear a 320 x 320 screen; the Wio Terminal's screen is 320 x 240. It looks like the behavior of this changed recently somewhere in the web of dependencies. Fixing the screen sizes for e.g. the qspi example causes it to at least correctly clear the screen, though it still doesn't work.

cbiffle:
And just in case anyone's riveted by these updates: so it turns out you can get the qspi example display updates to work by _raising the display clock frequency_ to 58 MHz, which I tried because it's the speed used by `usb_serial_display`, one of the demos that worked.

With that and the display size fix, the qspi demo on master works.

I assume the qspi demo did work at some point, so, not sure what changed. Maybe your clock setup code can't achieve 24 MHz exactly and its fallback behavior is different than it used to be?

cbiffle:
Aaand fwiw, with the display clock change to 58, all of the demos work on a newly unboxed Wio Terminal _except_ the wifi ones, which hang.

The sdcard demo begins by drawing on an uninitialized display, so the first pass tends to look like garbage, but after 2.5 seconds it clears the screen properly and tries again.

So, I conclude that something is wrong with the clock setup routine.

TDHolmes:
@cbiffle , thanks for looking into all of this. Do you want to pull in any more of these changes into this PR, or keep it to just the text baseline change?

cbiffle:
Thanks for asking! No, I'm really hesitant to offer any fixes when I don't understand the problem. After you find a working clock frequency, nudging it up or down by 1MHz can cause the display driver to hang, which seems... suspicious.

TDHolmes:
Lol agreed, quite sus. We can just take this for now then.

pabigot:
@cbiffle Thanks for your diagnosis and fix.  I've rebased this on current master, added the fix in https://github.com/atsamd-rs/atsamd/pull/604#issuecomment-1099736323, and done one more in [my main](/pabigot/atsamd/tree/pabmain).  Now I'm just down to wifi making the wio act like a brick.

PR #607, opened 2022-05-06
Fix UART pad definitions

### LINKED ISSUE #581: SERCOM0 samd51g19a - Using PA06 for RX in UART results in a Frame Error when reading
The configuration used:

```Rust
pub fn uart0(
    clocks: &mut GenericClockController,
    baud: impl Into<Hertz>,
    sercom0: pac::SERCOM0,
    mclk: &mut pac::MCLK,
    uart_rx: impl Into<Pin<PA06, AlternateD>>,
    uart_tx: impl Into<Pin<PA04, AlternateD>>,
) -> Uart0 {
    let gclk0 = clocks.gclk0();
    let clock = &clocks.sercom0_core(&gclk0).unwrap();
    let baud = baud.into();
    let pads = uart::Pads::default().rx(uart_rx.into()).tx(uart_tx.into());
    uart::Config::new(mclk, sercom0, pads, clock.freq())
        .baud(baud, BaudMode::Fractional(Oversampling::Bits16))
        .enable()
}
```
- With `RXC` and `TXC` interrupts enabled 
- Writing works, error is only triggered when trying to read 
- Hardware is and Itsybitsy m4. Tested on two different boards
- Pin does work as a ADC0 input

Using `Pin<PA05, AlternateD>` works.
Tested with PA07 as RX and it seemed funky too, but I didn't look deeper into it. Will take my debug probe later and confirm if this also causes a frame error or if it was just momentary funkyness.

jbeaurivage:
# Summary

The pad definitions in sercom::uart didn't take into account that the XCK pad would be unavailable for use, even if ynchronous mode isn't in use Especially, implement the same `impl_rxpotxpo` macro system for `thumbv7em` targets as the one already in use for `thumbv6m targets`. Additionnally, reject pad combinations where XCK would conflict on `thumbv6m` targets. Closes #581.

Technically these should be breaking changes, as some pad combinations that used to be valid won't be anymore, but considering that the combinations that were invalidated by this PR were already broken to begin with, I don't think we need to wait for the next major release to squeeze it in.

# Checklist

### DIFF
--- a/hal/src/sercom/uart/pads_thumbv7em.rs
+++ b/hal/src/sercom/uart/pads_thumbv7em.rs
@@ -1,46 +1,42 @@
 //! UART pad definitions for thumbv7em targets
 
+use super::{AnyConfig, Capability, CharSize, Config, Duplex, Rx, Tx};
+use crate::{
+    gpio::AnyPin,
+    sercom::*,
+    typelevel::{NoneT, Sealed},
+};
 use core::marker::PhantomData;
 
-use crate::pac::sercom0::usart_int::ctrla::{RXPO_A, TXPO_A};
-
-use super::{AnyConfig, Capability, CharSize, Config, Duplex, Rx, Tx};
-use crate::sercom::*;
-use crate::typelevel::{NoneT, Sealed};
-
-use crate::gpio::AnyPin;
-
 //=============================================================================
-// Rxpo
+// RxpoTxpo
 //=============================================================================
 
-/// Control the `RXPO` field as a function of the [`PadNum`] type
-pub trait Rxpo: Sealed {
-    /// Corresponding variant from the PAC `enum`
-    const RXPO: RXPO_A;
-}
-
-impl Rxpo for Pad0 {
-    const RXPO: RXPO_A = RXPO_A::PAD0;
-}
-impl Rxpo for Pad1 {
-    const RXPO: RXPO_A = RXPO_A::PAD1;
-}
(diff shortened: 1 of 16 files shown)


jbeaurivage:
This is ready for merging - @Josfemova confirms it works. @bradleyharden, maybe you want to take a look, since you're probably the most familiar with the push-down-permutations-macro-thingy?

hal/src/sercom/uart/pads_thumbv7em.rs
```diff
@@ -1,91 +1,156 @@
...
-    /// Corresponding variant from the PAC `enum`
-    const TXPO: TXPO_A;
-}
-
-impl Txpo for Pad0 {
-    const TXPO: TXPO_A = TXPO_A::TXPO_2;
-}
+/// Filter [`PadNum`] permutations and implement [`RxpoTxpo`]
+    #[rustfmt::skip]
+    macro_rules! impl_rxpotxpo {
```
bradleyharden, line 59: Shouldn't this be indented less?

bradleyharden (approved):
Looks good to me

hal/src/sercom/uart/pads_thumbv7em.rs
jbeaurivage: Ah yeah indeed. I'll fix that tomorrow, then I think we can merge

jbeaurivage:
I'll just go ahead and merge this since @bradleyharden already approved it

PR #610, opened 2022-05-13
Updated `atsamd-hal` to 0.15.1 for pyportal

wembly:
# Summary
Updated `atsamd-hal` from 0.14.0 to 0.15.1
Incremented crate version to 0.10.0
Added usb feature to bsp, and example
Added display feature to bsp and example

# Checklist

### DIFF
--- a/boards/pyportal/Cargo.toml
+++ b/boards/pyportal/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pyportal"
-version = "0.9.0"
+version = "0.10.0"
 authors = [
     "Shella Stephens <shella@infracoven.io",
     "Paul Sajna <sajattack@gmail.com>",
@@ -15,17 +15,32 @@ readme = "README.md"
 edition = "2021"
 
 [dependencies.cortex-m-rt]
-version = "0.7"
+version = "0.7.1"
 optional = true
 
 [dependencies.atsamd-hal]
-version = "0.14"
+version = "0.15.1"
 default-features = false
+
+[dependencies.display-interface-parallel-gpio]
+version="0.6.0"
+optional=true
+
+[dependencies.ili9341]
+version="0.5.0"
+optional=true
+
+[dependencies.usb-device]
+version = "0.2.8"
+optional = true
 
 [dev-dependencies]
 panic-halt = "0.2"
 panic-semihosting = "0.5"
 smart-leds = "~0.3"
+usbd-serial = "0.1.1"
+cortex-m = "0.7.4"
+embedded-graphics = "0.7.1"
 
 [dev-dependencies.ws2812-timer-delay]
 version = "~0.3"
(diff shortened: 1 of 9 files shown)


boards/pyportal/Cargo.toml
```diff
@@ -15,26 +15,43 @@ readme = "README.md"
...
+optional=true
+
+[dependencies.usb-device]
+version = "0.2.8"
+optional = true
+
 [dev-dependencies]
 panic-halt = "0.2"
 panic-semihosting = "0.5"
 smart-leds = "~0.3"
```
sajattack, line 40: We might want to move this out of dev-dependencies and add some support functions to the bsp. Up to you though.

sajattack:
Looks pretty good to me!

boards/pyportal/Cargo.toml
sajattack: Ugh, github kinda broke my comment, I was referring to `smart-leds`

wembly: following up on this: we discussed this online and decided not to add this dependency at this point in time

boards/pyportal/src/display.rs
```diff
@@ -0,0 +1,53 @@
+use super::pins::*;
+use atsamd_hal::ehal::blocking::delay::DelayMs;
+use atsamd_hal::prelude::*;
+use display_interface_parallel_gpio::{Generic8BitBus, PGPIO8BitInterface};
```
sajattack, line 4: surprised if the ili9341 crate doesn't re-export this

sajattack, line 4: surprised if the ili9341 crate doesn't re-export this

wembly: unfortunately it is not re-exported from what i can tell

boards/pyportal/src/pins.rs
```diff
@@ -1,116 +1,350 @@
...
+        name: neopixel,
+        aliases: {
+            Reset: NeopixelReset,
+            PushPullOutput: Neopixel,
+        }
+    }
+
+    PA00 {
+        // TFT(Thin-film-transistor liquid-crystal display) control ports
+        /// TFT Reset
```
bradleyharden, line 64: A few of these have part of the documentation as a comment and part as actual docs. Is there a reason for that?

wembly: i carried that over from the previous version, from what i've been able to discern the comments are to indicate logical groupings of pins in the code, while the documentation is for to document the pin itself

wembly:
> This looks good to me. Have all the examples been tested?

I've tested them all on the physical pyportal hardware i have by using `cargo run --example <example name> --release` if that's what your asking

bradleyharden:
Yeah, that's what I was asking. I think this is good to merge then. @sajattack, are you good?

PR #616, opened 2022-05-30
Use type-level programming instead of macros for `adc`

bradleyharden:
The existing `adc` modules use macros to be generic over the ADC
peripheral. Update the code to use type-level programming techniques
instead, but do so in a backwards-compatible way.

# Checklist

### DIFF
--- a/hal/src/thumbv7em/adc.rs
+++ b/hal/src/thumbv7em/adc.rs
@@ -1,5 +1,8 @@
 //! Analogue-to-Digital Conversion
-use crate::clock::GenericClockController;
+
+use core::ops::Deref;
+
+use crate::clock::{GClock, GenericClockController};
 #[rustfmt::skip]
 use crate::gpio::*;
 use crate::ehal::adc::{Channel, OneShot};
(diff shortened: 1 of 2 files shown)


hal/src/thumbv7em/adc.rs
```diff
@@ -235,100 +297,111 @@ impl<C> InterruptAdc<$ADC, C>
...
+    fn channel() -> u8 {
+        I::CHANNEL
+    }
 }
 
 macro_rules! adc_pins {
     (
         $(
-            $PinId:ident: ($ADC:ident, $CHAN:literal),
+            $PinId:ident: [ $( ($ADC:ident, $CHAN:literal) ),+ ],
```
bradleyharden, line 348: I made this a list of tuples rather than a single tuple, because some pins can be an input for both ADC0 and ADC1. We should update the pin table below to match the datasheet. It's missing a bunch of entries right now.

bradleyharden:
If you want to get more aggressive with the refactor, I see a lot of areas where this module could be cleaned up and simplified. But for now, I wanted to keep everything backwards-compatible.

bradleyharden:
@wembly, it sounds like you were able to get something working for `thumbv6m`. If so, feel free to make a PR to this branch on my repo, and I'll merge it in. Then we can get a review fromsomeone else for the whole package.

PR #617, opened 2022-06-03
Make items from thumbv(6|7e)m modules visible in the main docs page

glaeqen:
# Summary

Main page in docs does not contain contents of `thumb*` modules which in turn should be an implementation detail. We can either

- Add `doc` attributes to force a desired behaviour of `rustdoc`
- Make modules private which will cause automatic documentation inlining (but will introduce a breaking change)

This PR is supposed to serve as a discussion board.

### DIFF
--- a/hal/src/lib.rs
+++ b/hal/src/lib.rs
@@ -110,13 +110,17 @@ compile_error!("'usb' is enabled, but USB isn't supported on SAMD11");
 compile_error!("The 'usb' feature is enabled, but not a chip with USB support");
 
 #[cfg(any(feature = "samd11", feature = "samd21"))]
+#[doc(hidden)]
 pub mod thumbv6m;
 #[cfg(any(feature = "samd11", feature = "samd21"))]
+#[doc(inline)]
 pub use crate::thumbv6m::*;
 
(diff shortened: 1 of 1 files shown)


jbeaurivage:
I'm in favour of making the  `thumb*` modules private. We would have to bump semver as a breaking change.

vccggorski:
Almighty @bradleyharden, please tune in and let us know what you think. On a slightly related note, I noticed weird `rustdoc` behaviour in #450 that for `clock.rs` it inlines re-exports entries regardless if `v1` is `pub` or not. Suspicious.

bradleyharden:
I'm also in favor of making these private and an implementation detail. In fact, I think we should be trying to consolidate where we can and reduce or eliminate their submodules, like we did for `sercom`.

bradleyharden:
We've already released a non-breaking change at 0.15.1, so I think we can move forward with breaking changes now.

bradleyharden:
I'm going to merge this now. It doesn't break anything, but it puts new users on the right path. As a next step, let's add a deprecation warning to those modules, before we make them fully private.

PR #618, opened 2022-06-07
Fix bug in `dmac` where software triggers didn't work

jbeaurivage:
# Summary
I realized that the `dmac` examples were no longer working. After running `git bisect`, I narrowed down the bug to #432, where the register proxy architecture introduced a bug preventing DMAC software triggers from working. This PR fixes that, along with removing an unnecessary `unsafe` block in the `dmac` module.

# Checklist

### DIFF
--- a/hal/src/dmac/channel/mod.rs
+++ b/hal/src/dmac/channel/mod.rs
@@ -199,7 +199,7 @@ impl<Id: ChId, S: Status> Channel<Id, S> {
 
     #[inline]
     fn _trigger_private(&mut self) {
-        self.regs.swtrigctrl.write_bit(true);
+        self.regs.swtrigctrl.set_bit();
     }
 }
 
--- a/hal/src/dmac/channel/mod.rs
+++ b/hal/src/dmac/channel/mod.rs
@@ -252,19 +252,18 @@ impl<Id: ChId> Channel<Id, Ready> {
         // Configure the trigger source and trigger action
         // SAFETY: This is actually safe because we are writing the correct enum value
         // (imported from the PAC) into the register
-        unsafe {
-            #[cfg(any(feature = "samd11", feature = "samd21"))]
-            self.regs.chctrlb.modify(|_, w| {
-                w.trigsrc().bits(trig_src as u8);
-                w.trigact().bits(trig_act as u8)
-            });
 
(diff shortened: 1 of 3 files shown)


hal/src/dmac/channel/reg.rs
```diff
@@ -243,7 +243,7 @@ macro_rules! reg_proxy {
                     // to the bit controlled by the channel.
                     self.dmac
                         .[< $reg:lower >]
-                        .modify(|r, w| unsafe { w.bits(r.bits() & ((bit as u32) << Id::U8)) });
+                        .modify(|_, w| unsafe { w.bits((bit as u32) << Id::U8) });
```
TDHolmes, line 246: Isn't there a just write API if you aren't going to read?

hal/src/dmac/channel/mod.rs
```diff
@@ -252,19 +252,18 @@ impl<Id: ChId> Channel<Id, Ready> {
         // Configure the trigger source and trigger action
         // SAFETY: This is actually safe because we are writing the correct enum value
```
TDHolmes, line 253: No longer needed

jbeaurivage: True that. I'll update that now

hal/src/dmac/channel/reg.rs
jbeaurivage: You're right. I made a habit of using almost exclusively `modify` because that's normally what we want, but in this case since `bits` writes the whole register anyway, it doesn't really make a difference. I'm worried that would clear any other 'pending' software trigger requests though. The datasheet isn't really clear on this (ATSAMD21 19.8.1.8). thoughts?

jbeaurivage: Actually, I realized it was just my bit manipulation logic that was wrong. [This commit](https://github.com/atsamd-rs/atsamd/pull/618/commits/bd56e9dc382782179d830d20280be30ad4326a6e) fixes that. I'm confident this can be merged now!

bradleyharden:
Hmm, looks like GHA is having problems. But the tests ran successfully just before merging #626, and that PR doesn't touch any of the same files, so I think it is safe to merge anyway.

PR #622, opened 2022-06-18
Add Board support package for Adafruit Feather M4 CAN

chipsoft:
# Summary
Add Board support package for Adafruit Feather M4 CAN (NeoPixel support), add examples for Blinky and NeoPixel

# Checklist

## If Adding a new Board

### DIFF
--- a/boards/feather_m4_can/.cargo/config.toml
+++ b/boards/feather_m4_can/.cargo/config.toml
@@ -1,0 +1,15 @@
+# vim:ft=toml:
+[target.thumbv7em-none-eabihf]
+runner = 'arm-none-eabi-gdb'
+# runner = 'probe-run --chip ATSAMD51J19A'
+
+[build]
+target = "thumbv7em-none-eabihf"
+rustflags = [
+
+   # This is needed if your flash or ram addresses are not aligned to 0x10000 in memory.x
+   # See https://github.com/rust-embedded/cortex-m-quickstart/pull/95
+   "-C", "link-arg=--nmagic",
+
+   "-C", "link-arg=-Tlink.x",
+]
--- a/boards/feather_m4_can/CHANGELOG.md
+++ b/boards/feather_m4_can/CHANGELOG.md
@@ -1,0 +1,27 @@
+# Unreleased
+
+# v0.10.1
+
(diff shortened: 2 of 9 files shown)


sajattack:
I think this would be better as a `#[cfg(feature(can))]` like we did for the radio variants of feather m0 (https://github.com/atsamd-rs/atsamd/blob/b82ab9ab0566deace39a6e8af6c963c33d6f8540/boards/feather_m0/src/lib.rs#L177). Also I don't think we have CAN support in the HAL so I'm not sure how different it would be from the regular feather_m4.

chipsoft:
> I think this would be better as a `#[cfg(feature = "can"]` like we did for the radio variants of feather m0
> 
> https://github.com/atsamd-rs/atsamd/blob/b82ab9ab0566deace39a6e8af6c963c33d6f8540/boards/feather_m0/src/lib.rs#L177
> 
> rather than an entirely separate BSP. Also I don't think we have CAN support in the HAL so I'm not sure how different it would be from the regular feather_m4. If the board is completely different and not compatible with the regular feather_m4, let me know, but I suspect it's just a minor variant.

I've been experimenting with the Adafruit Feather M4 CAN board and found that the NeoPixel example doesn't work. The problem turned out to be that the NeoPixel on the Adafruit Feather M4 CAN board is connected differently than on the Adafruit Feather M4  board, and the power supply to the NeoPixel is controlled by a microcontroller. The ultimate goal of all my experimentation is to run full CAN/CANFD work on this board using Rust.

sajattack:
Yeah I still think you could remap the neopixel and add the CAN pins under a feature flag.

bradleyharden:
I agree with @sajattack. I would strongly prefer not to create a new BSP if it can work as a feature flag. The duplicated code becomes much more of a liability, especially if it gets out of sync with the existing `feather_m4`

ianrrees:
Having spent some time on that Feather M0 BSP in https://github.com/atsamd-rs/atsamd/issues/704, and lately on BSPs/examples more generally in https://github.com/atsamd-rs/atsamd/issues/830, I feel that we should have separate BSPs when there are hardware differences.  That said, I share the concerns about code duplication adding to maintenance burden.

Keen to revisit this once 830 is resolved!

PR #625, opened 2022-07-09
thumbv6m: fix USB clock recovery

lopsided98:
# Summary

USB clock recovery was disabled because it caused USB resets and instability. This occurred because the DFFLMUL.MUL was set to the wrong value. The multiplier must be 48000 to obtain a 48 MHz clock from the 1 kHz USB SOF signal.

There appears to be no harm in unconditionally enabling USB clock recovery when there is no external crystal, since it will just operate in open loop mode (as before) when USB is not connected.

# Checklist

## If Adding a new Board

### DIFF
--- a/hal/src/thumbv6m/clock.rs
+++ b/hal/src/thumbv6m/clock.rs
@@ -560,8 +560,8 @@ fn configure_and_enable_dfll48m(sysctrl: &mut SYSCTRL, use_external_crystal: boo
         sysctrl.dfllmul.write(|w| unsafe {
             w.cstep().bits(coarse / 4);
             w.fstep().bits(10);
-            // scaling factor between the clocks
-            w.mul().bits((48_000_000u32 / 32768) as u16)
+            // scaling factor for 1 kHz USB SOF signal
+            w.mul().bits((48_000_000u32 / 1000) as u16)
         });
 
         // Turn it on
@@ -575,11 +575,7 @@ fn configure_and_enable_dfll48m(sysctrl: &mut SYSCTRL, use_external_crystal: boo
             // chill cycle disable
             w.ccdis().set_bit();
 
-            // usb correction is not set due to instability issues around
-            // USB bus resets. TODO(twitchyliquid64): Maybe switch to OSC8M?
-            //
-            // TODO usb correction (still active for samd11??)
-            #[cfg(feature = "samd11")]
+            // usb correction
             w.usbcrm().set_bit();
 
             // bypass coarse lock (have calibration data)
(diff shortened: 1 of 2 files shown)


TDHolmes:
Thanks for the contribution. Can you go through the checklist? specifically adding a changing entry. Also, is this change valid for `samd11` too? Why were they working fine up til now?

lopsided98:
The SAMD11 datasheet contains the exact same sentence as the SAMD21:

> When the USB device is connected, a SOF will be sent every 1ms, thus DFLLVAL.MUX bits should be written to 0xBB80 to obtain a 48MHz clock.

Note that `DFLLVAL.MUX` doesn't exist, and the only sensible thing they could have meant is `DFLLMUL.MUL`.

Do you know that the SAMD11 actually worked correctly? The comment doesn't seem too certain.

TDHolmes:
I am not sure. @twitchyliquid64 any context?

TDHolmes (approved):
LGTM pending comments from twitchyliquid64

twitchyliquid64:
I barely used this chip but seem to remember its cursed. A quick look at the errata shows only three entries - theres no way thats right, its more likely that everyone else gave up too.

Given the obvious copypasta, its likely they used similar IP for the USB/DFLL peripheral on the d11 as the d21, and I don't see any catastrophic errata entries on the other chips in the series. So this is probably fine.

At the end of the day, if you've got it working fine on real hardware (both with and without USB connected), then I'm okay with this going in. But perhaps mention which revision of the chip you have on this PR?

hal/src/thumbv6m/clock.rs
```diff
@@ -575,11 +575,7 @@ fn configure_and_enable_dfll48m(sysctrl: &mut SYSCTRL, use_external_crystal: boo
             // chill cycle disable
             w.ccdis().set_bit();
 
-            // usb correction is not set due to instability issues around
-            // USB bus resets. TODO(twitchyliquid64): Maybe switch to OSC8M?
```
twitchyliquid64, line 579: Oh I remember this. I was trying to get my USB bus driver to work, and found that everytime the host issued a USB reset then my chip locked up. Never really got to the bottom of it.

If you can make your system reset the USB device and it _doesnt_ lock up, then this comment is stale.

n8tlarsen (approved):
I've also verified this change works in real hardware. The reason for instability comes when trying to cross clock domains. When the DFLL is not locked, no reads or writes to the DFLL clock domain can be completed, which often causes the core to stall.

n8tlarsen:
@lopsided98 What has to happen yet to complete the PR? I'm more than happy to help close this item as I'm working on a project which would benefit from this fix.

bradleyharden:
@n8tlarsen, this looks good to me. A changelog entry is there, and I don't see anything else blocking. It's a pretty simple change. Seems like a pure bug fix.

PR #627, opened 2022-08-23
Fix Wio Terminal ButtonController

### LINKED ISSUE #620: Wio Terminal: `ButtonController` doesn't report release events
Once a button is pressed, `ButtonController` stops reporting events for that button as if it's stuck in the pressed state. This can be observed through the `buttons` example.

This issue seems to have been introduced by commit 34ad122ab.

pabigot:
# Summary
If debounce isn't enabled pinstate always returns zero, so the code treats every event as a down state.

Fixes #620.

### DIFF
--- a/boards/wio_terminal/src/buttons.rs
+++ b/boards/wio_terminal/src/buttons.rs
@@ -1,7 +1,7 @@
 use atsamd_hal::clock::GenericClockController;
 use atsamd_hal::eic;
 use atsamd_hal::eic::pin::{
-    ExtInt10, ExtInt11, ExtInt12, ExtInt3, ExtInt4, ExtInt5, ExtInt7, Sense,
+    ExtInt10, ExtInt11, ExtInt12, ExtInt3, ExtInt4, ExtInt5, ExtInt7, ExternalInterrupt, Sense,
 };
 use atsamd_hal::pac::{interrupt, EIC, MCLK};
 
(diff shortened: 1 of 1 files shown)


boards/wio_terminal/src/buttons.rs
```diff
@@ -41,8 +41,15 @@ impl ButtonPins {
...
 
-        /*
         eic.button_debounce_pins(&[
+            11 as ExternalInterruptID,
+            12 as ExternalInterruptID,
+            3 as ExternalInterruptID,
+            4 as ExternalInterruptID,
+            5 as ExternalInterruptID,
+            10 as ExternalInterruptID,
+            7 as ExternalInterruptID,
```
bradleyharden, line 51: I think I see the problem. The `ExternalInterrupt` trait is not defined in a reasonable way anymore. This is probably a problem introduced by the migration to `gpio::v2`. I think I have a way to fix it, so that you can use `.id()` again. Let me see if I can implement that first, before we resort to manually specifying all the `ExternalInterruptID`s

pabigot: @bradleyharden That'd be great; thanks.  I hope to be back on wio/rust by the end of the week, when I'll take another stab at #628 and can update this too.

pabigot:
Updated now that #641 is in-place.

Issue #629, opened 2022-08-24
Read state of EIC pin


sakian:
It doesn't look like it's possible to read the state of an external interrupt pin after it's been moved into an EIC (e.g. `ExtInt0::new(pin)`). Would it make sense to implement the Embedded HAL `InputPin` trait for the EIC pin?

bradleyharden:
Probably. I'm not that familiar with the EIC API (which varies by chip), but it sounds reasonable. You could make a PR for it.

Although, I have to say, the last time I looked at EIC, I remember thinking that the whole API could use an overhaul. But that's much bigger project.

PR #630, opened 2022-08-25
Added InputPin trait impl to EIC pins

sakian:
# Summary
Adds embedded-hal `InputPin` trait implementation to EIC pins so that you can read the state of an EIC pin. This is useful for when you are sensing both edges and want to read the level to determine which edge was seen. Addresses #629.

# Checklist

### DIFF
--- a/hal/src/thumbv6m/eic/pin.rs
+++ b/hal/src/thumbv6m/eic/pin.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "unproven")]
+use crate::ehal::digital::v2::InputPin;
 use crate::gpio::{
     self, pin::*, AnyPin, FloatingInterrupt, PinId, PinMode, PullDownInterrupt, PullUpInterrupt,
 };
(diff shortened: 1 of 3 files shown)


jbeaurivage:
Looks pretty good to me, one thing I'd mention is that there might not be much need for feature-gating on `unproven`. I believe we're trying to get away from that.

hal/src/thumbv6m/eic/pin.rs
```diff
@@ -146,6 +148,23 @@ crate::paste::item! {
         }
     }
 
+    #[cfg(feature = "unproven")]
+    impl<GPIO, C> InputPin for [<$PadType $num>]<GPIO>
```
bradleyharden, line 152: I think it would be easier if you just used `GPIO: InputPin`

bradleyharden: Wait, no that won't work. We store a `Pin`, not an instance of `GPIO`. This is one of those things done for backwards compatibility. The solution is to just improve the API. But I guess this is fine for now.

jbeaurivage:
Looks pretty good to me, one thing I'd mention is that there might not be much need for feature-gating on `unproven`. I believe we're trying to get away from that.

bradleyharden:
@jbeaurivage, `InputPin` itself is gated on `unproven`, so we can't get rid of it for this case.

bradleyharden:
@sakian, feel free to ping if things sit for a while. This looks good to go.

PR #632, opened 2022-09-18
Add Nina pin aliases to arduino_nano33iot

RobinMcCorkell:
Add pin aliases for the Nina peripheral.

Sercom2 is configured on the necessary pins with pin mode C on all 4 SPI pins (https://ww1.microchip.com/downloads/en/DeviceDoc/SAM_D21_DA1_Family_DataSheet_DS40001882F.pdf)


bradleyharden:
@RobinMcCorkell, please don't delete the checklist. It's also there to help maintainers remember things to double check.

PR #635, opened 2022-10-24
Async API

jbeaurivage:
Provide APIs to use some peripherals using `async/await`.

Design summary can be seen here: https://gist.github.com/jbeaurivage/16955c5f002de4c09fdd6c26a30ef59e

## HAL features
One new feature is added to the HAL: `async`

## MSRV
This PR bumps the HAL MSRV to rustc 1.75. However everything compiles on stable.

## Peripherals implemented
* [x] Timer/Counter (TC)
* [x] I2C
* [x] UART
* [x] SPI
* [x] DMAC
* [x] GPIO interrupts (EIC)

## Todo
* [x] Document everything
* [x] Find best API for interrupt ownership
* [x] Convert to new peripheral-feature system, especially brand new modules

## Examples
Some examples are provided for `feather_m0` (and a few for `metro_m4`). Some more examples will be added in time, perhaps in a follow-up PR.

## Breaking changes

* The `EIC` drivers have a slightly new API.
* Removed wakers for sync `dmac::Transfer`s.


boards/metro_m4/.cargo/config
```diff
@@ -9,6 +9,6 @@ rustflags = [
    # This is needed if your flash or ram addresses are not aligned to 0x10000 in memory.x
    # See https://github.com/rust-embedded/cortex-m-quickstart/pull/95
    "-C", "link-arg=--nmagic",
-
+   "-C", "link-arg=-Tdefmt.x", # uncomment if using defmt
```
sajattack, line 12: should this be commented out?

hal/src/sercom/async_dma.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    #[inline]
+    fn incrementing(&self) -> bool {
+        false
+    }
+
+    #[inline]
+    fn buffer_len(&self) -> usize {
+        1
```
sajattack, line 83: why 1?

boards/metro_m4/.cargo/config
jbeaurivage: Yeah you're right. Fixed now.

hal/src/sercom/async_dma.rs
jbeaurivage: So the SERCOM peripherals expect all data to be read/written to the same register address. In the case of a transmission, it then may add it to an internal buffer, after which the data is transmitted (the reverse happens for receive transactions). Therefore the DMA transfer must direct all the incoming/outgoing data to the same memory address, regardless of the transaction length.

sajattack:
I'm sorry this sat so long. I'm pretty sure my feather m0 board is busted after I wired up a battery to it backwards. Would you mind porting more of the examples to metro m4 so I can get a better feel for this?

jbeaurivage:
> I'm sorry this sat so long. I'm pretty sure my feather m0 board is busted after I wired up a battery to it backwards. Would you mind porting more of the examples to metro m4 so I can get a better feel for this?

Of course! The feather_m0 examples are ported to metro_m4. Unfortunately they aren't yet using `clock::v2` as I'm not super sure how to use it correctly. One example I think will definitely need testing is `async_eic`, I'm not 100% sure about it yet, even though it works well on thumbv6 targets.

boards/feather_m0/examples/async_spi.rs
```diff
@@ -0,0 +1,78 @@
...
+        sclk,
+        mosi,
+        miso,
+    )
+    .into_future(Irqs)
+    .with_dma_channels(channel0, channel1);
+
+    loop {
+        defmt::info!("Sending 0x00 to SPI device...");
+        // This test is based on the BMP388 barometer. Feel free to use any I2C
```
sajattack, line 67: maybe remove this comment, since this isn't i2c

sajattack:
>  One example I think will definitely need testing is async_eic, I'm not 100% sure about it yet, even though it works well on thumbv6 targets.

I tested it. It works great.

```
INFO  Falling edge detected
└─ async_eic::____embassy_main_task::{async_fn#0} @ examples/async_eic.rs:59
INFO  Falling edge detected
└─ async_eic::____embassy_main_task::{async_fn#0} @ examples/async_eic.rs:59
INFO  Falling edge detected
└─ async_eic::____embassy_main_task::{async_fn#0} @ examples/async_eic.rs:59
INFO  Falling edge detected
└─ async_eic::____embassy_main_task::{async_fn#0} @ examples/async_eic.rs:59
INFO  Falling edge detected
```

boards/feather_m0/Cargo.toml
```diff
@@ -37,17 +37,23 @@ version = "0.3"
 optional = true
 
 [dev-dependencies]
-cortex-m-rtic = "1.0"
+embassy-executor = { version = "0.4.0", features = ["arch-cortex-m", "executor-thread", "task-arena-size-64"] }
```
sajattack, line 40: somehow this prevented the metro_m4 from compiling even though it's in an entirely separate bsp.  I changed it to 0.6.2 locally.

boards/metro_m4/examples/async_uart.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+#[embassy_executor::task]
+async fn receive_bytes(mut uart_rx: UartFutureRxDuplexDma<Config<bsp::UartPads>, Ch0>) {
+    uart_rx.as_mut().flush_rx_buffer();
+
+    loop {
+        let mut buf = [0x00; 10];
+        match uart_rx.read(&mut buf).await {
+            Ok(()) => defmt::info!("read {}", &buf),
```
sajattack, line 103: should this be hex or ascii formatted perhaps?

boards/metro_m4/examples/async_dmac.rs
```diff
@@ -0,0 +1,74 @@
...
+    // Get individual handles to DMA channels
+    let channels = dmac.split();
+
+    // Initialize DMA Channel 0
+    let mut channel = channels.0.init(PriorityLevel::Lvl0);
+
+    let mut source = [0xff; 1000];
+    let mut dest = [0x0; 1000];
+
+    defmt::info!(
```
sajattack, line 49: during my testing I did not see these prints just wfi

boards/feather_m0/examples/async_dmac.rs
```diff
@@ -0,0 +1,73 @@
...
+        .unwrap();
+
+    defmt::info!(
+        "Finished DMA transfer.\n\tSource: {}\n\tDestination: {}",
+        &source,
+        &dest
+    );
+
+    loop {
+        cortex_m::asm::wfi();
```
sajattack, line 71: should this be `core::hint::spin_loop`? genuinely not sure just curious

boards/feather_m0/Cargo.toml
jbeaurivage: Yeah looks like `0.4.0` was yanked. Fixed now

boards/feather_m0/examples/async_dmac.rs
jbeaurivage: I don't think it really matters; the idea is just that WFI should help reduce power consumption while idle. I can just as well change it to a `spin_loop` if needed.

sajattack: yeah makes sense. It doesn't matter and the power saving is good. I wonder if we should have both? but meh

jbeaurivage: Nah, the spin_loop hint is only necessary to tell the compiler not to optimize the loop out. As soon as there's at least one instruction in there, it's no longer necessary.

boards/metro_m4/examples/async_dmac.rs
jbeaurivage: Yeah, there was some unsoundness around DMA transfers. Hopefully should be fixed with #780, on top of which this PR is rebased.
I'd love to know if you're still observing that bug now

jbeaurivage:
Note to reviewers: please use "Rebase and merge" to preserve the git history.

kaspar030:
:tada:

PR #639, opened 2022-11-12
Add RTIC edge counter + EIC application for ATSAME54 Xplained Pro

johngigantic:
# Summary
Added an RTIC example which uses an interrupt pin. Tested and verified operation on an ATSAME54 Xplained Pro board. Also modified the ethernet interrupt pin to be a pull-up interrupt as a side note.

#Issues
I was unable to get the eic::pin::id() -> ExternalInterruptId function working, and instead hard-coded the ID value in.

# Checklist


johngigantic:
See issue #640. This example works, compiles, and runs, although I wasn't able to get the ExternalInterruptID value without hardcoding the value in.

PR #647, opened 2022-12-04
Make wio_terminal display frequency consistent

pabigot:
58 MHz works on all examples.

### DIFF
--- a/boards/wio_terminal/CHANGELOG.md
+++ b/boards/wio_terminal/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Unreleased
 
 - Fix display frequency in examples that did not work
+- Fix display offsets in buttons example
+- Fix buttons by re-enabling debounce code
+- Fix various documentation and clippy warnings
+- Fix Wi-Fi by delaying UART init until device reset
 
 # v0.6.1
 
--- a/boards/wio_terminal/CHANGELOG.md
+++ b/boards/wio_terminal/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix display frequency in examples that did not work
+
 # v0.6.1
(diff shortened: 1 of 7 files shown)


pabigot:
Some background adding on to https://github.com/atsamd-rs/atsamd/pull/604#issuecomment-1099634925:

Support for selecting display SPI speed was added in #320 (2020-11-15), where these were set to 58 MHz:

    buttons, clock, orientation, usb_serial_display

and these to 24 MHz:

    qspi, sdcard

possibly because they also use SPI for storage access.  These examples were added after #320:

    microphone, wifi_connect, wifi_scan

all of which picked 24 MHz.

The ILI9341 datasheet suggests that the maximum speed would be 10 MHz, but as noted in #320 as much as 80 MHz has been confirmed to work.

The display does not work for wifi_scan at 24 MHz; it does at 32 MHz, 40 MHz, and 58 MHz.  The other examples work at 58 MHz speed too, so make it consistent to reduce confusion in users creating new demos.

jbeaurivage:
Could you please update the BSP's README?

pabigot:
@jbeaurivage Rebased and README updated.  I'm not sure what you were looking for, let me know if this isn't it.

jbeaurivage:
@pabigot sorry my mistake, I meant updating the changelog! But the info you put in the readme definitely doesn't hurt

pabigot:
@jbeaurivage OK, that makes sense.  Should I add entries for my other recent fixes too?

jbeaurivage:
Yes please! No worries, all the entries can be in this pr.

pabigot:
@jbeaurivage done

jbeaurivage:
Thanks for all the improvements @pabigot!

PR #650, opened 2022-12-19
Remove const_err lint

pabigot:
As of Rust 1.66 this produces:
```
warning: lint `const_err` has been removed: converted into hard error, see issue #71800 <https://github.com/rust-lang/rust/issues/71800> for more information
 --> /mnt/devel/external/Rust/github.com/atsamd-rs/atsamd/pac/atsamd51p/src/lib.rs:3:9
  |
3 | #![deny(const_err)]
  |         ^^^^^^^^^
  |
  = note: `#[warn(renamed_and_removed_lints)]` on by default
```

The problem was fixed in [svd2rust 0.27.0](https://github.com/rust-embedded/svd2rust/blob/master/CHANGELOG.md#v0270---2022-10-24)  but using 0.27 in `update-pacs.sh`  breaks everything, so this patch can be applied to remove the problem until the more general solution can be applied.


bradleyharden:
@pabigot, did you happen to notice what the major problems were with the updated PACs? It would be nice to update. We should try not to fall too far behind.

pabigot:
@bradleyharden My recollection was I changed `update-pacs.sh` to use svd2rust 0.27.2 and builds produced cascading errors, though I can't replicate that now.

As of yesterday current svd2rust is at 0.28.0.  One point is that per the [CHANGELOG](https://github.com/rust-embedded/svd2rust/blob/master/CHANGELOG.md) MSRV for pacs would increase to 1.61 at 0.27.0.

I've tried builds in `wio_terminal` and `feather_m0` using 0.28.0, which also required changing `update-pac.sh` `svdrust --nightly` to `svdrust --atomics`   I get conflicting definitions of `CHANNEL` as a struct and as a type in `pac/atsmd51p/src/evsys/channel.rs`, and use of unresolved import `portable_atomic` in `atsamd/pac/atsamd21g/src/generic/atomic.rs` which continues to produce errors after the package is added.

So it shouldn't be as bad as I thought, but there's some weirdness going on.

bradleyharden:
Hmm, ok. We should probably prioritize upgrading. If we continue to let it go, it will only get harder to catch up in the future. Maybe I'll take a look myself this week.

bradleyharden:
Also, we bumped the MSRV to 1.65 with the merge of #450, because it uses GATs, so that's not an issue.

PR #654, opened 2022-12-26
Integration with the `mcan`

glaeqen:
This PR provides trait implementations from [`mcan-core`](https://crates.io/crates/mcan-core) crate. It enables the safe usage of the [`mcan`](https://crates.io/crates/mcan) crate.

`mcan` crate provides a HAL for the CAN peripheral embedded into ATSAME5{1, 4}.

### DIFF
--- a/boards/atsame54_xpro/Cargo.toml
+++ b/boards/atsame54_xpro/Cargo.toml
@@ -26,6 +26,7 @@ version = "0.2"
 optional = true
 
 [dev-dependencies]
+mcan = "0.2"
 cortex-m-rtic = "1.1"
 dwt-systick-monotonic = "1.1"
 panic-rtt-target = { version = "0.1", features = ["cortex-m"] }
@@ -36,6 +37,7 @@ default = ["rt", "atsamd-hal/same54p", "atsamd-hal/unproven"]
 rt = ["cortex-m-rt", "atsamd-hal/same54p-rt"]
 unproven = ["atsamd-hal/unproven"]
 usb = ["atsamd-hal/usb", "usb-device"]
+can = ["atsamd-hal/can"]
 
 [profile.dev]
 incremental = false
@@ -46,3 +48,7 @@ lto = true
 [profile.release]
 lto = true
 opt-level = "s"
+
+[[example]]
+name = "mcan"
+required-features = ["can"]
--- a/boards/atsame54_xpro/examples/mcan.rs
+++ b/boards/atsame54_xpro/examples/mcan.rs
@@ -1,0 +1,317 @@
+//! MCAN example
+//! Assumed bus:
+//! - 375 kb/s nominal bitrate
+//! - 750 kb/s data bitrate if sending/receiving CAN FD frames with bitrate
+//!   switching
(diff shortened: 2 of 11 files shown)


glaeqen:
I'll try to provide a similar example to the one in https://github.com/atsams-rs/atsamx7x-rust/pull/51 soon.

sajattack:
My only comment would be that the example is a little long and complicated, but I suppose that's just the nature of the beast. Overall, LGTM

glaeqen:
> My only comment would be that the example is a little long and complicated, but I suppose that's just the nature of the beast. Overall, LGTM

I could probably cut it down to something simpler, I wanted it to be a little more fun, thorough and interactive. Like, listen to messages on interrupt and print them out AND send a message on a button press. It also contains the filter setup which brings the value of an example up. Anyone who wants to use it can drop whatever they deem unnecessary - this is how I thought about it

glaeqen:
> oh, one more thing, please add the can feature to crates.json for appropriate PACs, so that it is tested.

Done, not sure about these `hal_doc_variants` though. Are they supposed *not* to include sam*E* chips? Only sam*D*?

sajattack:
> > oh, one more thing, please add the can feature to crates.json for appropriate PACs, so that it is tested.
> 
> Done, not sure about these `hal_doc_variants` though. Are they supposed _not_ to include sam_E_ chips? Only sam_D_?

Sounds like a mistake on my part. Feel free to fix here or in a followup, at your discretion.

sajattack:
Something weird going on here. Maybe a previous PR mixed up the features in hal/Cargo.toml? 
https://github.com/atsamd-rs/atsamd/actions/runs/4348898859/jobs/7597968312#step:5:86

glaeqen:
> > > oh, one more thing, please add the can feature to crates.json for appropriate PACs, so that it is tested.
> > 
> > 
> > Done, not sure about these `hal_doc_variants` though. Are they supposed _not_ to include sam_E_ chips? Only sam_D_?
> 
> Sounds like a mistake on my part. Feel free to fix here or in a followup, at your discretion.

#677, feel free to comment what do you think. Maybe I completely misunderstood the idea behind how it worked before.

glaeqen:
> Something weird going on here. Maybe a previous PR mixed up the features in hal/Cargo.toml? https://github.com/atsamd-rs/atsamd/actions/runs/4348898859/jobs/7597968312#step:5:86 @bradleyharden PTAL
> 
> https://github.com/atsamd-rs/atsamd/blob/d8d85b01e11a7a0b21bcf672f241578a5aad5cd0/hal/Cargo.toml#L123

FTR: Covered by #676, already rebased.

sajattack:
What do you think about calling out the example in the atsame54_xpro BSP changelog? I could go either way.

glaeqen:
> What do you think about calling out the example in the atsame54_xpro BSP changelog? I could go either way.

I wasn't sure about that. I'm not changing the BSP there if one ignores extra feature (forwarded to the HAL)? I guess that's the only thing that is worth being mentioned.

Issue #574, opened 2022-02-21
Consider running clippy more aggressively


TDHolmes:
After #569, it would be good to run clippy more aggressively in CI. We currently run it but don't block anything on the result. I think it makes sense to deny errors everywhere and deny warnings in the HAL and tier one BSPs

bradleyharden:
What is needed to implement this? Do I still owe some clippy fixes? I think so

TDHolmes:
I recall there still being some missing safety comments on unsafe pieces of the SPI module

PR #655, opened 2022-12-30
CI update

sajattack:
# Summary
- rework crates.json a bit
- build the HAL for every PAC with every relevant feature
- remove deprecated actions-rs workflows
- I think clippy errors will deny now but warnings are still allowed, let me know if we should change that

# Checklist

## If Adding a new Board

Let me know if there are any more improvements desired, I held off on building a docker image, because I wasn't sure how to best keep it up to date.

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Install Rust
         run: |
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -13,20 +13,14 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v2
-      - name: Install Rust (thumbv6m)
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-          target: thumbv6m-none-eabi
-      - name: Install Rust (thumbv7em)
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-          target: thumbv7em-none-eabihf
+
+      - name: Install Rust
+        run: |
+          rustup set profile minimal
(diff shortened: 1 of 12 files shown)


sajattack:
@TDHolmes I made some changes related to #574 , but didn't go all the way with denying warnings yet, let me know what you think.

.github/workflows/generate-docs.yml
```diff
@@ -12,21 +12,15 @@ jobs:
...
-          target: thumbv7em-none-eabihf
+        uses: actions/checkout@v3
+
+      - name: Install Rust
+        run: |
+          rustup set profile minimal
+          rustup override set ${{ matrix.toolchain }}
+          rustup target add thumbv6m-none-eabi
+          rustup target add thumbv7em-none-eabihf
+          rustup component add clippy
```
bradleyharden, line 23: Do we need clippy when building the docs?

sajattack: Oops, probably a copy-paste mistake

crates.json
```diff
@@ -2,199 +2,318 @@
   "boards": {
     "arduino_mkr1000": {
       "tier": 2,
-      "build": "cargo build --examples --all-features"
+      "build": "cargo build --examples --all-features",
+      "target": "thumbv6m-none-eabi"
```
TDHolmes, line 6: It's a bummer we need to specify the target here too since it is also in the cargo.toml of the BSP, right?

TDHolmes:
@sajattack nice! I wish there was a middle ground between denying warnings and not doing anything, but last I checked, the commenting didn't work, not to mention `actions-rs` is deprecated

crates.json
sajattack: I just put it there for convenience. It prevents us from installing both targets for a board that's only going to be using one or the other

TDHolmes: Is it possible to parse from the cargo.toml? Agreed it's good to only install the one that's needed. If it's difficult not a problem to keep things as is

bradleyharden:
I skimmed the changes, but it's clear to me that I don't know enough about GHA to really understand what's going on. It sounds like there could be more improvements, but since everything is green right now, and this PR is blocking further PRs, I'm going to merge. If we want to make further changes, let's do it with a separate PR.

PR #656, opened 2022-12-31
Refactor Cargo features

bradleyharden:
# Summary

- Refactor the Cargo features to provide a better, more fine-grained structure.
- Separate the features for selecting pins from the features for selecting peripherals.
- Reorganize and better document the Cargo.toml file.

**NOTE:** Reviewers will probably want to read through the entirety of `Cargo.toml`, but the rest of the changes will probably be overwhelming to review. Everything builds, but none of this is tested on hardware, since I don't have any hardware to work with at the moment.

# Checklist
  - [ ] ~~`CHANGELOG.md` for the BSP or HAL updated~~


hal/Cargo.toml
```diff
@@ -1,227 +1,342 @@
...
-    "Jesse Braham <jesse@beta7.io>",
+  "Wez Furlong <wez@wezfurlong.org>",
+  "Paul Sajna <sajattack@gmail.com>",
+  "Michael van Niekerk <mvniekerk@gmail.com>",
+  "Jesse Braham <jesse@beta7.io>",
+  "Bradley Harden <bradleyharden@gmail.com>",
 ]
+categories = ["embedded", "hardware-support", "no-std"]
 description = "HAL and Peripheral access API for ATSAMD11, ATSAMD21, ATSAMD51, ATSAME51, ATSAME53 and ATSAME54 microcontrollers"
+documentation = "https://docs.rs/crate/atsamd-hal/"
```
sajattack, line 15: should this point to github.io instead?

hal/Cargo.toml
```diff
@@ -1,227 +1,342 @@
...
-[dependencies.embedded-sdmmc]
-version = "0.3"
-optional = true
+# The peripheral access crates (PAC) for each supported chip variant are listed
+# here. Each is an optional dependency, which means each PAC name also appears
+# as a Cargo feature. Users should not select PAC features manually. Instead,
+# users should specify a corresponding variant (see below). The variant features
+# will select the correct PAC, as well as other configuration features.
 
+atsamd11c = {version = "0.12", path = "../pac/atsamd11c", optional = true}
```
sajattack, line 70: I'm not sure whether the bump-crates workflow will like us changing the format of these, I guess we'll find out.

hal/src/lib.rs
```diff
@@ -97,29 +80,23 @@ pub mod time;
...
-#[cfg(all(feature = "usb", feature = "samd11"))]
-compile_error!("'usb' is enabled, but USB isn't supported on SAMD11");
-
-#[cfg(all(
-    feature = "usb",
-    not(any(feature = "samd21", feature = "min-samd51g", feature = "library"))
-))]
-compile_error!("The 'usb' feature is enabled, but not a chip with USB support");
+#[cfg(all(feature = "usb", feature = "device", not(feature = "usbp")))]
+compile_error!("The 'usb' feature is enabled, but this chip does not support USB");
```
sajattack, line 87: we should have one of these for sdmmc and other stuff too maybe?

bradleyharden:
@sajattack had a suggestion on Matrix:

> maybe some of the gpio features could be called min-pb00 instead of just pb00
> just thought of that, since I was a bit confused until I read the comment

I responded, but I realized that discussion should really stay in the PR, so I'm copying my response here.

@sajattack, here was my thought process for feature naming:

- I wanted to create feature "namespaces" using prefixes. That's how I came up with pins- and periph-.
- However, using a periph- prefix for every single peripheral is extremely verbose, e.g. periph-sercom0, periph-tcc0, etc..
- For peripherals with no possible confusion with a user feature, I just named it the most natural thing, e.g. sercom0, tcc0, etc.
- For peripherals with a possible name collision, I appended a p to the user-selectable version, e.g. usb and usbp.
- I originally had the pb00 feature within the pins- namespace as pins-pb00. But I thought a specific pin is akin to a specific peripheral, which does not have the namespace prefix, so I removed it.

Questions for you:

- Do you have any broad feedback on this scheme?
- What was the min- prefix supposed to convey in your suggestion?
- Do you think I should change back to pins-pb00? Or use something else, like min-pb00?
- If I change pb00, should I also change the peripheral features?

hal/src/lib.rs
bradleyharden: Yes, absolutely. I'll see what other instances make sense.

hal/Cargo.toml
bradleyharden: I was wondering why the PAC dependencies were formatted that way. I found the spread out version to be very difficult to parse. I personally find it much easier to read when they are condensed to single lines.

However, I didn't consider that the specific syntax of the TOML file might be depended on elsewhere. That seems extremely brittle, though. Maybe we can find a better way? I found a [toml-cli](https://github.com/gnprice/toml-cli) project. Could we use that instead?

hal/Cargo.toml
bradleyharden: My first instinct is no, because the github.io docs are built from `master`, which is not on crates.io. But as you said yesterday, we should find a way to make those docs more visible.

I guess the bigger question is which set of docs should new users encounter first? A version tied to a specific crates.io release? Or a version built directly from `master`?

bradleyharden:
I just had a philosophical realization. You say:

> I was a bit confused until I read the comment

Several times recently, both at work and outside of work, people have effectively told me:

> This wasn't intuitive to me, and it didn't make sense until I read the documentation.

While I agree that we should try to make things as intuitive as possible, sometimes things _can't_ be made intuitive, either because the motivation is too complicated & subtle, or because people's intuitions don't match. At some point, I think the solution **has** to be "Read the documentation". But exactly where that happens is debatable.

Where do you think this situation falls? Would adding `min-` make things more intuitive to everyone? Or is it only intuitive if you're already familiar with our prior `min-samxxx` features?

sajattack:
> Questions for you:
> 
>     * Do you have any broad feedback on this scheme?
Not really, it seems reasonable enough 
>     * What was the min- prefix supposed to convey in your suggestion?
Similar to min-samd51g, it conveys that it includes pb00, and everything up to it
>     * Do you think I should change back to pins-pb00? Or use something else, like min-pb00?
Honestly I'm not too fussed about it, it was just an idea that came to me 
>     * If I change pb00, should I also change the peripheral features?
The -p ones are a little unintuitive as well

sajattack:
> I just had a philosophical realization. You say:
> 
> > I was a bit confused until I read the comment
> 
> Several times recently, both at work and outside of work, people have effectively told me:
> 
> > This wasn't intuitive to me, and it didn't make sense until I read the documentation.
> 
> While I agree that we should try to make things as intuitive as possible, sometimes things _can't_ be made intuitive, either because the motivation is too complicated & subtle, or because people's intuitions don't match. At some point, I think the solution **has** to be "Read the documentation". But exactly where that happens is debatable.

Part of the problem in this instance, is that, in doing a full code review, I see `pb00` in multiple places, but the documentation is only in one of them. That's where an intuitive name has benefits over documentation.

bradleyharden:
> The -p ones are a little unintuitive as well

Yeah, I'm not a huge fan of it either, but I didn't come up with anything better at the time.

> Similar to min-samd51g, it conveys that it includes pb00, and everything up to it

My hesitance comes from the fact that this statement isn't quite accurate. Pins aren't strictly additive. The best example is pin PA28, which I had to make special accommodations for. PA28 is included in the 48 and 64 pin packages for D21 chips, but it's not present in the same packages for D51 chips.

While thinking about your responses, I had an idea. What do you think of using a `has-` prefix for all individual pin & peripheral features, e.g. `has-sercom0`, `has-tcc0`, `has-pb00`, etc. It's more verbose, but I think it communicates exactly what is meant by the feature. Then, the `pins-` and `periph-` namespaces would exist to group `has-` features into reusable sets.

sajattack:
> > The -p ones are a little unintuitive as well
> 
> Yeah, I'm not a huge fan of it either, but I didn't come up with anything better at the time.
> 
> > Similar to min-samd51g, it conveys that it includes pb00, and everything up to it
> 
> My hesitance comes from the fact that this statement isn't quite accurate. Pins aren't strictly additive. The best example is pin PA28, which I had to make special accommodations for. PA28 is included in the 48 and 64 pin packages for D21 chips, but it's not present in the same packages for D51 chips.
> 
> While thinking about your responses, I had an idea. What do you think of using a `has-` prefix for all individual pin & peripheral features, e.g. `has-sercom0`, `has-tcc0`, `has-pb00`, etc? It's more verbose, but I think it communicates exactly what is meant by the feature. Then, the `pins-` and `periph-` namespaces would exist to group `has-` features into reusable sets.

Sounds like a good idea.

bradleyharden:
> `has-` eliminates both problems in my mind, so I'm not sure why your last sentence mentions `pins-` and `periph-`

The `pins-` and `periph-` prefixes would still exist to group together other features. The `periph-` features would only exist in the `Cargo.toml` file, so there shouldn't be any confusion there. But the `pins-` features would be in the code as well (as they are right now). Since you didn't comment on the `pins-` features, I take it they were self-explanatory enough.

I'll rename the features and push an update.

hal/Cargo.toml
bradleyharden: You're right, this will break the workflow. Let me look at updating it. Maybe there's a less brittle way to do this.

bradleyharden: It looks like the `tomlkit` package can parse, modify and dump TOML files while preserving comments and formatting. Let me see if I can modify the scripts to use it.

bradleyharden: @sajattack, I updated the two scripts I found that depend on the formatting of the TOML files. Are those the only two scripts you know of? They are the only two I could find.

Could you have a look at the modifications and make sure I didn't accidentally change the behavior?

sajattack (approved):

bradleyharden:
@sajattack, the scripts I modified were definitely very brittle. They used something that was TOML-aware for reading fields, but they still modified the contents of the file by hand. I'm not quite sure why. My guess is that the Python `toml` library doesn't preserve style and comments, so a round-trip read & write would not yield the same file. The new TOML library I used is fully capable of a round-trip without modifications.

I guess we'll just see what happens. If we run into errors when trying to release, we can fix things then.

PR #661, opened 2023-01-05
Fix Alternate for USB pins in BSP for boards that use SAMD51

jboynes:
The SAMD51 uses alternate `H` for USB functionality. This PR updates the pin definitions in the BSP for some boards using the SAMD51 that incorrectly model this as `AlternateG`.

This does not affect the [default](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/thumbv7em/usb/mod.rs#L16-L23) in the HAL, or the [implementation](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/thumbv7em/usb/bus.rs#L531) in `UsbBus` that always uses `AlternateH`

### DIFF
--- a/boards/feather_m4/CHANGELOG.md
+++ b/boards/feather_m4/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Use correct alternate for USB (#661)
+
 # v0.10.1
 
 - Update to `atsamd-hal` version `0.15.1`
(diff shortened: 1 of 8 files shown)


bradleyharden:
@jboynes, could you add notes to the respective changelogs?

jboynes:
> @jboynes, could you add notes to the respective changelogs?
@bradleyharden done

bradleyharden:
I'm not sure why CI failed. It looks unrelated. I can't figure out how to re-run, and I don't think there's any way any of these changes could cause a problem, so I'm going to merge. I'm pretty sure I saw this was green in an earlier commit anyway.

jboynes:
I took a quick look at the log and it looks like GitHub 429'd the requests to check out. I suspect some kind of rate limit on the action runner.

It looks it passed after you merged. Thank you,

PR #671, opened 2023-02-22
Add I2C slave mode to fix #636

### LINKED ISSUE #636: I2C slave mode support
What would be needed to support i2c slave mode? It seems like there is a lot of overlap with i2c master, but I am not sure what the structure would look like.

Register summary is in section 36.7 of the datasheet

tgross35:
# Summary

Add `I2cClient` and associated types to allow using I2C as a slave/client

# Checklist

## If Adding a new Board

N/A

N/A

### DIFF
--- a/hal/src/sercom/i2c.rs
+++ b/hal/src/sercom/i2c.rs
@@ -276,8 +276,8 @@ pub use config::*;
 
 mod impl_ehal;
 
-mod client;
-pub use client::*;
+// mod client;
+// pub use client::*;
 
 /// Word size for an I2C message
 pub type Word = u8;
@@ -296,18 +296,30 @@ pub enum InactiveTimeout {
     Us205 = 0x3,
 }
 
+pub trait I2cMode {}
+/// Marker type for I2C master mode
+pub struct Master;
+/// Marker type for I2C client/slave mode
+pub struct Sleve;
+impl I2cMode for Master {}
+impl I2cMode for Sleve {}
+
 /// Abstraction over a I2C peripheral, allowing to perform I2C transactions.
-pub struct I2c<C: AnyConfig> {
+pub struct I2c<C: AnyConfig<Mode = M>, M: I2cMode = Master> {
     config: C,
 }
 
-impl<C: AnyConfig> I2c<C> {
+/// Implementation for both modes
+impl<C: AnyConfig<Mode = M>, M: I2cMode> I2c<C, M> {
     /// Obtain a pointer to the `DATA` register. Necessary for DMA transfers.
     #[inline]
     pub fn data_ptr(&self) -> *mut Word {
         self.config.as_ref().registers.data_ptr()
     }
+}
 
+/// Host-only implementation
+impl<C: AnyConfig<Mode = Master>> I2c<C, Master> {
     /// Read the interrupt flags
     #[inline]
     pub fn read_flags(&self) -> Flags {
--- a/hal/src/sercom/i2c.rs
+++ b/hal/src/sercom/i2c.rs
@@ -276,6 +276,9 @@ pub use config::*;
 
 mod impl_ehal;
 
+mod client;
+pub use client::*;
+
 /// Word size for an I2C message
 pub type Word = u8;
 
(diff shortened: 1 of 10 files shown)


tgross35:
I'm currently just working on the very basic structure. What would be the best way to lay this out - all types in the existing I2C module with a prefix, everything in an entirely separate module?

jbeaurivage:
I would recommend trying to reuse the existing code rather than copy-pasting into a new module to avoid code duplication. Most likely a new generic parameter to the `Config` and `I2c` structs would create the distinction between master and slave. The master-specifc and slave-specific methods could always go in their own modules if you wish.

tgross35:
Awesome, thanks for the suggestion. That does sound much cleaner, but I was unsure about breaking the existing type API

jbeaurivage:
> Awesome, thanks for the suggestion. That does sound much cleaner, but I was unsure about breaking the existing type API

You can always add the master/slave type at the end, and provide a default type. This will prevent breaking the API in most cases:
```
pub struct Config<P, M = Master>
where
    P: PadSet,
    M: OpMode
{
    pub(super) registers: Registers<P::Sercom>,
    pads: P,
    op_mode: M
    freq: Hertz,
}
```

tgross35:
Alright, I think I got it incorporated it into the type system - no new implementation yet, just a 1:1 change. Would you be able to take a look and see if it seems right?

`AnyConfig` was my only concern, since it needs a `Mode` type and associated type defaults aren't stable, so I believe this would be breaking for anyone implementing that specific trait themselves. Which I think is probably unlikely, but unsure how that affects things.

hal/src/sercom/i2c.rs
```diff
@@ -293,18 +296,30 @@ pub enum InactiveTimeout {
     Us205 = 0x3,
 }
 
+pub trait I2cMode {}
```
jbeaurivage, line 299: I would make this trait sealed so that it can't be implemented outside of this crate:
```
pub trait I2cMode: Sealed {}
```
and also add a short doc comment.

hal/src/sercom/i2c.rs
```diff
@@ -276,6 +276,9 @@ pub use config::*;
 
 mod impl_ehal;
 
+// mod client;
```
jbeaurivage, line 279: I would try to standardize what terminology we use: master/slave vs host/client. Personally I don't care either way, with maybe a slight preference towards whatever the datasheet uses.

hal/src/sercom/i2c/config.rs
```diff
@@ -253,31 +258,33 @@ impl<P: PadSet> Config<P> {
 pub trait AnyConfig: Is<Type = SpecificConfig<Self>> {
     type Sercom: Sercom;
     type Pads: PadSet<Sercom = Self::Sercom>;
+    type Mode: I2cMode; // default to super::Master when associated_type_defaults is stable
```
jbeaurivage, line 261: I don't think `AnyConfig` is meant to be implemented outside of the HAL anyways.

jbeaurivage (commented):
Looks good so far, I haven't looked at the copy-pasted modules, as I'm assuming they're going away, and only the master/slave specific behavior will be broken out?

tgross35:
> Looks good so far, I haven't looked at the copy-pasted modules, as I'm assuming they're going away, and only the master/slave specific behavior will be broken out?

Thanks for the quick look! Yeah they will go away, I just need to go through and sort things into master/slave/both impls

hal/src/sercom/i2c.rs
tgross35: Thanks for bringing this up, I meant to ask. The datasheet does use host/client, so I can go ahead with that.

hal/src/sercom/i2c/config.rs
tgross35: Oh good point, `Sercom` and `Padset` are sealed so it's de facto sealed anyway. I'll make that explicit just so it's also clear in the docs.

ZettaScript:
Are you still working on this? As I need this feature I started rebasing this branch (lots of conflicts after 2 years!) and try to make it work.

tgross35:
> Are you still working on this? As I need this feature I started rebasing this branch (lots of conflicts after 2 years!) and try to make it work.

I never did get the chance to finish this up unfortunately, and eventually my need went away. So please feel free to pick this up, thank you for doing so!

PR #674, opened 2023-02-24
Update NVM API

vccnfitingh:
# Summary
- Extend the `Userpage` bitfield to cover the whole user row (instead of only the first 128 bits) and make the methods public.
- Add the ability to write the user row from an application running on the chip.
- Add support for enabling the NVM security bit (disallows external debug probe access except for chip erase).
- Add support for chip-erase lock and unlock.
- Expose the raw `NVMCTRL` registers to reduce the severity of workarounds when using features that are not yet supported by the HAL
- Refactor `Nvm::command_sync` to be less error-prone
- Add `Nvm::region_lock` method

To support the user row changes, some of the NVM read/write/erase methods were changed.
- Erasing the user row only makes sense as part of a read-modify-erase-write as provided by `modify_userpage`, so `erase_flash` (formerly just `erase`) can always use the only erase granularity that is valid for main flash (`Block`).
- The chip family supports writing to main flash with different granularities, so `WriteGranularity` is added to the arguments of `write_flash` and its `_from_slice` counterpart.

# Checklist


hal/src/thumbv7em/nvm.rs
```diff
@@ -305,64 +414,148 @@ impl Nvm {
...
+            Ok(())
         }
     }
 
-    /// Write to flash memory from a slice
+    /// Enable/disable region lock
     ///
-    /// # Safety
+    /// Flash memory is split into 32 regions. `u32` based mask allow to enable
+    /// (if bit is `1`) or disable (if bit is `0`) region lock in a given
```
glaeqen, line 433: I think this condition should be inverted, bit 0 is region lock enabled. My bad.

jbeaurivage:
Sounds good, that's something that could be looked at in a subsequent PR.

PR #677, opened 2023-03-07
Unify hal_(doc|build)_properties

glaeqen:


jbeaurivage:
@glaeqen, if you want to sync this branch to master, I'll be happy to merge.

glaeqen:
Hey, I must say that I got a little out of a loop with atsamd. I tried to rebase it once a few months ago but had a hard time understanding a new feature system. So my motivation somewhat waned 😅 Could try again though.

glaeqen:
`generate-docs.yml` workflow should be exercised first before we can say this is mergable.

sajattack:
> `generate-docs.yml` workflow should be exercised first before we can say this is mergable.

hmm, can you try it on a fork? I think if I do it here it will push and maybe mess up prod?

PR #683, opened 2023-05-14
Update PAC's to svd2rust 0.30.2

### LINKED ISSUE #648: PACs need to be regenerated to remove lint diagnostic
With Rust 1.66 all pacs produce:
```
    warning: lint `const_err` has been removed: converted into hard error, see issue #71800 <https://github.com/rust-lang/rust/issues/71800> for more information
     --> /mnt/devel/external/Rust/github.com/atsamd-rs/atsamd/pac/atsamd51p/src/lib.rs:3:9
      |
    3 | #![deny(const_err)]
      |         ^^^^^^^^^
      |
      = note: `#[warn(renamed_and_removed_lints)]` on by default
```

svdrust v0.27.0 or later is needed to remove this lint from the generated code.

Tremoneck:
# Summary
Update the PAC to use svd2rust 0.28.0 and form 0.10.0, while fixing up all the renamed registers.
svd2rust renamed severall registers from *_A to *SELECT_A, these were fixed up. This is a breaking change for dependent crates accessing registers, structs or enums directly.

Two patches were neccessary to make the generated code correct. 
- For the SAM_5 series the EVSYS generated a struct and a register both called CHANNEL. The register was renamed to CHANNELCTRL to remove the name conflict
- For some of the SAMD2 series a type is named differently between usage and definition. This was fixed manually and the corresponding issue is https://github.com/rust-embedded/svd2rust/issues/713

This pull request isn't finished as is, but requires some feedback:
For the SAM_5 should the register be renamed or the struct representing a channel?
Should we wait for a fix to land for https://github.com/rust-embedded/svd2rust/issues/713, go back to 0.27.0 or go with this manual patch?
The new generated code relies on a feature on critical-section. Should this be default and always available, or should it, like it is in this PR be a optional feature which can be enabled over the hal?

Fixes #648
Obsoletes #650

# Checklist


jbeaurivage:
@Tremoneck, thank you for this PR! I've had something similar but half finished for a while. Nice to see someone else step up to take it on.

To answer your questions:
> For the SAM_5 should the register be renamed or the struct representing a channel? 

I think logically the register should be renamed. Looks like an oversight on Atmel's part, as most registers are suffixed with *CTRL. That being said, I don't particularly like that the PACs now diverge from the datasheet. 

> Should we wait for a fix to land for rust-embedded/svd2rust#713, go back to 0.27.0 or go with this manual patch?

That depends. Do you expect rust-embedded/svd2rust#713 to end up generating the same code as your manual patch? Ie, will we have breaking changes when a fix lands and we re-generate the PACs in the future?,

Tremoneck:
I don't like the divergent from the Datasheet either. In the [Datasheet](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ProductDocuments/DataSheets/SAM-D5x-E5x-Family-Data-Sheet-DS60001507.pdf) (p. 792) the register is titled _Channel n Control_ but the name is given as the value in the SVD of CHANNEL. The alternative is calling the struct something along the lines of CHANNEL_REGISTERS. 

I expect the register to end up generating to the value I fixed them to. It's the same type name as the 0.27 would generate. So even if a fix changes the typename, we would still have the same breaking change.

jbeaurivage:
> I don't like the divergent from the Datasheet either. In the [Datasheet](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ProductDocuments/DataSheets/SAM-D5x-E5x-Family-Data-Sheet-DS60001507.pdf) (p. 792) the register is titled _Channel n Control_ but the name is given as the value in the SVD of CHANNEL. The alternative is calling the struct something along the lines of CHANNEL_REGISTERS.

What about calling the struct `CHANNELS`?

In any case, I'm on board with the manual patch while awaiting the `svd2rust` fix, especially if any migration steps are properly documented. Let me know if you have any more questions. Obviously new PAC/BSP/HAL releases will be necessary when this PR is merged.

Tremoneck:
Where should the breaking changes be documented? Should that be in the changelog on how to migrate to the next version.

I've added a new feature to the hal, which is only needed if you want to use the take for Peripherals. Where should this be documented? I really would like to add an example for when you need it and what else you need so you don't get compiler errors.

When everything is ready, should I rebase and squash the commits and add a longer explanation into the commit message with the required things to do to migrate.

jbeaurivage:
> Where should the breaking changes be documented? Should that be in the changelog on how to migrate to the next version.

The changelog, and maybe a short note in the readme as well?

> I've added a new feature to the hal, which is only needed if you want to use the take for Peripherals. Where should this be documented? I really would like to add an example for when you need it and what else you need so you don't get compiler errors.

What exactly does that feature entail? Is is something that's mandated by the new svd2rust version?

> When everything is ready, should I rebase and squash the commits and add a longer explanation into the commit message with the required things to do to migrate.

Yep, that sounds good!

pac/atsamd11c/src/lib.rs
```diff
@@ -914,18 +913,22 @@ pub struct Peripherals {
     pub WDT: WDT,
 }
 impl Peripherals {
-    #[doc = r"Returns all the peripherals *once*"]
+    #[doc = r" Returns all the peripherals *once*."]
+    #[cfg(feature = "critical-section")]
```
Tremoneck, line 917: The optional dependency is forced by this new addition to the generated pac. One can still use steal if one wants to avoid the dependency's, but has to use unsafe code.

jbeaurivage: Ah fair enough. Though I don't really see a case where someone wouldn't want to use the critical section feature - I would suggest enabling it in the HAL for all PACs. That way users and libraries can use the`critical-section` lib without having to worry about cargo features. After all, it's not like there should be another critical section provider in the dependency tree anyways - what do you think?

jbeaurivage:
@Tremoneck, svd2rust 0.29.0 is out. I don't know if you want to adapt to the new version, or stick to 0.28.0?

Tremoneck:
I've  been thinking  longg about the critical-section part. I believe you are  right with  providing always.  However, we  will  need documentation  in that  case that  you  will  require  another crate, that  provides the bindings,  as otherwise you will get  strange linker errors.

Might as well switch the script to 0.29, if I'm already updating it.

jbeaurivage:
What if the

> I've been thinking longg about the critical-section part. I believe you are right with providing always. However, we will need documentation in that case that you will require another crate, that provides the bindings, as otherwise you will get strange linker errors.

What if the PACs depended directly on `cortex-m` with the `critical-section-single-core` feature? That would directly take care of providing the critical-section bindings if I'm not mistaken.

jbeaurivage:
@Tremoneck, have you had the time to think about my last comment? I'm going to have a bit more bandwidth in the following weeks, I could assist in getting this ready for merge.

Tremoneck:
The critical section crate strongly encourages gating the implementation of critical sections behind a feature flag, so that it can be disabled, as linking two creates implementing it leads to two definitions and then linker errors. It would also make this crate incompatible with any crate that wants to implement it's own definition like a RTOS or an async runtime. 

For reference the STM32 crates https://github.com/stm32-rs/stm32-rs-nightlies/blob/3767c5e39a9c15e321c0230e53742f14b882be06/stm32f4/src/stm32f405/mod.rs#L4375 implement it as an optional feature as well. It is also designed in a way, where a user doesn't have to depent on an Implementor of a critical-section, as long as a critical section provider is linked in, it works.

jbeaurivage:
@Tremoneck that's fair. I suggest we follow stm32-rs' lead and expose a `critical-section` feature flag, and make it a default feature. That way users that are having linker collisions can easily disable it. Also, I don't know how you feel about upgrading to `svd2rust` 0.30? Hopefully it doesn't introduct other breaking changes.

jbeaurivage:
@Tremoneck, take a look at [#1](https://github.com/Tremoneck/atsamd/pull/1). If you agree and merge it, we could land this promptly.

jbeaurivage:
I thought some more about the `critical-section` issue. I think we're getting two things mixed up here:

* First, the `critical-section` crate recommends the feature-gating the critical section **implementation**. That is, if the PACs were providing their own idea of how a critical section should be implemented. In our case, an implementation is provided for us via the `critical-section-single-core` of the `cortex-m` crate. The PACs don't provide their own implementation, so we don't have to worry about it. The end user just has to include `cortex-m` as a dependency and all crates that depend on `critical-section` will now have access to that implementation. What would cause problems is if another crate also provided their own implementation, but no conflicts will ever happen with our PACs and HAL.
* `svd2rust` also feature-gates their use of `Peripherals::take()` only because some platforms don't have a `critical-section` implementation defined already (see [this](https://github.com/rust-embedded/svd2rust/pull/651#issue-1348672411) and [this](https://github.com/rust-embedded/svd2rust/pull/651#issuecomment-1225076308)). Feature-gating its use allows to mitigate that for those platforms specifically.

In our case, since `cortex-m` defines a critical-section impl, and our own crates don't, it's absolutely safe to use it unconditionally in the PACs' dependencies. [My PR](https://github.com/Tremoneck/atsamd/pull/1) now reflects that.

Tremoneck:
Yeah, thats a good point. Maybe I miss understood what the critical-section crate meant. Anyway, should there ever arise a need to remove this direct dependency on the cortex-m-rt crate it should still be possible to break it out as a default feature flag without breaking SemVer I think. I've merged your commits into my feature branch.

jbeaurivage:
Yeah sorry about that, I messed up the branch on top of which my PR was based. Thanks for catching it.
If you merge [#2](https://github.com/Tremoneck/atsamd/pull/3) to fix the builds I will land this. Thank you for following through!

ianrrees:
I've noticed some new warnings for ```warning: unknown lint: `private_bounds` ``` and `private_interfaces`, which seem to be caused by this update in combination with stable Rust.  As a meta-question, are you guys generally using unstable Rust?

jbeaurivage:
@ianrrees I usually use nightly compilers when working on async stuff (especially my HAL's [async branch](https://github.com/jbeaurivage/atsamd/tree/async). However even that is going to change with AFIT around the corner. Otherwise I try to stick with stable.

It's true that svd2rust seems to regularly try using some weird lints, perhaps too aggressively. I'm not too sure how to fix that on our end though.

ianrrees:
I see, thanks!  Will aim to follow up with svd2rust and see if we can arrive at a solution that works with both stable and unstable.

ianrrees:
Aha, svd2rust is already on it: https://github.com/rust-embedded/svd2rust/issues/760

In the meantime, I'll PR an addition of `#[allow(unknown_lints)]` to the affected PACs.

ianrrees:
> In the meantime, I'll PR an addition of #[allow(unknown_lints)] to the affected PACs.

Disregard - Rust 1.74.0 has these new lints

PR #685, opened 2023-05-14
Fix itsybitsy_m4 CI error

jbeaurivage:
# Summary
Fix `itsybitsy_m4` CI error on nightly.


jbeaurivage:
Self-approving so that CI can start being green again

PR #686, opened 2023-05-19
Add missing ADC pins for ATSAMD11D

Tremoneck:
# Summary
The ATSAMD11D has 20 or 24 pins. Some of them are also ADC pins, however the traits were not implemented for them.

# Checklist

### DIFF
--- a/hal/CHANGELOG.md
+++ b/hal/CHANGELOG.md
@@ -24,6 +24,7 @@
       - `Nvm::write_from_slice` -> `Nvm::write_flash_from_slice`
       - `Nvm::erase` -> `Nvm::erase_flash`
   - Refactor `Nvm::command_sync` to be less error-prone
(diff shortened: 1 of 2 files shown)


jbeaurivage:
@Tremoneck Just took a quick look at the datasheet. Unless I'm missing something, the pin multiplexing table calls out for PA02, PA03, PA04, PA05, PA06, PA07, PA10, PA11, PA14 and PA15 as ADC pins?

Tremoneck:
@jbeaurivage Sorry for the delay. 
While they are ADC pins, they are also available with the C variant. the Traits for these pins is implemented when the feature _samd11_ is available. That part is right above the code added in this PR.

jbeaurivage:
Right, I should've looked closed. My mistake! Looks good then.

PR #688, opened 2023-06-01
Fix incorrect APB clock state

glaeqen:
- TC4 APB clock should be disabled by default


glaeqen:
@bradleyharden and everyone that is interested: I2S related change is not yet there because I'm not 100% sure how to do it, shall we introduce `AhbTokens` struct for just a single token? Seems like an overkill, OTOH, it'll be nicely symmetrical to other abstractions.

glaeqen:
Ok, my bad. Turns out this is just typo in one part of the datasheet to falsly says that I2S is part of the AHB bus. Code was correct, so only TC4 seems wrong. This PR is IMHO ready for merging, unless @bradleyharden or whoever believes that datasheet lies with regard to TC4 as well.

PR #693, opened 2023-07-15
Fix release HAL crate workflow

jbeaurivage:
# Summary
Installing `lazy_static` no longer works as a workaround to force a refresh of the Cargo registry. Replace with `cargo update` instead.


sajattack (approved):

jbeaurivage:
I sneaked in a small fix to the `pygamer` nightly builds.

PR #709, opened 2023-12-10
New interrupt bindings

jbeaurivage:
Switch to an embassy-style macro for interrupt bindings instead of the old cortex-m-interrupt method.


jbeaurivage:
Whoops, meant to create this PR on my own fork.

PR #712, opened 2024-01-04
Readme and warnings

ianrrees:
# Summary
As discussed in chat, this PR moves some of the documentation in to the wiki, and makes a few small improvements.  Thinking is that the wiki is a better place for content that is more tied to external tools like openocd, probe-rs, etc.

# Checklist


jbeaurivage:
As discussed on Matrix, this looks great, thank you!

One final comment (and I think we should reserve that for a follow-up PR) - It would probably be good to add `dma` to the docs feature set.

PR #713, opened 2024-01-05
Add adafruit matrix portal m4

salsasteve:
# Summary
Created pin mapping for adafruit matrix portal m4 board. Created a blinky basic example. 

# Checklist

## If Adding a new Board

### DIFF
--- a/boards/matrix_portal_m4/Cargo.toml
+++ b/boards/matrix_portal_m4/Cargo.toml
@@ -19,7 +19,6 @@ version = "0.7"
 optional = true
 
 [dependencies.atsamd-hal]
-path = "../../hal"
 version = "0.16.0"
 default-features = false
 
@@ -29,7 +28,6 @@ optional = true
 
 [dependencies.cortex-m]
 version = "0.7"
-features = ["critical-section-single-core"]
 
 [dev-dependencies]
 panic-halt = "0.2"
--- a/boards/matrix_portal_m4/Cargo.toml
+++ b/boards/matrix_portal_m4/Cargo.toml
@@ -28,6 +28,7 @@ optional = true
 
 [dependencies.cortex-m]
 version = "0.7"
+features = ["critical-section-single-core"]
 
 [dev-dependencies]
 panic-halt = "0.2"
--- a/boards/matrix_portal_m4/Cargo.toml
+++ b/boards/matrix_portal_m4/Cargo.toml
@@ -1,0 +1,46 @@
+[package]
+name = "matrix_portal_m4"
+version = "0.1.0"
+authors = ["Salsa Steve <elmanstevelaguna@gmail.com>"]
+description = "Board Support crate for the Matrix Portal M4"
+keywords = ["no-std", "arm", "cortex-m", "embedded-hal"]
+categories = ["embedded", "hardware-support", "no-std"]
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/atsamd-rs/atsamd"
+readme = "README.md"
+edition = "2021"
+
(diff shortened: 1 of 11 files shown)


boards/matrix_portal_m4/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+# for cargo flash
+[package.metadata]
+chip = "ATSAMD51J19A"
+
+[dependencies.cortex-m-rt]
+version = "0.7"
+optional = true
+
+[dependencies.atsamd-hal]
+path = "../../hal"
```
jbeaurivage, line 22: Specify a `crates.io` version for`atsamd-hal` rather than a path dependency

boards/matrix_portal_m4/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+version = "0.16.0"
+default-features = false
+
+[dependencies.usb-device]
+version = "0.3.1"
+optional = true
+
+[dependencies.cortex-m]
+version = "0.7"
+features = ["critical-section-single-core"]
```
jbeaurivage, line 32: I think the `critical-section-single-core` feature should be left to the user to enable. Maybe they want a custom critical section implementation, which would conflict with this.

boards/matrix_portal_m4/README.md
```diff
@@ -0,0 +1,26 @@
+# Adafruit Matrix Portal M4 Board Support Crate
+
+This crate provides a type-safe API for working with the [Adafruit Matrix Portal M4
+board](https://www.adafruit.com/product/4745).
+
+## Prerequisites
+* Install the cross compile toolchain `rustup target add thumbv6m-none-eabi`
```
jbeaurivage, line 7: Change the toolchain to the correct one: `thumbv7em-none-eabihf`

crates.json
```diff
@@ -65,6 +65,11 @@
       "build": "cargo build --examples --all-features",
       "target": "thumbv7em-none-eabihf"
     },
+    "matrix_portal_m4": {
+      "tier": 1,
```
jbeaurivage, line 69: This board will be a Tier 2 rather than 1.

jbeaurivage (changes requested):
Thank you for this. I added a few comments that need to be addressed, but good work in general. Specifically, this board will be a Tier 2, so its dependency on `atsamd-hal` should not be a path dependency but rather use the crates.io version.

salsasteve:
Thank you for the review. I have fixed the issues above and retested the board again

boards/matrix_portal_m4/Cargo.toml
salsasteve: Removed `path = "../../hal"`

boards/matrix_portal_m4/Cargo.toml
salsasteve: Removed `features = ["critical-section-single-core"]`

boards/matrix_portal_m4/README.md
salsasteve: Changed thumbv6m-none-eabi to thumbv7em-none-eabihf

crates.json
salsasteve: Changed `"tier": 1` to `"tier": 2`

jbeaurivage:
@salsasteve, I'm sorry, seems like I was mistaken regarding the `critical-section-single-core` feature. Looks like all other BSPs enable it. Can you please reenable it?  Everything else looks good, so I will merge as soon as it's done.

jbeaurivage (approved):

PR #714, opened 2024-01-08
Fix all clippy lints

### LINKED ISSUE #574: Consider running clippy more aggressively

jbeaurivage:
# Summary
Fix all HAL clippy lints.

Also up for debate: I've switched the CI build step to clippy, and having it deny all clippy warnings. However some might think that it is too aggressive.

Closes #574.

# Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,7 +4,7 @@
 # Checklist
   - [ ] `CHANGELOG.md` for the BSP or HAL updated
   - [ ] All new or modified code is well documented, especially public items
-  - [ ] No new warnings or clippy suggestions have been introduced (see CI or check locally)
+  - [ ] No new warnings or clippy suggestions have been introduced - CI will **deny** clippy warnings by default! You may `#[allow]` certain lints where reasonable, but ideally justify those with a short comment. 
 
 ## If Adding a new Board
   - [ ] Board CI added to `crates.json`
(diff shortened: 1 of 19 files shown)


ianrrees:
I like the idea of requiring a clean Clippy pass in CI - it seems like overall there's less headache in preventing the warnings at the outset, vs everyone ignoring them in later development

jbeaurivage:
Sounds good. That doesn't mean we should blindly follow clippy's advice all the time; `#[allow]`ing certain lints is totally acceptable, but I think we should try to explain why we're allowing it with a comment where reasonable.

ianrrees:
Showing my ignorance here: does clippy also build the code?  Just wondering if there might be some kinds of issue that would pass `cargo clippy` but not `cargo build` - if so we might want CI to do both

jbeaurivage:
@ianrrees you're right, it doesn't build the code - sometimes things like linker errors will pop up when building but not checking. However, since we *do* build the BSP examples, I'm fairly confident these issues would be caught by CI regardless.

ianrrees:
> since we _do_ build the BSP examples, I'm fairly confident these issues would be caught by CI regardless.

Yep, that makes a lot of sense to me!  Any issues that "escaped" would be good to cover with additional examples - sortof a build time regression test suite.

Issue #332, opened 2020-11-24
Update embedded-hal dependency to 1.0


jessebraham:
Once [embedded-hal] has pushed a stable `1.0` release (the latest at time of writing is `1.0.0-alpha.4`), we should update our HAL to use it.

This will involve updating the dependency, then going through each peripheral and updating the imports/trait implementations/etc. until the project builds cleanly.

The stable release of [embedded-hal] can be tracked via [this issue] and [this milestone].

[embedded-hal]: https://github.com/rust-embedded/embedded-hal
[this issue]: https://github.com/rust-embedded/embedded-hal/issues/177
[this milestone]: https://github.com/rust-embedded/embedded-hal/milestone/1

Tremoneck:
One of the last remaining blockers for the embedded-hal 1.0 release are missing implementations of the API to show, that the traits work fine for the drivers and hals.

I would like to help with that here. Creating this implementation directly on master is a bad idea, as that would hinder any other improvements from getting released. Creating everything in my own branch would make a massive pull request, which would be really hard to review. Also there wouldn't be any proper feedback from the experienced devs here. My suggestion is therefor to create a new branch in this repo called something along the lines of _ehal-1.0_ where anyone interested in could help to implement these changes already and pulling self contained updates into this branch.

Advantages:
The current proposed traits can be checked and help release ehal 1.0
When the release comes the ehal branch can be merged back into master with most/all of the implementation ready.

Disadvantages:
Some or all of the time spent could be for nothing, if the traits change substantially.

sajattack:
Thanks for the offer of help :smile: I think it would be more beneficial to regenerate the PACs (#648), resolve the HAL breakage that causes, and *then* update ehal. Maintainer time is also very limited these days, so unfortunately these other issues haven't been addressed yet.

tgross35:
Since there has been a RC (rather than alphas), it seems like a 1.0 release is close and looking for last minute feedback. See https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-1805300873

tgross35:
The `esp-hal` crate supports 1.0 under a feature flag, maybe that would be workable here? https://github.com/esp-rs/esp-hal/blob/07ed22df17f28f1b1890aa47c3ae599182397712/esp-hal-common/Cargo.toml#L23

rursprung:
[...]

PR #715, opened 2024-01-13
Add initial support for `embedded-hal` version 1.0

tgross35:
# Summary

This adds a dependency and feature for embedded-hal v1.0 and a basic implementation for the `DelayNs` trait. I think it is best to start with this small subset to make sure CI and everything are set up properly, then add other traits one at a time.

Part of  #332

# Checklist

## If Adding a new Board

### DIFF
--- a/hal/src/sleeping_delay.rs
+++ b/hal/src/sleeping_delay.rs
@@ -4,9 +4,13 @@ use cortex_m::asm;
 use fugit::ExtU32;
 
 use crate::ehal::blocking::delay::{DelayMs, DelayUs};
+#[cfg(feature = "embedded-hal-1")]
+use crate::ehal1::delay::DelayNs;
 use crate::timer_traits::InterruptDrivenTimer;
 
 const NUM_US_IN_S: u32 = 1_000_000;
+#[cfg(feature = "embedded-hal-1")]
+const NUM_NS_IN_S: u32 = 1_000_000_000;
 
 /// Delay and sleep while we do (WFI) using a timer
 pub struct SleepingDelay<TIM> {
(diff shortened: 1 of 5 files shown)


tgross35:
What is the best way to make sure this gets tested in CI? I am not sure if I should update the matrix for this.

hal/src/sleeping_delay.rs
```diff
@@ -35,6 +39,34 @@ where
...
+#[cfg(feature = "embedded-hal-1")]
+impl<TIM> DelayNs for SleepingDelay<TIM>
+where
+    TIM: InterruptDrivenTimer,
+{
+    fn delay_ns(&mut self, ns: u32) {
+        // Determine how many cycles we need to run for this delay, if any
+        // Avoid timers that run longer than a second because for 48 MHz-based timers,
+        //   there is no valid divisor + cycle count greater than ~1.3s, so we'd panic.
+        let mut loop_count: u32 = 1 + (ns / NUM_NS_IN_S);
```
tgross35, line 51: I think our timers take a u32 nanosecond count, so this loop may be able to be removed. Need to verify

hal/src/sleeping_delay.rs
```diff
@@ -49,6 +81,7 @@ where
         let mut count: u32 = 1 + (us / NUM_US_IN_S);
 
         // Start the timer and sleep!
+        // TODO: why does this use nanos?
         self.timer.start((us / count).nanos());
```
tgross35, line 85: I must not understand fugit correctly, but it looks to me like this is treating microseconds as nanos with no conversion. What is going on here?

jbeaurivage:
Thanks @tgross35! I'll create an embedded-hal-1 branch in the `atsamd-hal` repo, so that we can split the effort in multiple PR, and merge that once we're ready. Can you please change this PR to use that as its base branch?

I'll have to review your code  bit more in detail concerning the timer stuff, but I don't think we should hide `embedded-hal` v1.0 behind a feature. Implementations for the 0.2 and 1.0 versions can easily coexist for one release, after which I plan on removing 0.2 if everyone else agrees. It think it's also a good opportunity to remove the `unproven` feature, which imp doesn't really serve a purpose at this point. 

As far as naming, I think I would reexport `embedded-hal` v1.0 as `ehal`, and v0.2 as `ehal_0_2`. To me it makes it a bit clearer that 1.0 is the "default". What do you think?

tgross35:
Hey, thanks for looking @jbeaurivage! The reason I started doing it in this additive way, rather than replacing, is that v1.0 has a list of removed traits that https://github.com/rust-embedded/embedded-hal/blob/master/docs/migrating-from-0.2-to-1.0.md#removed-traits. In particular, `CountDown`, `Periodic`, and `Pwm` all provide pretty important functionality.

But maybe it is okay to switch to eh1.0 as the primary implementation while still keeping eh0.2 around to fill in the missing traits. In this case, I think it would be better to not gate either of them behind a feature because some of the eh1.0 relies on eh0.2 (e.g. in this PR `eh1::DelayNs` makes use of `CountDown`). What are your thoughts?

jbeaurivage:
Yeah, what I meant is for the transition release, we should provide implementations for *both* versions of ehal in parallel.

Furthermore, I would suggest we take the actual logic out of the ehal v0.2 implementations, and instead put them in inherent methods on the peripherals (or on the ehal v1.0 traits of you prefer) . Implementing v0.2 traits then becomes a matter of forwarding to the correct method call. The idea being that 1.0 implementations should be totally independent from 0.2 impls (the reverse isn't necessary, however). Then, when we'll want to remove 0.2, we can simply delete everything related to 0.2 without affecting anything else.

sajattack:
superceded by #723

PR #718, opened 2024-01-18
Update usb-device

ianrrees:
# Summary
The usb-device crate is used by both the HAL and USB Class crates, to create a USB Device.  The same version of usb-device needs to be used for all the dependencies of a firmware, which can lead to difficulty for instance if a required USB Class crate hasn't been updated.

This PR updates usb-device to the latest released version.  This is a breaking change for both usb-device and our HAL.

This commit updates our tier 1 BSPs only.

(when we cut the usb-device 0.3.0 release, we forgot to include a new type in the prelude, which made it somewhat annoying to use - hence the 0.2 -> 0.3.1)

# Checklist

### DIFF
--- a/boards/atsame54_xpro/Cargo.toml
+++ b/boards/atsame54_xpro/Cargo.toml
@@ -23,11 +23,11 @@ features = ["critical-section-single-core"]
 
 [dependencies.atsamd-hal]
 path = "../../hal"
-version = "0.16.0"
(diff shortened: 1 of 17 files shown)


sajattack (approved):

sajattack:
Looks like rustfmt is mad about the isochronous whitespace but otherwise LGTM

Issue #740, opened 2024-06-21
I2C Transaction Implementation Possibly Not According to Specification?


fko-kuptec:
First of all, sorry for the many edits. I unfortunately hit [Ctrl+Enter] for a newline, but instead created an unfinished issue. Now it is done. 😅

The specification of the new [`I2c::transaction`](https://docs.rs/embedded-hal/latest/embedded_hal/i2c/trait.I2c.html#tymethod.transaction) method does not directly mention this, but to me it would make sense to not send a `SAD+R/W` between multiple consecutive `Write` operations. The [current implementation](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/sercom/i2c/impl_ehal.rs#L29), however, does that. Is this just an oversight or on purpose?

## Specification

    Execute the provided operations on the I2C bus.
    
    Transaction contract:
    - Before executing the first operation an ST is sent automatically. This is followed by SAD+R/W as appropriate.
    - Data from adjacent operations of the same type are sent after each other without an SP or SR.
    - Between adjacent operations of a different type an SR and SAD+R/W is sent.
    - After executing the last operation an SP is sent automatically.
    - If the last operation is a `Read` the master does not send an acknowledge for the last byte.
    
    - `ST` = start condition
    - `SAD+R/W` = slave address followed by bit 1 to indicate reading or 0 to indicate writing
    - `SR` = repeated start condition
    - `SP` = stop condition

## Example

```rust
fn write_register(slave_address: u8, register_address: u8, data: &[u8]) -> Result<(), Error> {
    let mut ops = [
        Operation::Write(&[register_address]),
        Operation::Write(data),
    ];
    
    i2c.transaction(slave_address, &mut ops);
}
```

As far as I could find out, this API call would produce with the current implementation the following sequence on the bus:

```
ST SAD+W [register_address] SP ST SAD+W [data] SP
```

I, however, would expect the following:

```
ST SAD+W [register_address] [data] SP
```

If this is indeed not intended behaviour, I could probably fix it and create a pull request. 😊

jbeaurivage:
Closed via #741

PR #741, opened 2024-06-21
Make I2C transactions continuous according to specification (#740)

fko-kuptec:
# Summary
Changes the implementation of the [`ehal::i2c::I2c::transaction`](https://docs.rs/embedded-hal/latest/embedded_hal/i2c/trait.I2c.html#tymethod.transaction) method to match the specification. The corresponding issue is #740.

I would be thankful if someone could have a look at it, to make sure that it now actually matches the specified behaviour. My code seems works now, but I might have misunderstood something.

# Checklist

### DIFF
--- a/hal/CHANGELOG.md
+++ b/hal/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased Changes
 
+- Fix I2C transaction to be as continuous as possible according to `embedded-hal` specification
 - Allow configuring USB clock with `GenericClockController` on atsamd11
 - fix samd51j not having i2s support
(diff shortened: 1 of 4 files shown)


jbeaurivage:
Hi @fko-kuptec, I agree with what you said in #740. Let me check if this PR achieves that, I can potentially test with a logic analyzer some time next week.

fko-kuptec:
> Hi @fko-kuptec, I agree with what you said in #740. Let me check if this PR actually does achieve that, I can potentially test with a logic analyzer some time next week.

Thanke you 😊

PR #743, opened 2024-08-16
Wait at end of ehal 0.2 blocking SPI writes

ianrrees:
# Summary
Discovered while tracking down an issue with the PyGamer display identified [here](https://github.com/atsamd-rs/atsamd/pull/455) - basically for non-duplex SPIs, the embedded-hal 0.2 blocking writes were returning as soon as the last word was written to the SERCOM, not waiting until it had been shifted out.  This resulted in the command/data line for the LCD controller being toggled out of sync with the SPI data stream.

I've tested this only with the PyGamer example.

# Checklist

### DIFF
--- a/hal/CHANGELOG.md
+++ b/hal/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased Changes
 
+- Fix blocking behaviour for `embedded-hal` 0.2 SPI writes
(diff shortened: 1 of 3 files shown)


sajattack:
clippy's a little bit unhappy with this, but once you resolve that it should be good to go.

sajattack (approved):

Issue #744, opened 2024-08-18
CI falures for tier >1 BSPs shouldn't interfere with HAL work


ianrrees:
Our current CI setup builds all the BSP examples, and a failure in one to build with stable Rust results in subsequent builds not being run.  This creates a problem when a BSP failure is caused by some change outside of our control, for instance a dependency breaking SemVer conventions in an update.

This is particularly egregious when a tier >1 BSP build fails, for a CI run meant to test changes to the HAL:

![image](https://github.com/user-attachments/assets/8f8f2fd5-c211-4a1f-a8d8-d4efa5ee3d2f)

One possible solution is to do the Tier 1 BSP builds before starting on the lower-tier BSPs.

ianrrees:
I've made a naive attempt at fixing this by changing the list-BSPs action to provide the BSPs ordered by tier - but the build-BSPs workflow seems to do the builds in random order anyway.

Two possible next approaches spring to mind:

1. See if the jobs matrix can work with JSON objects, or just strings.  If so, instead of supplying bsps as `{"bsp":["arduino_mkr1000","arduino_mkrvidor4000"...`, use something that includes the tier like `{"bsp":[{"name": "arduino_mkr1000", "tier": 2},{"name": "arduino_mkrvidor4000","tier": 2}...`.  I believe this would allow the build-bsp workflow to allow `continue-on-error` to reference the tier.
2. Split the build-bsp workflow in to two, for tier 1 and >1 BSPs.

ianrrees:
There are a few subtly different ways to have the BSP build jobs continue after a failure, which to use basically comes down to whether we want the workflow to show a  ❌ (set [`fail-fast`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) to false) or ✔️ (use a tier clause in `continue-on-error`) depending on whether _any_ BSP builds fail.  We currently use `continue-on-error` for builds with Rust nightly, so get a ✔️ despite all the nightly builds failing clippy.

Some experimentation and discussion in Matrix shows that the success or failure of the CI workflow doesn't directly control whether a PR can be merged (without ticking the box with the big warning ;) ).  Currently, the required checks are set through the project's branch protection settings - for the BSPs, this is a manually curated list of the Tier 1 builds with stable Rust.

[...]

PR #745, opened 2024-08-18
CI: continue BSP building workflow if build any BSP fails

ianrrees:
# Summary
Potential fix for https://github.com/atsamd-rs/atsamd/issues/744 - this change to CI has the "Build BSPs" workflow continue regardless of tier >1 BSP builds failing.

One downside to this approach is that the overall workflow run may show success despite there actually being BSP build failures.  This PR is a test case to see how obvious this situation is in practice - a tier 2 BSP is expected to fail.

### DIFF
--- a/.github/actions/list-BSPs/action.yml
+++ b/.github/actions/list-BSPs/action.yml
@@ -12,5 +12,5 @@ runs:
     - id: compute-matrix
       shell: bash
       run: |
-        matrix_json=$(cat crates.json | jq -Mr -c '{ "bsp": (.boards | keys ), "toolchain": ["stable", "nightly"] }')
+        matrix_json=$(cat crates.json | jq -Mr -c '{ "bsp": [ (.boards | to_entries | .[] | {"name": (.key), "tier": .value.tier}) ] , "toolchain": ["stable", "nightly"] }')
         echo "matrix=${matrix_json}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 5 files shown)


ianrrees:
The required statuses above come from the project's branch protection settings, so this change is really just about whether the CI jobs are run or not.

jbeaurivage:
I wouldn't say that a Tier 2 is necessarily expected to fail. As discussed on Matrix, Tier 2 BSPs are locked to a specific `crates.io` HAL version, so they normally should succeed when nothing has been changed (unless a dependency breaks semver, like is the case for the `itsybitsy_m0`. 

I don't know how much complexity this would add, but just a suggestion to get around this:

Would it be feasible to only run CI on BSPs if:
* It is a Tier 1 BSP
* If it's a Tier 2 BSP, run CI only if the BSP had some files changed in the PR

ianrrees:
> I wouldn't say that a Tier 2 is necessarily expected to fail

Sorry, that was poor wording on my part, meant that in this specific PR the itsybitsy m0 is expected to fail.

> Would it be feasible to only run CI on BSPs if...

As external issues arise, with this change we'll know about them but they won't interfere with work.  To me, this seems a bit better than letting issues silently accumulate, or relying on users to report them.

That said, I think that would be possible, and likely worth doing if we start running in to issues with time/cost of building all the BSPs.

PR #750, opened 2024-08-27
Update pygamer to HAL 0.17

ianrrees:
# Summary
Initial work by @bradleyharden for v2 SERCOM, with more updates to account for improvements in our project and others.

Still some issues to work through, but at this stage all the examples compile:
 - [X] Remove neopixel support (can be reverted once working drivers are available)
 - [X] Decide whether to remove or fix the USB polling example (Fixed)
 - [X] Clean up Git history
 - [X] Update `CHANGELOG.md` for the BSP


ianrrees:
From some not-very-thorough testing today, I think fixing the timer-based neopixel issues will at least require some work in ws2812-timer-delay-rs because the timing it generates (via GPIO and waiting on a supplied timer) isn't quite right on SAMD51 with the new HAL.  I've tried adjusting the library locally, and have the examples working well with opt-level="s", but then dialing it up to 1 or higher causes problems again.  On paper, and based on the WS2812 datasheet I found, the current delays in that crate shouldn't work, so I think it's likely that fixing them for this case would just move the problem to some other user of that crate.

Although the SPI neopixel example isn't working at present, that architecture seems a lot more reasonable to me; in particular the timer+GPIO approach would be a stretch to make work on a 48MHz SAMD21, but should be fine using a SERCOM on that part.  I'm thinking that the appropriate thing might be to move the PyGamer neopixel examples away from the timer-based crate and to the SPI-based one.

ianrrees:
Debugging the SPI-driven WS2812 examples revealed that there are some subtle changes in behaviour of SPI writes between the old and new HAL.  The old behaviour doesn't work reliably for me in release profile (the pixels show the rainbow, but flicker quite a bit), and the new behaviour doesn't work at all for WS2812 despite behaving correctly according to the `SpiBus` trait AFAICT.

So, at present, I think that the neopixel examples are broken for reasons outside of atsamd-rs.  I've opened #752 to track that, since it's not specific to PyGamer although some of the issues are worse with a string of multiple neopixels.  The question is whether we want to get those examples working before moving forward with this PR.

My vote is to remove the neopixel examples from PyGamer until we can get them working, and proceed with this PR.

Issue #752, opened 2024-08-28
Neopixel examples are problematic


ianrrees:
In #750 I've been working on updating PyGamer to the current HAL, and I'm finding that the WS2812 / neopixel is a tricky problem to solve generically.  In the timer-based WS2812 driver, there's a fundamental issue with the time it takes to toggle a GPIO varying between platforms, and also being comparable to the protocol's timing margins.  The SPI driver relies on behaviours of the provided `SpiBus` which aren't described by the trait (nor reliably provided by our current implementation).

So, it's not really clear to me how we should proceed here.  A few ideas:

1. With #751 point 2 in place, it _might_ be feasible to get the WS2812 SPI driver to work well enough for a demo with a few asterisks like only working on release profile (the status quo).  I think we'd need to modify the driver so that it shifts out 3 bytes of WS2812 data (12 SPI bytes, because each bit of WS2812 data is synthesized from 4 bits on SPI) per call to `SpiBus::write()` - this would minimise the variation in bit timing to any particular WS2812, and hopefully it doesn't matter as much for the data between WS2812s.
2. When the timer+GPIO driver is updated to use embedded-hal 1.0 `DelayNs` ([in progress](https://github.com/smart-leds-rs/ws2812-timer-delay-rs/issues/5)), we could possibly introduce a parameter for the time that a GPIO write takes, to allow tuning for the particular board.
3. At least on the PyGamer, another option is to use the TCC peripheral to generate the WS2812 bit timing.  Going this route might provide a good example of driving a peripheral directly, which could help for people who want to use peripherals/features which aren't provided by the HAL.

jbeaurivage:
#800 Fixes neopixels when using SPI without DMA.

sajattack:
Should we close this then or are there other cases we need to test first?

jbeaurivage:
I think at this point it's just a matter of reinstating the neopixel examples in the pygamer BSP - we had temporarily removed them to avoid shipping broken examples. I'm not sure if other BSPs would also benefit from using SPI rather than bitbanging the protocol.

jbeaurivage:
Before this issue gets closed, I 'd like to point out that wile we seem to be gravitating towards using SPI to drive our neopixel examples, there could be other ways to drive them using hardware acceleration.

[...]

PR #753, opened 2024-08-29
grand_central_m4: Bump usb-device version

supersimple33:
# Summary
I am bumping the version of usb-device for the grand central.

# Checklist

### DIFF
--- a/boards/grand_central_m4/CHANGELOG.md
+++ b/boards/grand_central_m4/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - update path of Cargo config
+- bumped usb-device version to 0.3
 
 # v0.6.0
 
--- a/boards/grand_central_m4/CHANGELOG.md
+++ b/boards/grand_central_m4/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 - update path of Cargo config
-- bumped usb-device version to 0.3
+- update to `atsamd-hal-0.17` and other dependencies (#753)
 
 # v0.6.0
 
--- a/boards/grand_central_m4/Cargo.toml
+++ b/boards/grand_central_m4/Cargo.toml
@@ -20,7 +20,7 @@ version = "0.17.0"
 default-features = false
 
 [dependencies.usb-device]
-version = "0.3.1"
+version = "0.3"
 optional = true
 
 [dependencies.cortex-m]
--- a/boards/grand_central_m4/Cargo.toml
+++ b/boards/grand_central_m4/Cargo.toml
@@ -23,6 +23,10 @@ default-features = false
 version = "0.3.1"
 optional = true
 
+[dependencies.cortex-m]
+version = "0.7"
+features = ["critical-section-single-core"]
+
 [dev-dependencies]
 cortex-m = "0.7"
 usbd-serial = "0.2"
--- a/boards/grand_central_m4/Cargo.toml
+++ b/boards/grand_central_m4/Cargo.toml
@@ -1,9 +1,7 @@
 [package]
 name = "grand_central_m4"
 version = "0.6.0"
-authors = [
-    "Dustin Little <dlittle@toyatech.net>"
-]
+authors = ["Dustin Little <dlittle@toyatech.net>"]
 description = "Board Support crate for the Adafruit Grand Central M4 Express"
 keywords = ["no-std", "arm", "cortex-m", "embedded-hal"]
 categories = ["embedded", "hardware-support", "no-std"]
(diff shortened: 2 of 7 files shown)


ianrrees:
Thanks @supersimple33 !  I think this might be a more involved change, because the Grand Central BSP is using HAL v 0.15 which I think uses usb-device 0.2 .

I'd suggest you clone atsamd-rs locally, cd in to `boards/grand_central_m4` and run `cargo build --examples --all-features`.  Once you're in a position where that works, change the HAL dependency to 0.17 and usb-device to 0.3.1 (there may be a couple other updates I'm forgetting).  You can then work through the errors building one example at a time via `cargo build --example=blinky_basic` for example - metro_m4 would be a good reference point, as it uses the current HAL.

Per #752 the neopixel example probably won't actually work - I guess the right thing to do is get it to compile but otherwise leave it in place.

I'm happy to help with the update if you've got hardware to test with.

ianrrees:
Also, don't mind the CI failures on nightly - looks like we need to update docstrings - or itsybitsy.

supersimple33:
I'm getting the following error now @ianrrees  
```
  = note: rust-lld: error: undefined symbol: _critical_section_1_0_acquire
          >>> referenced by lib.rs:193 (/Users/addisonhanrattie/.cargo/registry/src/index.crates.io-6f17d22bba15001f/critical-section-1.1.3/src/lib.rs:193)
          >>>               /Users/addisonhanrattie/Documents/atsamd/boards/grand_central_m4/target/thumbv7em-none-eabihf/debug/examples/blinky_basic-66cbd02f4a7e08ab.blinky_basic.dd7903c5ecf9b979-cgu.0.rcgu.o:(critical_section::with::h392fdb763c500007)
          
          rust-lld: error: undefined symbol: _critical_section_1_0_release
          >>> referenced by lib.rs:210 (/Users/addisonhanrattie/.cargo/registry/src/index.crates.io-6f17d22bba15001f/critical-section-1.1.3/src/lib.rs:210)
          >>>               /Users/addisonhanrattie/Documents/atsamd/boards/grand_central_m4/target/thumbv7em-none-eabihf/debug/examples/blinky_basic-66cbd02f4a7e08ab.blinky_basic.dd7903c5ecf9b979-cgu.0.rcgu.o:(core::ptr::drop_in_place$LT$critical_section..with..Guard$GT$::hebcbe303f9685640)
```
when compiling.


I ran cargo tree on the `grand_central_m4` and the `metro_m4` directories and I noticed that in `metro_m4` the cortex-m crate has an extra dependency of critical_section not in the grand_central_m4 equivalent which fails to compile. Any advice on how to fix this?

jbeaurivage:
@supersimple33: you have a missing critical section implementation. Just add thr `critical-section-single-core` feature to the `cortex-m` dependency and you should be good to go!

ianrrees:
I recently started a wiki page with notes on updating firmwares to newer HALs - have added the critical section thing there.

https://github.com/atsamd-rs/atsamd/wiki/HAL-Update-Guide

boards/grand_central_m4/Cargo.toml
```diff
@@ -12,32 +10,34 @@ edition = "2021"
...
-ws2812-timer-delay = "0.3"
-
 [dependencies.cortex-m-rt]
 version = "0.7"
 optional = true
 
 [dependencies.atsamd-hal]
-version = "0.15"
+path = "../../hal"
+version = "0.17.0"
```
jbeaurivage, line 19: The `grand_central_m4` is a Tier 2 BSP; it should depend on the `crates.io` version of the HAL, and not specify a path dependency.

pac/atsamd51p/Cargo.toml
```diff
@@ -11,7 +11,7 @@ readme = "README.md"
 edition = "2021"
 
 [dependencies]
-cortex-m = "0.7"
+cortex-m = { version = "0.7", features = ["critical-section-single-core"] }
```
jbeaurivage, line 14: The PAC should NOT enable the `critical-section-single-core` feature, only the BSP. Otherwise, users of the PAC would not have the option to bring in their own critical section implementation as needed.

boards/grand_central_m4/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - update path of Cargo config
+- bumped usb-device version to 0.3
```
jbeaurivage: I would also definitely add that the HAL dependency has been upgraded to 0.17

jbeaurivage (changes requested):
Good stuff overall! There are a few minor but important things that should be addresses before we can merge this.

boards/grand_central_m4/CHANGELOG.md
jbeaurivage: Whoops, looks like you committed as I was writing my comment.

supersimple33: yeah realized that when I was looking things over, oops

boards/grand_central_m4/Cargo.toml
supersimple33: ahh ok

pac/atsamd51p/Cargo.toml
supersimple33: Ok yeah that makes sense

jbeaurivage (approved):

supersimple33:
@jbeaurivage @ianrrees thanks for the guidance on this

PR #754, opened 2024-08-30
Update itsybitsy_m0 and itsybitsy_m4 to HAL 0.17

ianrrees:
# Summary
Quite a few small updates

I'm a bit unsettled about the HAL prelude.  Using it winds up bringing in multiple traits which apparently require ugly disambiguation, it seems like this should be cleaner:

```Rust
InterruptDrivenTimer::start(&mut timer, Hertz::MHz(4).into_duration());
```

# Checklist

### DIFF
--- a/boards/itsybitsy_m0/examples/ssd1306_terminalmode_128x64_i2c.rs
+++ b/boards/itsybitsy_m0/examples/ssd1306_terminalmode_128x64_i2c.rs
@@ -77,7 +77,6 @@ use bsp::entry;
 use hal::clock::GenericClockController;
 use hal::delay::Delay;
 use hal::prelude::*;
-use hal::time::KiloHertz;
 use pac::{CorePeripherals, Peripherals};
 
 #[entry]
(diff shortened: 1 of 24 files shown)


jbeaurivage:
That should probably be a separate discussion, but I'm not a fan of the prelude in the first place. I'm happy to merge though if you think this is ready!

boards/itsybitsy_m0/examples/ssd1306_terminalmode_128x64_i2c.rs
```diff
@@ -97,7 +96,8 @@ fn main() -> ! {
     let i2c_sercom: bsp::I2cSercom = peripherals.SERCOM3;
     let i2c = bsp::i2c_master(
         &mut clocks,
-        KiloHertz(400),
+        400.kHz(),
+        // KiloHertz(400),
```
jbeaurivage, line 100: Maybe just get rid of the commented out line?

ianrrees: Good catch, fixed!

PR #756, opened 2024-08-31
Update PACS to svd2rust 0.34

jbeaurivage:
# Summary
The latest sv2rust versions introduced a lot of breaking changes, namely:

* Pretty much everything, including peripherals and registers, use CamelCase now.
* Peripheral accessors are now functions instead of struct fields.

I also had to fix some SVD errors in the samd21 USB peripheral, where the SVD disagreed with the datasheet. 

# Checklist

## If Adding a new Board


sajattack:
Thanks for getting this done.

One question though.
You mentioned 
> I also had to fix some SVD errors in the samd21 USB peripheral, where the SVD disagreed with the datasheet.
Can you create those changes as xslt patches so we don't have to redo them next time?

jbeaurivage:
> Can you create those changes as xslt patches so we don't have to redo them next time?

Absolutely, already done! That's what I meant by "fixing SVD errors", though it might be hard to find when there are 5000+ files changed 😅

> Actually, 2 questions. are there any changes from microchip to the SVDs we can integrate at the same time?

Let me go dig in the microchip docs to see if I can't find an updated version

sajattack:
> Absolutely, already done! That's what I meant by "fixing SVD errors", though it might be hard to find when there are 5000+ files changed 😅

Ah my bad github only loaded the first 3000 and I didn't see the svd folder changes :stuck_out_tongue: 
image

jbeaurivage:
All done @sajattack!

sajattack:
I did a spot check (diff) of the ATSAMD51J19A we have and the ones from here 
https://packs.download.microchip.com/#collapse-Microchip-SAMD51-DFP-pdsc

and there are some minor differences.

jbeaurivage:
You're right, there are some differences. I must've gotten mine from elsewhere. 

I brought in the new SVDs. However I had to manually go edit the PACs for SAME53 and SAME54 PACs, specifically: the `ENID` field in `src/gmac/tidm.rs` is duplicated, even though the SVD doesn't have duplicated. It's not great, but it's also not a something we can fix with an XSLT patch. It looks like a bug with `svd2rust`, let's hope it gets fixed before the next time around.

PR #764, opened 2024-10-01
fix: Changes to the `dmac` public API

jbeaurivage:
# Summary

I noticed that there were a few mistakes in the public API of the `dmac` module. Some things were missing, and some others were publicly exposed when they shouldn't.

* Remove `BlockTransferControl` and `DmacDescriptor` structs from the public API
* Add getters/setters for level0, level1, level2 and level3 for `PriorityLevelMask` and `RoundRobinMask` structs

### DIFF
--- a/hal/src/dmac/channel/mod.rs
+++ b/hal/src/dmac/channel/mod.rs
@@ -31,6 +31,8 @@
 //! `Uninitialized` state. You will be required to call [`Channel::init`]
 //! again before being able to use it with a `Transfer`.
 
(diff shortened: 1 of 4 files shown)


jbeaurivage:
Now that I think about it, I don't think any of the changes in this PR are breaking. The `BlockTransferControl` and `DmacDescriptor` structs, and the `const DEFAULT_DESCRIPTOR` were all `#[doc(hidden)]`, which explicitly excludes them from the publiic API. Making them private now isn't breaking.

PR #766, opened 2024-10-08
Promote the `pygamer` BSP to Tier 1 status

kyp44:
# Summary
Updates PAC references for the `pygamer` BSP, which was necessary due to [#576](https://github.com/atsamd-rs/atsamd/pull/756).

# Checklist


jbeaurivage:
Thanks @kyp44! I'll keep this PR on standby for now, I want to make a new `pygamer` release before merging this to main, since this will mark the BSP's entry into Tier 1 status.

boards/pygamer/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 # v0.10.0
 
+- update PAC references for item name name changes (#756)
```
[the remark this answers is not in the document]
kyp44: Done. Sorry, misinterpreted the CHANGLEOG format.

jbeaurivage:
[Follow-up of the conversation [here](https://github.com/kyp44/atsamd/pull/1)]:

> @jbeaurivage I did not update those dependencies, but it would definitely be nice to be able to use the latest versions of those.
> EDIT: I don't mind working on updating these (and perhaps any other outdated dependencies) if you like, either in this branch or in a new one. Just let me know.

I think that would be great. We should strive to provide the latest and greatest for T1 boards. I think the `embedded-hal` v1.0 SPI implementation is a bit buggy at the moment and has some fragile timing and race conditions. I'll work on trying to fix these. Hopefully you can help with testing, since I don't have a pygamer or an LCD display.

boards/pygamer/examples/pwm_tcc0.rs
```diff
@@ -40,9 +40,9 @@ fn main() -> ! {
     let mut pwm0 = Tcc0Pwm::new(
         &clocks.tcc0_tcc1(&gclk).unwrap(),
         1.kHz(),
-        peripherals.TCC0,
+        peripherals.tcc0,
         TCC0Pinout::Pa23(tcc0_pin),
-        &mut peripherals.MCLK,
+        &mut peripherals.mclK,
```
jbeaurivage, line 45: There's a typo here which fails the build:
```diff
- &mut peripherals.mclK,
+ &mut peripherals.mclk,
```

jbeaurivage:
Thanks for writeup. I would suggest to work on updating the dependencies here (Could rename the PR to something along the lines of "Promote pygamer to Tier 1 status").

About your questions:

* `micromath` and `embedded-sdmmc`: Can definitely be moved to dev-dependencies
* `cortex-m-rtic` seems to be used in the `button_rtic` example. Most likely it shouldn't be too much effort to update the example to use RTIC v2
* I think we can yank out embedded-hal v0.2

Some additional thoughts

* Upgrading st7735-lcd to 0.10 will have the nice addition of using the embedded-hal v1.0 implementations of the HAL.
* I've recently been reading through our issues, and it seems there are a lot of reports of fragile timing/bad performance for SPI. Particularly, it affects timing-dependent drivers (e.g. neopixels), and might bring buffer overflows or significant slowdowns at high frequencies.

  I've just opened #772, which brings DMA transfers for `embedded-hal` SPI (and later on, I2C and UART) implementations. It would be great if you could test the new DMA impls with the pygamer display, and perhaps the [recently removed neopixel examples](https://github.com/atsamd-rs/atsamd/commit/f0e93162381ea6091bf8c8a159f019e14205893f).

  I don't expect the updated examples or any DMA stuff to make their way into this PR (depends on which will get merged first - probably this one), but it would definitely help build confidence in the implementation, and hopefully enable us to bring the neopixel examples back in a subsequent PR.

boards/pygamer/Cargo.toml
```diff
@@ -65,15 +57,3 @@ chip = "ATSAMD51J19A"
 [[example]]
 name = "usb_poll"
 required-features = ["usb"]
-
-[[example]]
-name = "sd_card"
```
jbeaurivage, line 70: Did you remove these examples from the manifest because they no longer have required features?

I think as a bear practice we should declare all examples in Cargo.toml, even those which don't require extra features.

kyp44: Previously, only examples that required features where listed in `Cargo.toml`. Per the above, several features are no longer used in the actual library (the dependencies were simply moved to the dev dependencies) so were removed. The examples that needed those features were removed from the manifest. If we want to declare them all in the manifest as best practice, I certainly have no problem adding them back in.

jbeaurivage: Yes, let's declare them all in the manifest. Thanks!

kyp44: Completed.

Issue #751, opened 2024-08-28
SpiBus implementation enhancements


ianrrees:
Really just notes from a few hours spent with the SPI-driven WS2812 (neopixel) crate.

1. We currently implement SpiBus only for Duplex SPIs, however it could be nice to provide implementations for Tx (and Rx?) SPIs as well, PyGamer provides two examples of peripherals that could use this feature: the WS2812 and the LCD.  It's not always possible to provide a non-connected pin to the SPI ctor, and the Tx implementation could be more efficient than a Duplex one.  The impl of `SpiBus::read()` for Tx SPI should probably panic with an appropriate message.
2. The implementation we've got now does not take advantage of the pipelines in the SERCOM, so in some cases it stalls between words and so is considerably slower than it could be.  For example, in the `SpiBus::write()` impl we have an implementation that doesn't send word N+1 OUT until N has shifted IN:
``` Rust
for word in words {
    let _ = self.transfer_word_in_place(*word)?;
}
```
but something a little more complicated can have considerably higher throughput, especially when the SPI clock is a higher fraction of the CPU clock (tinkering with a SAMD51, I'm seeing twice as fast in release mode with a 50MHz SPI):
``` Rust
let mut read_count = 0;
for word in words {
    self.block_on_flags(Flags::DRE)?;
    self.config.as_mut().regs.write_data(word.as_());

    if self.read_flags().contains(Flags::RXC) {
        read_count += 1;
        let _ = unsafe { self.read_data().as_() };
    }
}

while read_count < words.len() {
    self.block_on_flags(Flags::RXC)?;
    read_count += 1;
    let _ = unsafe { self.read_data().as_() };
}
```
3. For higher-bandwidth transfers, we could think about an implementation that uses DMA.  This has an additional benefit that it should precisely control the timing between words, in a way that won't depend on the optimization settings of the compiler.  That would help with peripherals like the WS2812, though I'm not sure it would be adequate to fix that particular implementation.

ianrrees:
Thinking more about 1 above; since that behaviour might be a bit unexpected and wouldn't be caught at compile time, it might be better to wrap the unidirectional `Spi<_, Tx>` in a new `PanicOnRead` type which implements `SpiBus`.  We could prototype that in a BSP, which might be good enough as it's rather niche.

kyp44:
[...]

PR #772, opened 2024-10-23
DMA-enabled `embedded-hal` blocking implementations

### LINKED ISSUE #657: `Spi` implementation relies on fragile timing
I'm working on a project where a Feather M4 board connects to a Wiznet W5500 Ethernet chip via SPI for reading and processing raw Ethernet frames. The ATSAMD is running at 120 MHz and the SPI Master via SERCOM1 is operated at 20 MHz.

While generally all parts reached a working state by themselves (SPI communication, USB ACM, GPIOs and a custom SysTick based timer) I soon started seeing issues which were hard to diagnose. The SPI SERCOM started reporting overflow errors after some time (after 10-60s into operation), and that only when building with the release profile, unoptimized seemed to work.
After days of digging I've realized that my SysTick exception (at only 1ms rate, doing nothing but an increment of a SW counter) could influence the timing enough for SPI `write` / `transfer` to break! I expect any other interrupt could do the same.

The issue seems to be the implementation of Duplex `write` and `transfer` within `src\sercom\spi\impl_ehal_thumbv7em.rs`.
As they are really similar, let's take `transfer` as the base for discussion:

```rust
#[inline]
fn transfer<'w>(&mut self, words: &'w mut [Word<$Length>]) -> Result<&'w [Word<$Length>], Error> {
    let cells = core::cell::Cell::from_mut(words).as_slice_of_cells();
    let mut to_send = cells.iter();
    let mut to_recv = cells.iter();
    while to_recv.len() > 0 {
        let flags = self.read_flags_errors()?;
        if to_send.len() > 0 && flags.contains(Flags::DRE) {
            let word = match to_send.next() {
                Some(cell) => cell.get(),
                None => unreachable!(),
            };
            self.config.as_mut().regs.write_data(word as u32);
        }
        if to_recv.len() > to_send.len() && flags.contains(Flags::RXC) {
            let word = self.config.as_mut().regs.read_data() as Word<$Length>;
            match to_recv.next() {
                Some(cell) => cell.set(word),
                None => unreachable!(),
            }
        }
    }
    Ok(words)
}
```
The `while` loop is there to ensure that we read as many words as we write. This alone however is not sufficient. As the SERCOM only offers a buffer for one word, reading may not limp behind more than one word. Let's call this situation `(-1)`.
To understand the issue I recorded the contents of `INTFLAGS` of the SERCOM in response to sending just one single word (by writing the `DATA` register). It turns out to be: `[0, ..., 1, ..., 5, ..., 7, ...]` where `DRE=1, TXC=2, RXC=4`.

As you can see, the `DRE` bit is set first, so depending on execution times (optimization, interrupts, ...) it is unpredictable what happens. Assume we start in the first iteration of the `while` loop:

- `Iteration 1`: First `if` hits because `DRE` is set => transmit first word. Second `if` does not hit as `RXC` was clear at time of sampling. We're one read behind (-1).
- `Iteration N>1`: now it depends at what time we take the next snapshot of the `INTFLAGS`
  1. Possibility: Both, `DRE` and  `RXC` are observed, and the first `if` hits and we transmit (-2) but shortly after (hopefully) the second `if` hits and we manage to read a word to get back to (-1) avoiding overflow.
  2. Possibility: Only `DRE` is observed -> same situation as `Iteration 1`. We're now (-2) and in the following iterations we'll have to catch up reading with even closer timing to avoid the overflow.

It appears the existing code is quite sensitive to timing variations - one could say it has real-time requirements of its own.
I was able to write a workaround for my project using a thin newtype as wrapper:

```rust
impl<C, A> Transfer<u8> for SafeSpi<C, A>
where
    C: ValidConfig,
    A: Capability,
{
    type Error = Error;

    fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Self::Error> {
        let cells = Cell::from_mut(words).as_slice_of_cells();
        let mut to_send = cells.iter();
        let mut to_recv = cells.iter();
        while to_recv.len() > 0 {
            if to_send.len() > 0 {
                while !self.0.read_flags_errors()?.contains(Flags::DRE) {}
                let word = match to_send.next() {
                    Some(cell) => cell.get(),
                    None => unreachable!(),
                };
                unsafe { self.0.write_data(word as u32) };
            }
            if to_recv.len() > to_send.len() {
                while !self.0.read_flags_errors()?.contains(Flags::RXC) {}
                let word = (unsafe { self.0.read_data() }) as u8;
                match to_recv.next() {
                    Some(cell) => cell.set(word),
                    None => unreachable!(),
                }
            }
        }
        Ok(words)
    }
}
```

This solved the issue in my case so I'm sharing these findings. I believe a conservative implementation like this should be shipped with the crate, as a safer default, even though I see a few subtle issues:

- The existing implementation avoids dead times by "streaming" new data, while my implementation waits for RXC before pushing another word, introducing a delay between words.
- If a constant stream of words is not upheld, will the SPI SERCOM assume "end of transfer" prematurely and unassert CS when using Hardware Slave Select? If so it would probably render HWSS useless since some SPI clients rely on CS active during the entire transaction. Only software-driven CS would be possible. Maybe not an issue but needs documentation.
- Perhaps this can't even be made reliable enough without involving DMA?

I'm willing to help improve the implementation but I feel like I neither have enough experience nor the tools to analyze the outcome at the waveform level.

(Find [attached](https://github.com/atsamd-rs/atsamd/files/10348691/record_spi_flags.rs.txt) the Rust program I used for recording the `INTFLAGS` register after sending a word. This is all assembly but SERCOM1 must be configured (in Rust) prior to running this.)

Crates used:

```
embedded-hal = "0.2.7"
embedded-nal = "0.6.0"
embedded-time = "0.12.1"
cortex-m = { version = "0.7.6", features = ["critical-section-single-core"] }
cortex-m-rt = "0.7.2"
feather_m4 = { version = "0.10.1", features = ["usb"] }
atsamd-hal = { version = "0.15.1" }
usb-device = "0.2.9"
usbd-serial = "0.1.1"
w5500 = { path = "priv/w5500" } # unpublished master w/ raw Ethernet frame reading
critical-section = "1.1.1"
heapless = "0.7.16"
```

jbeaurivage:
### SPI timing and performance issues
In #657, @Felix-El has pointed out that the blocking SPI implementations were somewhat fragile and prone to race conditions. @ianrrees  also had some similar comments in #751. Essentially, our SPI transfers are done word by word, which might:

* Introduce buffer overflow errors if the SPI is ran at a too high frequency
* Break invariants of some timing-dependent protocols build on top of SPI, such as LCD displays and addressable LEDs.

### SPI transfers using DMA

Currently, the HAL has minimal support for making SPI transfers using DMA. It exposes the `send_with_dma` and `receive_with_dma` methods. However these:
1. Are non-blocking
1. Take exclusively `'static` buffers in order to satisfy memory safety invariants (namely, due to point 1)
1. Most importantly, can not be used to build [`embedded-hal::spi::SpiBus`](https://docs.rs/embedded-hal/latest/embedded_hal/spi/trait.SpiBus.html) trait implementations.

The same points are also true for our UART and I2C drivers.

## Proposal

In this PR, I suggest a new implementation to make DMA-enabled SPI transfers. It provides blocking, safe implementations that don't need to take `'static` buffers. The API to use DMA is as simple as configuring 2 DMA channels, then calling `Spi::with_dma_channels`, and all subsequent transfers will use DMA. 

What I propose is that we encourage users who have strict timing or performance requirements, or need very high frequency transfers, to use the DMA implementations. The word-by-word implementation can therefore remain simple, without the need for us to expend lots of energy trying to optimize it for performance. Users with loose timings or lower frequencies can still use it with reduced configuration complexity (which is really not a lot).

The rationale is that we can probably optimize the word-by-word implementation a bit, but we'll soon hit an upper limit, which we can only exceed using DMA anyways. DMA transfers are also impervious to preemption by higher priority interrupts.

I want to add the same mechanisms for I2C and UART, since most of the effort is already done and reusable there.

## What's included

  - [x] DMA-enabled blocking SpiBus implementation
  - [x] Mark the various `send_with_dma` and `receive_with_dma` methods as deprecated. They should be removed either in this PR or in a subsequent release -- perhaps at the same time as the ehal 0.2 stuff. Keep the UART `send_with_dma` and `receive_with_dma` nonblocking methods, as they can be used on a split UART to do some interesting things.
  - [x] ~~Configurable NOP word for SPI transfers~~ Turns out this was already implemented
  - [x] DMA-enabled blocking I2c implementation
  - [x] DMA-enabled blocking Uart implementation

#### SPI-specific improvements

  - [x] Disable receiver when the only capability is TX - reduces changes of unnecessary overflow errors. Reference: #657 ([see comment](https://github.com/atsamd-rs/atsamd/issues/657#issuecomment-1372199143)).
  - [x] Provide SpiBus implementations for RX-only and TX-only SPIs (DMA and non-DMA), which panic when unsupported operations are attempted. References: #752, #751 ([see comment](https://github.com/atsamd-rs/atsamd/issues/751#issuecomment-2344634018)).

## What's not included

* `embedded-hal-nb` DMA implementations. These would either require the user to pass `'static` buffers everywhere, or require all DMA-enabled methods to be unsafe. Both options are incompatible with the traits `embedded-hal-nb` provides. For non-blocking concurrent code, I would strongly suggest users look at async instead (#635).

* There have also been reports of fragile timing and performance in our blocking `embedded-hal v0.2` implementations. I'm not particularly interested in updating those. In fact, I would like to remove embedded-hal v0.2 from atsamd-hal entirely (perhaps in another PR).  

## Breaking changes

* Bumps MSRV to Rust 1.77.2

## Checklist

## Notes for reviewers:

* This PR is built on top of #764, which should be merged first.

* Please merge this PR using rebase (and not squash) to conserve changelog history.

Closes #657.

### DIFF
--- a/boards/metro_m0/examples/spi.rs
+++ b/boards/metro_m0/examples/spi.rs
@@ -1,4 +1,4 @@
-//! This example showcases the spi module, and uses DMA to perform SPI
+//! This example showcases the spi module, and uses DMA to perform I2C
 //! transactions.
 
 #![no_std]
@@ -9,7 +9,7 @@ use panic_halt as _;
 #[cfg(feature = "use_semihosting")]
 use panic_semihosting as _;
 
-use metro_m4 as bsp;
+use metro_m0 as bsp;
 
 use bsp::entry;
 use bsp::hal;
@@ -25,31 +25,32 @@ use hal::fugit::RateExtU32;
 #[entry]
 fn main() -> ! {
     let mut peripherals = Peripherals::take().unwrap();
-    let mut clocks = GenericClockController::with_external_32kosc(
+    let mut clocks = GenericClockController::with_internal_32kosc(
         peripherals.gclk,
-        &mut peripherals.mclk,
-        &mut peripherals.osc32kctrl,
-        &mut peripherals.oscctrl,
+        &mut peripherals.pm,
+        &mut peripherals.sysctrl,
         &mut peripherals.nvmctrl,
     );
 
+    let mut pm = peripherals.pm;
     let dmac = peripherals.dmac;
     let pins = bsp::Pins::new(peripherals.port);
 
     // Take SPI pins
-    let (miso, mosi, sclk) = (pins.miso, pins.mosi, pins.sclk);
+    let (miso, mosi, sclk) = (pins.miso, pins.mosi, pins.sck);
 
     // Setup DMA channels for later use
-    let mut dmac = DmaController::init(dmac, &mut peripherals.pm);
+    let mut dmac = DmaController::init(dmac, &mut pm);
     let channels = dmac.split();
     let chan0 = channels.0.init(PriorityLevel::Lvl0);
     let chan1 = channels.1.init(PriorityLevel::Lvl0);
 
+    // Create a Spi with DMA enabled
     let mut spi = bsp::spi_master(
         &mut clocks,
         100.kHz(),
-        peripherals.sercom2,
-        &mut peripherals.mclk,
+        peripherals.sercom4,
+        &mut pm,
         sclk,
         mosi,
         miso,
--- a/boards/atsame54_xpro/Cargo.toml
+++ b/boards/atsame54_xpro/Cargo.toml
@@ -39,9 +39,16 @@ rtt-target = { version = "0.3", features = ["cortex-m"] }
 
 [features]
 default = ["rt", "atsamd-hal/same54p"]
+dma = ["atsamd-hal/dma"]
 rt = ["cortex-m-rt", "atsamd-hal/same54p-rt"]
 usb = ["atsamd-hal/usb", "usb-device"]
 can = ["atsamd-hal/can"]
+
+[[example]]
+name = "blinky_basic"
+
+[[example]]
+name = "blinky_rtic"
 
 [[example]]
 name = "mcan"
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -127,8 +127,19 @@ required-features = ["rtic"]
 
 [[example]]
 name = "uart"
+
+[[example]]
+name = "uart_dma_nonblocking"
+required-features = ["dma"]
+
+[[example]]
+name = "uart_dma_blocking"
 required-features = ["dma"]
 
 [[example]]
 name = "i2c"
+required-features = ["dma"]
+
+[[example]]
+name = "spi"
 required-features = ["dma"]
--- a/boards/feather_m0/examples/i2c.rs
+++ b/boards/feather_m0/examples/i2c.rs
@@ -1,4 +1,5 @@
-//! This example showcases the i2c module.
+//! This example showcases the i2c module, and uses DMA to perform I2C
+//! transactions.
 
 #![no_std]
 #![no_main]
@@ -23,8 +24,9 @@ use hal::ehal::i2c::I2c;
 use hal::fugit::RateExtU32;
 use hal::sercom::i2c;
 
-const LENGTH: usize = 1;
-const ADDRESS: u8 = 0x77;
+// This example is based on the BMP388 pressure sensor. Adjust the device and
+// register addresses to your liking
+const ADDRESS: u8 = 0x76;
 
 #[entry]
 fn main() -> ! {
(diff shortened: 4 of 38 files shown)


boards/metro_m4/examples/spi.rs
```diff
@@ -1,28 +1,29 @@
+//! This example showcases the i2c module.
```
sajattack, line 1: does it?

boards/metro_m0/examples/spi.rs
```diff
@@ -0,0 +1,82 @@
+//! This example showcases the i2c module.
```
sajattack, line 1: nother one

sajattack (approved):
couple comment typos I trust you'll resolve but LGTM

jbeaurivage:
Thanks for reviewing @sajattack.

There's just one last thing I'm wondering about before I want to close this for good.

In the ehal 0.2 SPI implementation, we (obviously) wait for a DRE every time we want to send a word on the bus, but we never wait for a TXC, even at the end of a transaction.

The new implementation *does* wait for a TXC after each transaction. I believe that may be unnecessary, according to embedded-hal's SpiBus spec, and potentially even harmful - I suspect it may be a contributor to the timing issues we've been experiencing with SPI. After rereading the spec, turns out flushing should only be necessary when deasserting CS or deinitializing the peripheral - the user should flush manually. 

I think I'll remove those bus flushes and test out a bit more before committing to merging this.

jbeaurivage:
I've decided to leave the bus flushes alone for now. Seems like removing them introduces some subtle timing bugs, and sometimes even panics. I want to investigate further before touching, this, perhaps in a later PR.

PR #774, opened 2024-10-25
Update svd2rust to 0.33.5 and re-generate PACs

maximeborges:
# Summary
- Updating `svd2rust` to `0.33.5`
  - Fixes `unsupported proc macro punctuation character '{'` when rebuilding PACs via `update-pacs.sh`
- Move `GMAC::TIDM::ENIDn` into `atsamd51.xsl` (already present in 53/54 SVDs)


jbeaurivage:
Thank you for this. I was actually wondering why the ENID fields were getting duplicated for some chips. I didn't find anything suspicious after digging through the SVDs - I didn't realize the bug was coming from our own XSL patches.

PR #777, opened 2024-10-31
Upgrade `pygamer` BSP display driver and graphics dependencies

kyp44:
# Summary
Most of the `pygamer` BSP dependencies were upgraded with #766, which was merged. The remaining dependencies are upgraded to the latest version with this PR:

- `st7735-lcd` to v0.10.0
-  `embedded-graphics` to v0.8.1
-  `tinybmp` to v0.6

# Important Note
These dependency upgrades depend on #772, which has already been merged into the `pygamer-upgrade-graphics` branch but not yet into `master`. Hence, this PR should probably not be merged until #772 has been merged first.

# Checklist

### DIFF
--- a/boards/pygamer/Cargo.toml
+++ b/boards/pygamer/Cargo.toml
@@ -47,6 +47,8 @@ usbd-serial = "0.2"
 [features]
 # ask the HAL to enable atsamd51j support
 default = ["rt", "atsamd-hal/samd51j"]
+dma = ["atsamd-hal/dma"]
+max-channels = ["dma", "atsamd-hal/max-channels"]
 panic_led = []
 rt = ["cortex-m-rt", "atsamd-hal/samd51j-rt"]
 usb = ["atsamd-hal/usb", "usb-device"]
--- a/boards/pygamer/Cargo.toml
+++ b/boards/pygamer/Cargo.toml
@@ -16,7 +16,9 @@ version = "0.11.0"
 
 [dependencies]
 cortex-m = {version = "0.7", features = ["critical-section-single-core"]}
-st7735-lcd = "0.8.1"
+embedded-hal-bus = "0.2.0"
+# This version is pinned as recommended by: https://docs.rs/embedded-hal-bus/0.2.0/embedded_hal_bus/spi/struct.ExclusiveDevice.html#method.new_no_delay
+st7735-lcd = "=0.10.0"
 
 [dependencies.cortex-m-rt]
 optional = true
@@ -29,18 +31,17 @@ version = "0.18.2"
 
 [dependencies.usb-device]
 optional = true
-version = "0.3.1"
+version = "0.3.2"
 
 [dev-dependencies]
-embedded-graphics = "0.7.1"
-embedded-hal-bus = "0.2.0"
+embedded-graphics = "0.8.1"
 embedded-sdmmc = "0.8.0"
 lis3dh = "0.4.3"
 micromath = "2.1"
-panic-halt = "0.2"
+panic-halt = "1"
 rtic = {version = "2.1.1", features = ["thumbv7-backend"]}
 smart-leds = "0.4"
-tinybmp = "0.3.1"
+tinybmp = "0.6"
 usbd-serial = "0.2"
 
 [features]
--- a/boards/pygamer/src/pins.rs
+++ b/boards/pygamer/src/pins.rs
@@ -582,13 +582,22 @@ pub struct Display {
     pub tft_backlight: TftBacklightReset,
 }
 
-/// This empty error occurs if there is an issue setting up the on-board
-/// display.
+/// Error that can occur when initializing the display.
 #[derive(Debug)]
-pub struct DisplayError;
+pub enum DisplayError {
+    /// Could not configure the SERCOM4 clock.
+    SercomClock,
+    /// Could not configure the SPI port to drive the display.
+    Spi,
+    /// Could not setup the ST7735 display driver.
+    Driver,
+    /// Could not configure the TC2/TC3 clock for PWM control of the backlight.
+    Tc2Tc3Clock,
+}
 impl From<()> for DisplayError {
+    #[inline]
     fn from(_value: ()) -> Self {
-        DisplayError
+        Self::Driver
     }
 }
 
@@ -615,7 +624,9 @@ impl Display {
         delay: &mut hal::delay::Delay,
     ) -> Result<(DisplayDriver, Pwm2<PA01>), DisplayError> {
         let gclk0 = clocks.gclk0();
-        let clock = &clocks.sercom4_core(&gclk0).ok_or(DisplayError)?;
+        let clock = &clocks
+            .sercom4_core(&gclk0)
+            .ok_or(DisplayError::SercomClock)?;
         let pads = spi::Pads::default()
             .sclk(self.tft_sclk)
             .data_out(self.tft_mosi);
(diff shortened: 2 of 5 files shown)


kyp44:
I also just noticed that `README.md` lists `pygamer` as a BSP for the [`atsame53j`](https://docs.rs/atsame53j/), but the PyGamer actually features a ATSAMD51J19 microcontroller ([source](https://www.adafruit.com/product/4242)), which is also reflected in the actual BSP. This PR branch now corrects this, as well as marks it as Tier 1.

sajattack:
Thanks for contributing. I'll revisit this after #772 merges

sajattack (approved):

sajattack (changes requested):
you need to update crates.json to make it tier 1

sajattack:
Oh that was done in another pr?

sajattack (dismissed):

boards/pygamer/src/pins.rs
```diff
@@ -581,30 +582,54 @@ pub struct Display {
...
+/// This empty error occurs if there is an issue setting up the on-board
+/// display.
+#[derive(Debug)]
+pub struct DisplayError;
+impl From<()> for DisplayError {
+    fn from(_value: ()) -> Self {
+        DisplayError
+    }
+}
+
```
jbeaurivage, line 594: I feel like this is not descriptive enough. It does tell the user that an error occured, but it doesn't help determining what failed or why. Plus, it seems like all errors that may occur are of the same type, except for two of them that come from completely different sources:

* Setting up the clocks:
   Typically, the BSPs unwrap on an error when trying to setup a `sercom_*` or `tc*` clock. I don't think there's really a reasonable way to recover from this anyway, so `unwrap`ping is completely fine. Or better yet, using `expect`.

* Initializing the `ExclusiveDevice` struct:
   I seems like the only way `ExclusiveDevice::new_no_delay` can fail is if there's an error setting the CS pin to HIGH on initialization. For `atsamd-hal`, setting the state of an output pin is infaillible; this error will never happen. Again, I think `unwrap`ping or `expect` ing is perfectly ok here.

That means that the user will know for sure that the error comes from the display without having to add variants to your error enum, and not some other, miscellanous thing occured, like the clocks having been setup wrong.

It's just a personal preference though, and there's nothing inherently wrong with this approach, so feel fre to apply my suggestion or not.

boards/pygamer/Cargo.toml
```diff
@@ -29,18 +31,17 @@ version = "0.18.2"
...
+embedded-graphics = "0.8.1"
 embedded-sdmmc = "0.8.0"
 lis3dh = "0.4.3"
 micromath = "2.1"
-panic-halt = "0.2"
+panic-halt = "1"
 rtic = {version = "2.1.1", features = ["thumbv7-backend"]}
 smart-leds = "0.4"
-tinybmp = "0.3.1"
+tinybmp = "0.6"
```
jbeaurivage, line 44: For the sake of completeness, could you add these features to Cargo.toml:

```toml
[features]
dma = ["atsamd-hal/dma"]
max-channels = ["dma", "atsamd-hal/max-channels"]
```

jbeaurivage:
> Oh that was done in another pr?

Yep, already taken care of!

kyp44:
> Oh that was done in another pr?

Yup, looks like it was bumped to Tier 1 in another PR at some point, though it was not me who did it (didn't even know about the `crates.json` file)

boards/pygamer/src/pins.rs
kyp44: So, previously, `pygamer::pins::Display::init` returned a `Result<_, ()>`. This now triggers [this clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#result_unit_err), so, in the interest of not introducing new clippy lints (technically I suppose it wasn't my changes that triggered this), I added the empty `DisplayError` type.

So we have three options for the return type of `pygamer::pins::Display::init`:
1. Return the `Result<_, DisplayError>` as it currently does.
2. Keep it the way it was, returning `Result<_, ()>` and suppress the lint.
3. Simply return the pair `(DisplayDriver, Pwm2<PA01>)` instead of a `Result` and unwrap/panic the two errors above.

I have no particular preference here, but note that options 1 and 3 are breaking API changes, not that that matters too much since `pygamer` is not yet at version 1.0.0. I would comment though that, since this is library code, it might be best to let the end user decide whether failing to setup the `sercom` clock (and therefore the display) is recoverable or not. They could also just want to handle that problem in a way different from a panic.

What are everyone's thoughts here?

boards/pygamer/Cargo.toml
kyp44: I actually wondered whether these new features should be exposed in the BSPs, but I checked and noticed that there are other HAL features that are not exposed in Tier 1 BSPs, so I wasn't sure what determines whether a feature should be exposed or not. As a specific example, the HAL `defmt` feature is not exposed in the Tier 1 `metro_m4` BSP. Obviously, features are exposed if they are used in the BSP examples.

Adjacent question: Do we want to use DMA in any of the `pygamer` examples that use SPI? I noticed you did so for other Tier 1 BSPs. My thought was that for sure I would have the neopixels example use DMA SPI, but that is waiting on [this PR](https://github.com/smart-leds-rs/ws2812-spi-rs/pull/39) as we discussed @jbeaurivage.

boards/pygamer/src/pins.rs
jbeaurivage: Again, I don't really have a dog in the fight, and I don't want to start bikeshedding here. I'm just saying that as a user, I would be surprised to see a `DisplayError` which is actually a clock setup error. I would then suggest changing the `DisplayError` definition to something similar to this:

```rust
pub enum DisplayError {
    SercomClockMisconfigured,
    DriverError,
}
```

which is more in line with the principle of least surprise. I will say though that I'm surprised that all the functions in `st7735-lcd` return `Result<(), ()>`. That API certainly isn't the best at helping their users figure out what exactly went wrong.

boards/pygamer/Cargo.toml
jbeaurivage: Reexporting the HAL features are essentially there for 3 reasons:

1. Obviously, if examples use the features, we need to reexport them
1. Even if no example uses them, having them reexported is convenient for the user; otherwise they'd have to specify `atsamd-hal` in their deps with the features they want, potentially leading to version mismatches
1. A more minor reason, but most of our CI scripts call `cargo clippy --examples --all-features` for each BSP. Reexporting the features helps us get a little more coverage in our test builds.

The fact that they aren't reexported in all T1 BSPs is just an oversight. I could always open a PR/issue eventually to get that sorted out.

Up to you to use DMA or not in the examples. I'm not opposed to having a little variety in our examples across BSPs. Again, more API surface being tested (if people do run the examples!), and it gives newcomers a glimpse into what is and isn't possible to do with the HAL.

There has been some discussion around standardizing the BSP examples, but that's certainly more than I'm willing to chew right now!

boards/pygamer/src/pins.rs
sajattack: 😅

boards/pygamer/Cargo.toml
kyp44: With the latest commit, exposed these features for convenience. Only two current examples use SPI, so I'm going to hold off on demoing DMA SPI until the neopixel example is added pending the `ws2812-spi-rs` PR.

Not attempting to tackle the more general issue of exposing all HAL features in all the Tier 1 BSPs.

PR #780, opened 2024-11-09
feat(spi): Unlock DMA transfers for SpiBus::transfer_in_place

jbeaurivage:
Fixes unsoundness issues caused by mishandled memory barriers in the `dmac` module.

These were causing observable undefined behaviour in the SpiBus::transfer_in_place implementation when trying to use DMA. The implementation was therefore defaulted to using word-by-word transfers in #772.

Now that the bug is fixed, we also unlock DMA in-place transfers for SPI. I've tested the changes and can confirm that there is no longer exhibited unsoundness.

This PR also brings minor refactors to the `dmac` module, and fixes warnings when compiling with the `dma` Cargo feature turned off.

# Checklist


hal/src/sercom/spi/impl_ehal/dma.rs
```diff
@@ -307,13 +307,18 @@ where
 
     #[inline]
     fn transfer_in_place(&mut self, words: &mut [C::Word]) -> Result<(), Self::Error> {
-        for word in words {
-            let read = self.transfer_word_in_place(*word)?;
-            *word = read;
+        // Safefy: Aliasing the buffer is only safe because the DMA read will always be
```
sajattack, line 310:
```suggestion
        // Safety: Aliasing the buffer is only safe because the DMA read will always be
```

PR #797, opened 2024-12-03
Allow the use of any GCLK with CAN

rnd-ash:
# Summary
This makes it possible to setup any Gclk with the MCAN API, rather than being limited to always using Gclk0, which might not be possible depending on the application.

# Checklist


jbeaurivage:
I wonder why this was restricted to GCLK0 only in the first place? At least from reading the datasheet, the peripheral does indeed seem to support any GCLK source.

PR #802, opened 2024-12-07
examples(pygamer): Restore neopixel examples using SPI driver

### LINKED ISSUE #752: Neopixel examples are problematic

jbeaurivage:
# Summary
Reinstate the neopixel examples in the pygamer BSP. The examples now use the `ws2812-spi` driver instead of bitbanging the protocol. Note that the `neopixel_tilt` example could not be restored, because creating an `Spi` struct requires a clock pin, even though it's not used by the LEDs. For the pygamer, that pin can be one of SDA or SCL, both of which are required to talk to the accelerometer over I2C.

# Checklist
  - [x] No new warnings or clippy suggestions have been introduced

Closes #752.

### DIFF
--- a/boards/pygamer/Cargo.toml
+++ b/boards/pygamer/Cargo.toml
@@ -33,6 +33,11 @@ version = "0.20.0"
 optional = true
 version = "0.3.2"
 
+[dependencies.ws2812-spi]
+version = "0.5.0"
+features = ["mosi_idle_high"]
+optional = true
+
 [dev-dependencies]
 embedded-graphics = "0.8.1"
 embedded-sdmmc = "0.8.0"
@@ -52,6 +57,7 @@ max-channels = ["dma", "atsamd-hal/max-channels"]
 panic_led = []
 rt = ["cortex-m-rt", "atsamd-hal/samd51j-rt"]
 usb = ["atsamd-hal/usb", "usb-device"]
+neopixel-spi = ["dep:ws2812-spi"]
 # Enable async support from atsamd-hal
 async = ["atsamd-hal/async"]
 
@@ -89,3 +95,23 @@ name = "timer"
 [[example]]
 name = "usb_poll"
 required-features = ["usb"]
+
+[[example]]
+name = "neopixel_adc_battery"
+required-features = ["neopixel-spi"]
+
+[[example]]
+name = "neopixel_adc_light"
+required-features = ["neopixel-spi"]
+
+[[example]]
+name = "neopixel_button"
+required-features = ["neopixel-spi"]
+
+[[example]]
+name = "neopixel_easing"
+required-features = ["neopixel-spi"]
+
+[[example]]
+name = "neopixel_rainbow"
+required-features = ["neopixel-spi"]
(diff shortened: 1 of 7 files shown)


jbeaurivage:
@kyp44, I'd like to solicit your help testing these examples on a pygamer if you don't mind.

kyp44:
@jbeaurivage  I'll definitely test these when I get a chance. Do we not need to wait for [this PR](https://github.com/smart-leds-rs/ws2812-spi-rs/pull/39) to be merged though?

jbeaurivage:
Thanks! And nope, these examples work with the mainstream release of ws2812-spi. That PR is to help with DMA SPI transactions, whereas these examples don't need DMA to work. Just compiling with optimizations enabled.

ianrrees:
I've just had a play with this branch on my pygamer and can confirm all the neopixel examples work (except the battery one, as I don't have a battery handy).

My memory is fuzzy but think I found the ehal SPI traits don't provide strong enough guarantees to meet the neopixel timing requirements.  Since we control the SPI implementation here too, maybe that's not a problem, but longer term it might be good to use neopixels as an example of how to write a peripheral driver outside of atsamd-rs - using the TC/TCC or similar to control the timing.

jbeaurivage:
Since hardware tests are conclusive, this is ready to be merged.

kyp44:
@jbeaurivage Sorry, I was busy with life/kids stuff all weekend, but I just tested all the examples. As @ianrrees found, they all work really well, including the battery example. My PyGamer battery happened to be low so I watched the NeoPixels meter go up as I let it charge. The button example was especially neat!

Once the [PR over at `ws2812-spi-rs`](https://github.com/smart-leds-rs/ws2812-spi-rs/pull/39) is merged, I think it would be great to add or convert an example to use DMA. I re-incorporated neopixels into my stress test example and, interestingly, even without using DMA, the neopixels now seem to work much better even when there are a lot of interrupts going on, though I still observed the colors glitching from time to time, presumably when the SPI write gets interrupted. I am hoping using DMA will help with this. In any case, this is generally a big improvement over what we had before so kudos!

PR #803, opened 2024-12-10
Allow the construction of Delay using the clock::v2 API

rnd-ash:
This PR introduces `Delay::new_with_source` function, which is compatible with the V2 clocking API, allowing the usage of Delay after calling `clock_system_at_reset` (Which is required for MCAN)

Example usage:

```rust
let (pclk_can0, gclk0) = pclk::Pclk::enable(tokens.pclks.can0, clocks.gclk0);
let (mut delay, gclk0) = Delay::new_with_source(cpu_core.SYST, gclk0);
```

### DIFF
--- a/hal/src/delay.rs
+++ b/hal/src/delay.rs
@@ -8,6 +8,8 @@ use crate::clock::GenericClockController;
 use crate::ehal::delay::DelayNs;
 use crate::ehal_02;
 use crate::time::Hertz;
+
+#[hal_cfg("rtc-d5x")]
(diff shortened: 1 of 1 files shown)


jbeaurivage:
@rnd-ash, is there a reason why we'd want to restrict the clock source to GCLK0 only? According to the datasheet, the CAN peripheral should work with any GCLK.

rnd-ash:
ok @jbeaurivage all tests pass now

Issue #765, opened 2024-10-03
RTIC v2.0.0 support


kyp44:
Hi, I'm new here. I'm working on a game targeting the PyGamer, which has a nice BSP here. Being a game, using async Rust is extremely valuable for me, and information on using async on SAMD processors seems to be very scare. It doesn't seem like Embassy supports SAMD processors at all, but RTIC v2 now uses async for software tasks, whereas v1 did not.

The HAL crate here does support RTIC (via the `rtic` feature), but it only seems to support RTIC v1 by implementing the old [`Monotonic`](https://docs.rs/rtic-monotonic/latest/rtic_monotonic/trait.Monotonic.html) trait. Things have changed with v2 and now there is a different [`Monotonic`](https://docs.rs/rtic-time/2.0.0/rtic_time/trait.Monotonic.html) trait that has to be implemented, which involves some async functions.

The current paradigm for RTIC v2 is to select the time driver from the [`rtic_monotonics`](https://docs.rs/rtic-monotonics/latest/rtic_monotonics/) crate, in which the type that implements [`Monotonic`](https://docs.rs/rtic-time/2.0.0/rtic_time/trait.Monotonic.html) is created using macros instead of passing it to the `app` RTIC macro as before. Looking at the code for the supported microcontrollers (noting that SAMD is not among them), it seems like [`Monotonic`](https://docs.rs/rtic-time/2.0.0/rtic_time/trait.Monotonic.html) is most easily implemented indirectly by instead implementing the simpler [`TimerQueueBackend`](https://docs.rs/rtic-time/latest/rtic_time/timer_queue/trait.TimerQueueBackend.html) trait, which does not involve any async methods.

I think I have enough embedded experience to implement an RTC-based driver for RTIC v2. Given that the v1 driver was implemented in this HAL instead of within the RTIC project, I wanted to solicit opinions on whether a new v2 driver should also just live with the HAL here (including relevant macros), or by adding SAMD support to the RTIC project? Thoughts?

NOTE: I am aware that I can already use RTIC v2 on SAMD chips using the `SysTick` driver that ships with RTIC, but I want a more efficient driver that does not generate frequent, periodic interrupts.

jbeaurivage:
Hi @kyp44 and welcome! 

Ideally the implementation would live in this crate. By all means, feel free to open up a PR with a RTIC v2 monotonic implementation. I've actually been meaning to do it too, just haven't found the time yet - I'd be happy to assist you with that.

[...]

Issue #784, opened 2024-11-13
Remove deprecated code - embedded-hal 0.2


jbeaurivage:
This is a tracking issue for the removal of deprecated code.

- [ ] Remove support for `embedded-hal` v0.2
- [ ] Remove deprecated items in `sercom::dma`
- [ ] Optionnally, when #635 lands, remove the `spi_future` module - Would appreciate @bradleyharden's input on this.

Sympatron:
I wouldn't remove `embedded-hal` v0.2 just yet. It can still be useful. Maybe hiding it behind a feature for now would be enogh.

jbeaurivage:
Fine by me - out of curiosity, what would you consider a good timeline to start removing it?

Sympatron:
I just looked into it further and I think it's not as bad as I thought. I knew some things were missing in v1, because they couldn't settle on an API. But I think they were sufficiently useless in their current state that they probably weren't used anyway. 

Some drivers might still use v0.2, but those could be easily adapted (if they are still maintained...) or you could use [`embedded_hal_compat`](https://docs.rs/embedded-hal-compat/latest/embedded_hal_compat/).

So it's probably safe to remove v0.2 with the next minor version bump.

wt:
Does this issue include the future incompatibilities reported by the cargo/rust?

PR #804, opened 2024-12-21
RTIC v2 RTC monotonic drivers

kyp44:
# Summary
Adds RTC-based RTIC v2 monotonics drivers.

This addresses issue #765 and is the culmination of a lot of discussion there.

Note that it was [decided](https://hackmd.io/c_mFUZL-Q2C6614MlrrxOg?view#Agenda-2024-12-18) along with the RTIC team that ATSAMD RTIC monotonics should be released as part of the HAL instead of in `rtic-monotonics`.

Note that this release of the monotonics is opinionated in the sense that only the best monotonic is provided for each chip variant. See the module documentation of [`rtc::rtic`](https://github.com/kyp44/atsamd/blob/rtic-v2-rtc-monotonics/hal/src/rtc/rtic/mod.rs) for details. For reference, the alternative of releasing two monotonics, one for each RTC mode, is preserved in the [`rtic-v2-rtc-monotonics-full` branch](https://github.com/kyp44/atsamd/tree/rtic-v2-rtc-monotonics-full). Compare the module documentation of [`rtc::rtic` in that branch](https://github.com/kyp44/atsamd/blob/rtic-v2-rtc-monotonics-full/hal/src/rtc/rtic/mod.rs) for details on why there is clearly a best choice for each chip variant.

# Checklist
  - [X] No new warnings or clippy suggestions have been introduced - CI will **deny** clippy warnings by default!


kyp44:
Ok, I _just_ discovered a [stress test](https://github.com/kyp44/pygamer-tests/tree/master/monotonic-stress) freezing issue when building in release mode, which I had not really tested before. Please do not merge until I get this fixed.

jbeaurivage:
@kyp44 does the last commit address all the freezing issues? I'm hoping to have the time to read the PR in detail soon™.

kyp44:
@jbeaurivage Yeah, this does fix the issue, which, despite taking me hours to troubleshoot, was a very simple fix. I am still letting the stress test run for a long period of time, but I am 98% confident that there will not be further issues. I like to wait at least 36 hours to get a hardware counter rollover to be sure. [Here](https://thorsen.pm/timer/1734919557/129600) is a time showing how much time is left. If you want to be sure not to waste any of your time, wait until this timer is done, and I will check in and confirm that it's still going.

hal/src/rtc/modes.rs
```diff
@@ -0,0 +1,337 @@
...
+/// functions.
+#[hal_macro_helper]
+pub trait RtcMode {
+    /// The type of the COUNT register.
+    type Count: Copy + PartialEq + Eq;
+
+    /// Sets this mode in the CTRL register.
+    unsafe fn set_mode(rtc: &Rtc);
+    /// Sets a compare value.
+    unsafe fn set_compare(rtc: &Rtc, number: usize, value: Self::Count);
```
jbeaurivage, line 84: Here, I would suggest adding a `# Safety` section to the doc comments, explaining why these functions are unsafe and what are the invariants that must be upheld.

hal/src/rtc/rtic/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+//! [synchronization delay](https://onlinedocs.microchip.com/oxy/GUID-F5813793-E016-46F5-A9E2-718D8BCED496-en-US-14/GUID-0C52DB00-4BF6-4F41-85B5-B76529875364.html),
+//! the number returned may not always increment by one every time it changes.
+//! In fact, testing shows that it typically increments by four every time it
+//! changes. This is true regardless of the clock rate used, as the
+//! synchronization delay scales along with the clock period.
+
+/// Items for RTIC v1.
+///
+/// TODO: This probably needs to be modernized (e.g. it does not implement
+/// half-period counting) or deprecated/removed.
```
jbeaurivage, line 121: I'm partial to adding a `#[deprecated]` attribute. Then we can yeet it at the next opportune occasion.

hal/src/rtc/rtic/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+use super::modes::{mode0::RtcMode0, mode1::RtcMode1, RtcMode};
+use atsamd_hal_macros::hal_macro_helper;
+
+pub mod prelude {
+    pub use super::rtc_clock;
+    pub use crate::rtc_monotonic;
+    pub use rtic_time::Monotonic;
+
+    pub use fugit::{self, ExtU32, ExtU32Ceil, ExtU64, ExtU64Ceil};
+}
```
jbeaurivage, line 176: I don't want to start bikeshedding here, but I'm not a big fan of preludes in general. Although here, we already reexport `fugit` in `lib.rs`, so there's potential for conflict if a user does this:

```rs
use atsamd_hal::fugit::ExtU32;
use atsamd_hal::rtc::rtic::prelude::*;
```

hal/src/rtc/rtic/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+    ($name:ident, $clock_rate: ty) => {
+        $crate::__internal_create_rtc_struct!($name, RtcBackend, $clock_rate);
+    };
+}
+
+/// This function was copied from the private function in `rtic-monotonics`,
+/// so that should be used when the monotonics move there.
+const fn cortex_logical2hw(logical: u8, nvic_prio_bits: u8) -> u8 {
+    ((1 << nvic_prio_bits) - logical) << (8 - nvic_prio_bits)
+}
```
jbeaurivage, line 318: There is already a very similar function in the HAL ([here](https://github.com/atsamd-rs/atsamd/blob/e281477b438416d176137165b8375db136257839/hal/src/async_hal/interrupts.rs#L495)). If you're up for it, it would be nice to consolidate them to avoid duplication.

hal/src/rtc/rtic/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+
+/// This function was copied from the private function in `rtic-monotonics`.
+unsafe fn set_monotonic_prio(prio_bits: u8, interrupt: impl cortex_m::interrupt::InterruptNumber) {
+    extern "C" {
+        static RTIC_ASYNC_MAX_LOGICAL_PRIO: u8;
```
jbeaurivage, line 323: An extern static is pretty unusual in Rust. I've personally scratched my head for a while on exactly this `RTIC_ASYNC_MAX_LOGICAL_PRIO` thing not so long ago. I would definitely explain in the docs somewhere what this u8 means, and that `rtic` will initialize the static, but users writing rtic-less apps should set the static themselves using

```rs
#[no_mangle]
pub static RTIC_ASYNC_MAX_LOGICAL_PRIO: u8 = (something);
```

jbeaurivage (changes requested):
I have a few comments, some of which require changes, and  some of which are at your discretion. But overall well done, this is some quality work!

Ideally I'd like to somehow get this tested on a thumbv6 target as a sanity check before merging.

hal/src/rtc/rtic/mod.rs
jbeaurivage: If we choose to deprecate the item, its removal should be requested in #784.

jbeaurivage:
Oh also, not a big deal at all, but `#[hal_macro_helper]` is meant to be able to add the `#[hal_cfg]` attribute to blocks and statements inside functions, since that's not currently supported by the compiler (apart from the `#[cfg]` attribute) . For items outside of functions, simply doing `use atsamd_hal_macros::hal_cfg` is sufficient.

hal/src/rtc/modes.rs
kyp44: These particular methods are unsafe because writing to the required registers is unsafe in the PAC (e.g. the [compare registers](https://onlinedocs.microchip.com/oxy/GUID-F5813793-E016-46F5-A9E2-718D8BCED496-en-US-14/GUID-AAD45C69-F58F-477C-ACF8-F3C655D43B73.html)) and this was simply passed up to the enclosing function. However, there are other registers that are safe to write to in the PAC (e.g. the [interrupt flag register](https://onlinedocs.microchip.com/oxy/GUID-F5813793-E016-46F5-A9E2-718D8BCED496-en-US-14/GUID-D193603F-4A13-494B-BCAD-500E9B650382.html)). It is not clear to me why some are safe and others unsafe.

Really this entire low-level interface (i.e. the `RtcMode` trait) is unsafe from the perspective of ensuring that the RTC is used correctly, and not necessary unsafe from a memory perspective. This unsafety is necessary because the interface has to be static because the RTIC monotonic has a static interface. By static I of course mean that all methods are associated functions and no state is maintained.

Given this, I'm not sure what the approach should be. Should every method be marked as unsafe, or should they all be safe? Is `unsafe` in Rust supposed to just refer to memory safety or does safety at the peripheral usage level count? Either way, probably every method should at least have a `# Safety` section to document how the peripheral can be used safely. Thoughts?

hal/src/rtc/rtic/mod.rs
kyp44: 👍 I agree that deprecation makes sense here rather than putting forth the effort to modernize.

I'm assuming that it only makes sense to add the item to #784 once this gets merged.

hal/src/rtc/rtic/mod.rs
kyp44: I think it's fine to remove the RTIC-specific prelude, though I would recommend adding `fugit::ExtU64` and `fugit::RateExtU64` to the overall HAL [prelude](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/prelude.rs) since the `u32` versions are there. I would also opt to add the following to the HAL prelude:

```rust
#[cfg(feature = "rtic")]
pub use rtic_time::Monotonic as _;
```

This is just so the methods of the monotonic can be used without users having to add `rtic_time` as a dependency. An alternative would be to just re-export this trait somewhere rather than having it in the prelude. Thoughts?

hal/src/rtc/rtic/mod.rs
kyp44: The [similar function](https://github.com/atsamd-rs/atsamd/blob/e281477b438416d176137165b8375db136257839/hal/src/async_hal/interrupts.rs#L495) is gated behind the `async` feature. Currently, the `rtic` feature is independent from this, but perhaps it makes sense for `rtic` to also enable `async`?

Also, I wanted to point out that [here](https://github.com/atsamd-rs/atsamd/blob/e281477b438416d176137165b8375db136257839/hal/src/async_hal/interrupts.rs#L444) defines `NVIC_PRIO_BITS` based on the chip variant, but this is actually defined in the PAC (e.g. [here](https://docs.rs/atsamd51j/latest/atsamd51j/constant.NVIC_PRIO_BITS.html) or [here](https://atsamd-rs.github.io/docs/samd21j/thumbv6m-none-eabi/doc/atsamd21j/constant.NVIC_PRIO_BITS.html)), so does not need to be defined manually in the HAL. I'll note that the modified [`set_monotonic_prio`](https://github.com/kyp44/atsamd/blob/d05b3005414c1c64f3e6d5e323d1274423e8d70a/hal/src/rtc/rtic/mod.rs#L321C11-L321C29) would also need to depend on `NVIC_PRIO_BITS`.

hal/src/rtc/rtic/mod.rs
kyp44: The requested documentation was added with commit `71b308`.

kyp44:
@jbeaurivage  I just added comments to the particular items above as some require further discussion I think.

The latest commit (`71b308`) addresses/fixes some items as noted above, including your note about `#[hal_macro_helper]`.

Regarding testing on thumbv6 targets, I may be getting a Feather M0 in the next couple of weeks to carry out and test the clocks v2 API work. If someone is able to test the monotonic in the meantime that would be great! I will also note that I just had the monotonic on the PyGamer running for over a week straight with no issues.

As for breaking changes, this PR does technically have breaking changes, but it is something minor and silly. In particular, in the current `master` with `rtic` enabled, the types `rtc::Duration` and `rtc::Instant` are publicly defined, I suppose to support the the RTIC v1 `Monotonic` implementation. In the current PR branch, these are moved to `rtc::rtic::v1` as they are specific to RTIC v1.

hal/src/rtc/rtic/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+//! be done manually.
+//!
+//! # Example
+//!
+//! ```
+//! use atsamd_hal::prelude::*;
+//!
+//! // Create the monotonic struct named `Mono`
+//! rtc_monotonic!(Mono, rtc_clock::Clock32k);
+//!
```
jbeaurivage, line 84: Here I would suggest adding something like this, since users are much more likely to read the module docs than stumble across the docs for the `set_monotonic_prio` function:

```rs
// Uncomment if not using the RTIC RTOS:

// #[no_mangle]
// static RTIC_ASYNC_MAX_LOGICAL_PRIO: u8 = 1;

// This tells the monotonic driver the maximum interrupt
// priority it's allowed to use. RTIC sets it automatically,
// but you need to set it manually if you're not using RTIC.
```

hal/src/rtc/rtic/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+//! fn init() {
+//!     # // This is normally provided by the selected PAC
+//!     # let rtc = unsafe { core::mem::transmute(()) };
+//!     # let mut mclk = unsafe { core::mem::transmute(()) };
+//!     # let mut osc32kctrl = unsafe { core::mem::transmute(()) };
+//!     // Here the RTC clock source should be configured using the clocks API
+//!
```
jbeaurivage, line 91: I wonder if it would make sense to add the clock setup code here? Or otherwise add an example using the monotonics in a T1 BSP so we can refer users if they want to see a full working example.

Of course, if we add the clock setup here, we need to remember to update it once we switch over to clocks v2.

jbeaurivage (commented):
Ok, I've tested this on a SAMD21, and found it to be reasonably accurate. I found a few other minor discussion points. Once we've addressed all remaining items I'll definitely be ready to merge this.

hal/src/rtc/rtic/mod.rs
jbeaurivage: There are two things to consider here:

* If we choose to include `fugit::{ExtU64, RateExtU64}` in the prelude, this might be a breaking change. Suppose a user has this bit of code:

  ```rs
  async fn wait_100_ms(){
      use atsamd_hal::prelude::*;

      Some32BitDelay::delay(100.millis()).await;
  }
  ```

  If we introduce the u64 versions, now the compiler won't be able to disambiguate between `ExtU32::millis` and `ExtU64::millis`, meaning the user will have to change their code to `Some32BitDelay::delay(100_u32.millis())`. Which is fine, we just have to be aware of it.

* I have no problems with adding this to the prelude:

  ```rs
  #[cfg(feature = "rtic")]
  pub use rtic_time::Monotonic as _;
  ```

  That being said I would also put this in `lib.rs` for completness:

  ```rs
  #[cfg(feature = "rtic")]
  pub use rtic_time;
  ```

hal/src/rtc/modes.rs
jbeaurivage: Oh I see what you mean. I think in general `unsafe` should be reserved for memory unsafety, and not ensuring that other types of invariants are upheld (unless they're memory related, of course). The pacs mark all the `bits` register accessor functions as unsafe, because it can't know if the bits you're writing are a valid bit pattern for the specific register. In the case of COMP and PER, any bit pattern is valid, so there's no need special care.

Therefore in my opinion all methods should be marked safe, especially since they can't even be used outside the HAL.

hal/src/rtc/rtic/mod.rs
jbeaurivage: Well, if we're going to be in the business of managing interrupts, I think it should make sense to have an `interrupts` module (outside of the `async` feature gate) to keep all the functionality in one place. I guess technically one could use the `Monotonic` trait without async, if they limited themselves only to the `now()` method.

> but this is actually defined in the PAC

Sweet, I never noticed, let's use that instead then. 

I'll make a PR that moves the interrupt functionality not specific to async into its own module.

hal/src/rtc/modes.rs
kyp44: What I'll do is have every method be safe and then have a `# Safety` section for each that describes the conditions required before calling the method, which is usually just that the RTC has already been set to the mode by calling `set_mode`.

hal/src/rtc/rtic/mod.rs
kyp44: It looks like the `ExtU32` (and `RateExtU32`) are used elswhere in the HAL, in timers and whatnot. But I think the `u64` extensions are only used for the RTIC monotonic, so maybe we could just import these in the prelude as well only if the `rtic` feature is enabled to minimize the impact.

hal/src/rtc/rtic/mod.rs
kyp44: Sounds good, we'll wait for #807 to be merged, then keep the `rtic` and `async` features independent.

I'll note that you could definitely use `rtic` without `async` if you just want to use async for delays, but don't necessarily need async interfaces to other peripherals. This may actually be the case for my project, especially since the drivers for the devices I need do not have async interfaces.

hal/src/rtc/rtic/mod.rs
kyp44: Actually, with the v1 clocks API there is no way to set the RTC clock source for SAMx5x targets, but it does default to the internal 32k clock. This is because, on SAMD11/21, the RTC is just another generic clock (so it can be configured in v1) but it is a special thing on SAMx5x, which was never implemented in the clock v1 API. Note that it can be set in clocks v2. Not sure how we want to handle this in the example.

Once this gets merged I did intend to go back and update any RTIC examples in BSPs to use the new RTIC v2 monotonics. Does it make sense to go ahead and do that as part of this PR?

jbeaurivage: Yeah I think the easiest would be to just add full examples to BSPs. Let's do that instead of overcomplicating the docs. Feel free to add them in a new PR after this one is complete, I don't want to hold it up some more unnecessarily.

hal/src/rtc/rtic/mod.rs
jbeaurivage: Sounds like a good compromise.

hal/src/rtc/rtic/mod.rs
kyp44: Agreed, we'll leave the above documentation the same since it's pretty general (won't need changed once the eventual clocks v2 migration happens). Added a new PR to my TODO list pending this one getting merged.

hal/src/rtc/rtic/mod.rs
kyp44: Changed in the latest commit per the discussion.

hal/src/rtc/rtic/mod.rs
kyp44: Moves the documentation in the latest commit.

hal/src/rtc/modes.rs
kyp44: Implements the changes in the latest commit.

kyp44:
Just pushed a new commit (`d3f23c79`), which addresses additional items as noted above. Note that this rebases against some new commits in `master`, which seems to resolve the automated checks that were failing before. Commit hash references above have been updated.

It also occurred to me that, as far as I am tracking, breaking changes are confined to the `rtic` feature. Of course a breaking change is still a breaking change overall.

Also, I think the only is pending issue is waiting for PR #807 to merge for the interrupt stuff.

jbeaurivage:
Awesome, thanks for the effort @kyp44! Let's wait for #804 and then I'll get this merged.

hal/src/rtc/rtic/mod.rs
kyp44: Implements this in the latest commit (rebases against master again to incorporate #807).

jbeaurivage (approved):

PR #808, opened 2025-01-07
Allow USB to work with V2 clocking API

rnd-ash:
# Summary

Continuing my work to allow more peripherals to work with the V2 clocking API, this PR adds a new method (Maybe name should be changed to something more permanent), that allows the UsbBus to be constructed with the new V2 clocking API. 

I've left the original constructor untouched in order to maintain compatibility.

Now, in the BSPs, USB can be constructed like so:

```rust
pub fn usb(
    usb: pac::Usb,
    pclk_usb: Pclk<Usb, Gclk0Id>, // Can use any Gclk here
    mclk: &mut pac::Mclk,
    d_minus: impl Into<UsbDm>,
    d_plus: impl Into<UsbDp>,
) -> UsbBusAllocator<UsbBus> {
    UsbBusAllocator::new(UsbBus::new_with_v2_clock(pclk_usb, mclk, d_minus.into(), d_plus.into(), usb))
}
```

### DIFF
--- a/hal/src/delay.rs
+++ b/hal/src/delay.rs
@@ -8,6 +8,8 @@ use crate::clock::GenericClockController;
 use crate::ehal::delay::DelayNs;
 use crate::ehal_02;
 use crate::time::Hertz;
+
+#[hal_cfg("rtc-d5x")]
 use crate::typelevel::Increment;
 
 #[hal_cfg("rtc-d5x")]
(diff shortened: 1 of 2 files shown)


rnd-ash:
When I created this PR, I was learning the crate, and I didn't realise until recently that you can call pclk.into() to create a USB clock that the V1 API requires.

Therefore, I believe it's safe to close this PR as it is semi pointless

PR #814, opened 2025-01-23
New ADC V2 with Async support

rnd-ash:
# Summary

This PR adds a new ADC API, based around a similar system to channels used by EIC and DMA peripherals.

At the moment I'll keep this as a draft PR for feedback whilst I am still finalizing bits and also adding SAMD11/21 support

## Features
* A new settings builder system for configuring the ADC
* Channels system (Each Analog capable pin has its own channel)
* Both blocking and Async read support
* Buffer filling support (Free running continuous mode)
* CPU temperature reading
* Internal voltages reading

## Progress checklist
* [x] SAMx51 support
* [ ] SAMx11/21 support
* [x] Channels API
* [ ] Settings builder (Mostly completed, need to add settings for Vref calibration)
* [x] Async/Blocking single reads
* [x] Async/Blocking buffer reads
* [x] CPU temperature reading


jbeaurivage (commented):
I just finished a first pass at reviewing this. In general, I would say this is an excellent implementation. Given your willingness to learn best practices, my review is more detailed than usual, so feel free to ignore some of these comments if you feel they're too nitpicky.

That being said, I do have some other comments that I feel deserve some attention:

* SAMD11 support. I feel we're so close it makes sense to finish the implementation for SAMD11 chips as well. As far as I can tell, the ADC is exactly the same between D21 and D11 targets. The only thing that might change and that I'm unsure of, is the maximum supported clock speed.
* We've already discussed this a little, but I'm not so sure the `Channel` API is so necessary after all. When I initially suggested this, I was under the impression that the ADC could perform conversions simultaneously on multiple channels, similar to EIC or DAMC. That's obviously not the case. As far as I can tell, there is also a 1:1 relationship between ADC channels and physical pins. So while the ADC can mux between channels, there is no pin-channel muxing going on. I think the ADC could simply take `&mut` refs to already-configured pins instead. This would cut down on complexity a whole bunch.
* Before merging, the Tier 1 BSP examples will have to be updated, and clippy warnings resolved. Ideally we would show examples where the CPU clock is ran at full speed, and a second GCLK is started at a slower speed to clock the ADC.
* I like that the d51 implementation takes advantage of the v2 clock system. However, when looking at example code there seems to be a flaw in either the adc or the clocking system API: one needs to (unsafely) steal the peripherals to create a `MCLK` instance in order to configure the ADC. I'm not really comfortable releasing in this state: this seems like a big oversight for a module that claims to use type safety to avoid clocking mistakes at runtime...

hal/src/peripherals/adc/d5x/mod.rs
```diff
@@ -0,0 +1,290 @@
+pub mod pin;
+
+use pac::adc0::avgctrl::Samplenumselect;
+use pac::adc0::ctrlb::Resselselect;
+
+use super::{async_api, Adc, AdcAccumulation, Config, Error, Flags};
```
jbeaurivage, line 6: The `async_api` import should be feature-gated

jbeaurivage (changes requested):

hal/src/peripherals/adc/config.rs
```diff
@@ -0,0 +1,173 @@
...
+///
+/// ## For single sample
+/// ```
+/// SPS = (GCLK_ADC / clk_divider) / (sample_clock_cycles + bit_width)
+/// ```
+/// ## For multiple samples (Averaging or Summed)
+/// ```
+/// SPS = (GCLK_ADC / clk_divider) / (n * (sample_clock_cycles + bit_width))
+/// ```
+#[derive(Copy, Clone)]
```
jbeaurivage, line 66: I don't think it's appropriate for the `Config` type to be `Copy`.

jbeaurivage:
Here's a proof of concept of what I mean by using the typesafe clocking system. It's still untested and needs some more feature gating to accomodate D11 and D21 targets, but I think it gets the idea across: https://github.com/jbeaurivage/atsamd/tree/adc-v2-clocking

With an example usage:
```rs
let pins = Pins::new(cx.device.port);
let (buses, clocks, tokens) = clock_system_at_reset(
    cx.device.oscctrl,
    cx.device.osc32kctrl,
    cx.device.gclk,
    cx.device.mclk,
    &mut cx.device.nvmctrl,
);
let mut apb = buses.apb;
let adc0_clk = apb.enable(tokens.apbs.adc0);

let adc0_settings = Config::new()
    .clock_cycles_per_sample(5)
    .clock_divider(Prescalerselect::Div2)
    .sample_resolution(AdcResolution::_12bit)
    .accumulation_method(AdcAccumulation::Single);

let gclk0 = clocks.gclk0;
let (pclk_adc0, gclk0) = Pclk::enable(tokens.pclks.adc0, gclk0);
let (_pclk_adc1, _gclk0) = Pclk::enable(tokens.pclks.adc1, gclk0);

let (adc0, channels_adc0) =
    Adc::new(cx.device.adc0, adc0_settings, adc0_clk, pclk_adc0.into()).unwrap();
```

hal/src/peripherals/adc/d11/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+            .modify(|_, w| w.ressel().variant(config.bit_width));
+        self.sync();
+
+        self.adc
+            .sampctrl()
+            .modify(|_, w| unsafe { w.samplen().bits(config.sample_clock_cycles) }); // sample length
+        self.sync();
+        self.adc.inputctrl().modify(|_, w| {
+            w.muxneg().gnd();
+            w.gain().variant(Gainselect::Div2)
```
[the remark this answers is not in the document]
rnd-ash: Not sure, but when testing, it appears that gain of 1x actually seems to provide a 2x amplification of the output voltage reported, and 1/2 outputs 1x. Ill need to read the data sheet more as to why this might be the case, maybe something else needs configuring?

hal/src/peripherals/calibration/d5x.rs
```diff
@@ -59,3 +73,51 @@ pub fn adc1_biasref_scale_cal() -> u8 {
...
     cal(3, 0, 0b111) as u8
 }
+
+pub fn tl() -> f32 {
+    parts_to_f32(cal(0x80, 7, 0b11111111), cal(0x80 + 1, 3, 0b1111))
+}
+
+pub fn th() -> f32 {
+    parts_to_f32(cal(0x80 + 2, 3, 0b11111111), cal(0x80 + 2, 7, 0b1111))
+}
```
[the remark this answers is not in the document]
rnd-ash: or should we make them private in the crate? Since these functions are only to extract tl and th as per the datasheet for calibration of temperature sensor

hal/src/peripherals/adc/config.rs
rnd-ash: Copy just makes it a little easier if you want to use the same ADC Settings for both ADC0 and ADC1 (in the case of sam5x)

hal/src/peripherals/adc/d11/pin.rs
```diff
@@ -0,0 +1,75 @@
...
+    PB07: (Adc0, 15),
+
+    #[hal_cfg("pa08")]
+    PA08: (Adc0, 16),
+    #[hal_cfg("pa09")]
+    PA09: (Adc0, 17),
+    #[hal_cfg("pa10")]
+    PA10: (Adc0, 18),
+    #[hal_cfg("pa11")]
+    PA11: (Adc0, 19),
```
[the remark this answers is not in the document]
rnd-ash: Done. Check my recent commits, I added pins specific for SAMD11 as they are not mapped the same

hal/src/peripherals/calibration/d5x.rs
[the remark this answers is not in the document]
jbeaurivage: No, I think they can remain public - maybe they can be of use to someone outside this crate, and there's no invariant to uphold to use them correctly.

jbeaurivage:
You can check out [This PR](https://github.com/rnd-ash/atsamd/pull/5), which implements some of the larger changes mentioned [here](https://github.com/atsamd-rs/atsamd/pull/814#pullrequestreview-2584186921)

I also want to point out that when testing on a SAMD51 board (metro M4), I regularly get the ADC stalling on waiting for the RESRDY flag to get set. So far the only way I managed to get ADC readings was:
* Using the clocks at reset (48 MHz), and
* Compiling in release mode.

To me this sounds like there is still a clock synchronization issue or a race condition somewhere.

So, among the things left to do:

- [ ] Merge https://github.com/rnd-ash/atsamd/pull/5 if you agree with this approach
- [ ] Document `tl` and `th`
- [ ] Update the T1 examples
- [ ] Fix the RESRDY bug

hal/src/peripherals/adc/d11/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+                // before it returns the result
+                self.adc.avgctrl().modify(|_, w| {
+                    w.samplenum().variant(adc_sample_count);
+                    unsafe {
+                        // Table 45-3 SAME51 datasheet
+                        w.adjres()
+                            .bits(core::cmp::min(adc_sample_count as u8, 0x04))
+                    }
+                });
+            }
```
jbeaurivage, line 98: I think the ADJRES math needs to be reviewed here... The bits define the division coefficient in 2^n increments. For example, with a sample count of 4, the result would get divided by 16. I doubt that's actually the intention here?

hal/src/peripherals/adc/d5x/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+                // before it returns the result
+                self.adc.avgctrl().modify(|_, w| {
+                    w.samplenum().variant(adc_sample_count);
+                    unsafe {
+                        // Table 45-3 SAME51 datasheet
+                        w.adjres()
+                            .bits(core::cmp::min(adc_sample_count as u8, 0x04))
+                    }
+                });
+            }
```
jbeaurivage, line 128: Same thing here

hal/src/peripherals/adc/d11/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+        &p.adc
+    }
+
+    #[inline]
+    fn enable_pm(pm: &mut pac::Pm) {
+        pm.apbcmask().modify(|_, w| w.adc_().set_bit());
+    }
+
+    #[inline]
+    fn calibrate(_instance: &Self::Instance) {}
```
jbeaurivage, line 36: Is it normal that the calibration code is missing for D11/D21 targets?

hal/src/peripherals/adc/d11/mod.rs
jbeaurivage: Well, it appears I was looking at the wrong table after all. After testing on hardware it indeed does appear to work as intended.

jbeaurivage:
@rnd-ash, if you're happy with this, I am too. I think you've been doing extensive testing recently. Only two things to get CI green and get this merged:
- [ ] Fix a small clippy lint in `calibration/d11.rs` (line 44).
- [ ] Fix the pygamer examples

rnd-ash:
I'll be able to fix these next week or this weekend when I'm back in the UK and have access to all my test boards (SAMD21 I need to verify it doesn't read old results in the ADC during sequential reads)

hal/src/peripherals/adc/config.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// Configure the ADC to sample at 250_000 SPS (Assuming the clock source is
+    /// 48MHz) using the following settings:
+    /// * clock divider factor of 32
+    /// * 6 clock cycles per sample
+    /// * 12bit sampling
+    /// * Single accumulation (No averaging or summing)
+    ///
+    /// ## Additional reading settings by default
+    /// * Use VDDANA as reference voltage for a full 0.0-3.3V reading
+    pub fn new() -> Self {
```
ianrrees, line 84: With a little change, this design could make firmware using this API a bit easier to read, at present understanding code that uses this would require digging in to the HAL docs/source.

These could be supplied as arguments rather than magic values, or if the intent is to use a builder pattern perhaps the struct could use `Option<>` for all fields (so `new()` returns a struct with all fields `None`).

hal/src/peripherals/adc/config.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// returned value will be 16bits long no matter what the sample Bit
+    /// width was selected as
+    pub fn accumulation_method(mut self, method: Accumulation) -> Self {
+        self.accumulation = method;
+        self
+    }
+
+    /// This adjusts the number of ADC clock cycles taken to sample a single
+    /// sample. The higher this number, the longer it will take the ADC to
+    /// sample each sample.
```
ianrrees, line 139: What is the maximum value, and assuming it's not 8b what happens if a config is specified that's over it?

hal/src/peripherals/adc/config.rs
```diff
@@ -0,0 +1,171 @@
...
+    ///
+    /// Configure the ADC to sample at 250_000 SPS (Assuming the clock source is
+    /// 48MHz) using the following settings:
+    /// * clock divider factor of 32
+    /// * 6 clock cycles per sample
+    /// * 12bit sampling
+    /// * Single accumulation (No averaging or summing)
+    ///
+    /// ## Additional reading settings by default
+    /// * Use VDDANA as reference voltage for a full 0.0-3.3V reading
```
ianrrees, line 83: It looks like the implementation actually uses the 1V reference

ianrrees (changes requested):
With stable Rust, I note `cargo doc` flags several dead links in the documentation, so please do have a look at those and ensure the rendered docs are what you intend.

Thanks for all this work!  Looks pretty good.

hal/src/peripherals/adc/async_api.rs
```diff
@@ -0,0 +1,81 @@
...
+impl<A: AdcInstance> Handler<A::Interrupt> for InterruptHandler<A> {
+    unsafe fn on_interrupt() {
+        let mut peripherals = crate::pac::Peripherals::steal();
+        let adc = A::peripheral_reg_block(&mut peripherals);
+
+        let flags_pending = Flags::from_bits_truncate(adc.intflag().read().bits());
+        let enabled_flags = Flags::from_bits_truncate(adc.intenset().read().bits());
+
+        if enabled_flags.intersects(flags_pending) {
+            adc.intenclr().write(|w| w.bits(flags_pending.bits()));
```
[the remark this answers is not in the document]
rnd-ash: Yes, since the ADC can trigger at 1M/s which is really fast, so it's to prevent a race condition where the CPU is midway through processing the waker or poll method when the ADC fires another interrupt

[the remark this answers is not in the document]
jbeaurivage: @ianrrees, can confirm that this is indeed the correct way to handle interrupts for an async task. The interrupt flag is what is used to communicate to the task that it's done, but if you didn't disable the interrupt before exiting the handler, it would keep calling itself in a loop indefinitely, never giving the task a chance to clear the flag.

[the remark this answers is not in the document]
ianrrees: Ah, so the enable bit is used to indicate whether the task has been scheduled, and the flag is managed by the peripheral and the task.

My concern is that it needs to be possible to make a firmware that can guarantee either it has processed every item (ADC conversion result, in this case) or it will indicate that it has failed to do so.  To make this guarantee, we need to be sure that there's nowhere in our code that a pause in the CPU (to handle a higher priority interrupt or whatever) could result in missing an item without raising some sort of error.

From my read of the datasheet, the OVERRUN bit should be set when a new result becomes ready before the previous one was read out, regardless of whether the interrupt enable bit is set or not, so disabling the interrupt like this should be OK.

boards/feather_m4/examples/adc.rs
```diff
@@ -0,0 +1,65 @@
...
+};
+
+#[entry]
+fn main() -> ! {
+    let mut peripherals = Peripherals::take().unwrap();
+    let _core = CorePeripherals::take().unwrap();
+
+    let pins = Pins::new(peripherals.port);
+
+    // TODO: currently this example only runs at the chip's
```
ianrrees, line 30: Let's reference https://github.com/atsamd-rs/atsamd/issues/819 in the comment (presuming this is the referenced bug!)

hal/src/peripherals/adc/async_api.rs
```diff
@@ -0,0 +1,81 @@
...
+where
+    F: crate::async_hal::interrupts::Binding<I::Interrupt, InterruptHandler<I>>,
+{
+    #[inline]
+    pub(super) async fn wait_flags(&mut self, flags_to_wait: Flags) -> Result<(), Error> {
+        use core::task::Poll;
+
+        // We automatically check for errors
+        let flags_to_wait = flags_to_wait | Flags::OVERRUN;
+        self.disable_interrupts(Flags::all());
```
ianrrees, line 53: I'm curious why this disables all interrupts, as opposed to just `flags_to_wait` (which is what gets re-enabled)

hal/src/peripherals/adc/async_api.rs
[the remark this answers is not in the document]
rnd-ash: This is something we've spoken about. I think at a later date when I get around to doing the DAC peripheral, it'll be time for me to add DMA to both the ADC and DAC. This should mitigate the issues with multiple fast continuous reads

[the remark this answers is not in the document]
ianrrees: This isn't actually about the speed of ADC triggers - it's about ensuring there are no timing holes which could result in missing a result.  Yes, maybe there would be more apparent problems when running the ADC faster, but if there is a timing hole then problems can arise even with a relatively slow rate.

In any case, the datasheet indicates that OVERRUN is independent of the enable bit, so this doesn't appear to be a timing hole.

hal/src/peripherals/adc/async_api.rs
rnd-ash: its to prevent a race condition as far as @jbeaurivage explained to me. Disable everything before starting the polling (Which then enables the interrupts once it is ready)

hal/src/peripherals/adc/config.rs
rnd-ash: Intvcc1 - VDDANA reference voltage (Refselselect register)

hal/src/peripherals/adc/d11/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+        self.adc.ctrla().modify(|_, w| w.enable().clear_bit());
+        self.sync();
+    }
+
+    #[inline]
+    pub(super) fn start_conversion(&mut self) {
+        // The double trigger here is in case the VREF value changed between
+        // reads, this discards the conversion made just after the VREF changed,
+        // which the data sheet tells us to do in order to not get a faulty reading
+        // right after changing VREF value
```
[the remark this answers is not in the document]
rnd-ash: Ill do some testing tomorrow, since this was to prevent stagnant readings from the ADC when cycling between channels very quickly (Read 1 sample, cycle to another channel,...), but since then, I found the powering down the ADC and powering it up between samples helped, so Ill test without this double trigger and see if the problem is still gone

boards/feather_m0/examples/adc.rs
```diff
@@ -1,41 +1,62 @@
...
-    let pins = bsp::Pins::new(peripherals.port);
-    let mut delay = hal::delay::Delay::new(core.SYST, &mut clocks);
-    let mut adc = Adc::adc(peripherals.adc, &mut peripherals.pm, &mut clocks);
-    let mut a0: bsp::A0 = pins.a0.into();
+    let gclk0 = clocks.gclk0();
+    let adc_clock = clocks.adc(&gclk0).unwrap();
+
+    let adc_settings = Config::new()
+        .clock_cycles_per_sample(5)
+        // Overruns if clock divider < 128 in debug mode
```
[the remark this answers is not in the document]
rnd-ash: Agreed, DMA is my next thing to work on with the ADC generally speaking

hal/src/peripherals/adc/async_api.rs
ianrrees: I guess the risk is that we later want to do something with one of the other flags, `WINMON` or whatever.  In that case though, this code would probably need to be revisited either way.

I see that at least the SPI async implementation does the same thing so may as well be consistent.

jbeaurivage: Good point @ianrrees, we could simply disable the flags we want to wait on. My thinking (and the same goes for all async implementations in the HAL), is that disabling everything can prevent previously enabled (and unused) interrupt sources from uselessly waking the task.

boards/feather_m4/examples/adc.rs
jbeaurivage: I think we can now remove the comment and either leave the clocks in their reset state, or feed a clock up to 100 MHz to the ADC.

jbeaurivage: Same with the `metro_m4` ADC examples.

hal/src/peripherals/adc/d11/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub(super) fn clear_flags(&mut self, flags: Flags) {
+        unsafe {
+            self.adc.intflag().write(|w| w.bits(flags.bits()));
+        }
+    }
+
+    /// Check the interrupt flags, clears them and returns `Err` if an overflow
+    /// occured
+    #[inline]
+    pub(super) fn check_and_clear_flags(&mut self, flags: Flags) -> Result<(), Error> {
```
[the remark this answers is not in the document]
jbeaurivage: Check out https://github.com/rnd-ash/atsamd/pull/10

boards/feather_m0/examples/adc.rs
```diff
@@ -1,41 +1,62 @@
...
+    )
+    .unwrap();
+    let mut adc_pin = pins.a0.into_alternate();
 
     loop {
-        let data: u16 = adc.read(&mut a0).unwrap();
-        hprintln!("{}", data).ok();
-        delay.delay_ms(1000u16);
+        let mut buffer = [0; 16];
+        let res = adc.read_buffer_blocking(&mut adc_pin, &mut buffer).unwrap();
```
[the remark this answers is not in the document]
jbeaurivage: Fixed in https://github.com/rnd-ash/atsamd/pull/10

boards/feather_m0/examples/adc.rs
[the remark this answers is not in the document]
jbeaurivage: Fixed in https://github.com/rnd-ash/atsamd/pull/10

boards/feather_m4/examples/adc.rs
jbeaurivage: Fixed in https://github.com/rnd-ash/atsamd/pull/10

hal/src/peripherals/adc/config.rs
jbeaurivage: After thinking about it, I agree that since there isn't an obvious default that  > 80% of users would use without a valid reason to deviate, we should force users to always explicitly pick their ADC parameters, UNLESS we'd want to leave the parameters to the defaults at reset.

jbeaurivage:
@rnd-ash, @ianrrees still had a few comments that are worth being addressed. I've fixed a bunch in https://github.com/rnd-ash/atsamd/pull/10. The only outstanding item that might require more discussion is the handling of the sampling parameters, ie, if we should force users to explicitly specify everything or not.

rnd-ash:
@ianrrees i think we are ready to merge this, just 1 minor bug with the SAMD21's temperature sensor not reading accurately, but I can (For now at least), just disable the cpu temperature function for SAMD21 if it proves to be a blocker.

I have not tested this on a D11 target either, as I don't have one, would it be possible for you to verify functionality?

ianrrees:
Awesome, thanks!  Yep, my next few days are going to be a bit variable in terms of computer time, but will try to get to it soon.

ianrrees:
Alrighty, have put a little bit of time in to this and have a few thoughts in no particular order:

1. The work looks really good!
2. I've fiddled with the Git history of this work and made a branch that's as-if this work was all done using rebases instead of merges, then rebased that on to the current master (and made a couple slight improvements).  The new branch is at https://github.com/ianrrees/atsamd/tree/adcv2_rebase .  I'd suggest that we look at whether it makes sense to keep the work in >1 commit (I guess it does, in particular removing the problematic examples) and clean up the history if so.  Regardless, we should keep it updated as needed via rebasing on to master.  Happy for it to live in my GH, or @rnd-ash if you'd like to update your branch that would look like (presuming you've not added more commits in the mean time):
```
git checkout adc_v2
git branch adc_v2_backup
git remote add ianrrees git@github.com:ianrrees/atsamd.git
git fetch ianrrees
git reset ianrrees/adcv2_rebase
git push --force
```
3.  Unless it gets fixed in the meantime, I'd like to have a play with the SAMD21 temperature sensor issue before merging - should be able to put some time in to it in the next couple days.  Not really a show stopper, just while we're in the area it seems like a good thing to sort out.

rnd-ash:
Done @ianrrees , should be up to date with your branch

hal/src/peripherals/calibration/d5x.rs
```diff
@@ -6,15 +6,35 @@ use core::ptr;
...
 
         (value >> bit_shift) & bit_mask
     }
 }
 
+// Needed for temperature calibration values stored in NVM
+fn parts_to_f32(int: u32, dec: u32) -> f32 {
+    let mut dec = dec as f32;
+
+    if dec < 10.0 {
```
ianrrees, line 28: This if/else seems suspect to me, it interprets both the scale and the value from the `dec` parameter.  The datasheet is really poor in this area, but my interpretation is that dec is a BCD coded field, so the method should simply be (probably not a function):

(int * 10 + dec) as f32 / 10.0

ianrrees: I'm playing with the CircuitPython implementation by comparison, [here's how they do it](https://github.com/adafruit/circuitpython/blob/002881826a04669abaaa4f2d5220379f786cb159/ports/atmel-samd/common-hal/microcontroller/Processor.c#L94-L108) - looks to be BCD (dec is the value in 1/10 degree C).

ianrrees: Sorry for all the reply-to-self on this, it's taking me a while to get caught up on the docs.  The D21 datasheet talks a bit more about the format in 37.11.8.2.1 Temperature Log Row, it's BCD so I'll PR a small change

boards/feather_m0/examples/adc.rs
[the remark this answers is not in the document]
ianrrees: We should also remove the comment about the overrun if divider is too small.

hal/src/peripherals/calibration/d5x.rs
rnd-ash: This is now fixed, both D51 and D21 is using integer + float part/10 - According to microchip itself

hal/src/peripherals/adc/d11/mod.rs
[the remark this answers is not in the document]
rnd-ash: @jbeaurivage on SAMx21 datasheet says that `Intvcc1` is 1/2 VDDANA, thus, Div2 must be used in order to scale Intvcc1 to 3.3V (By halving the input)

hal/src/peripherals/adc/d11/mod.rs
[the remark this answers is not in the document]
rnd-ash: Double trigger now done via a check and discard. If a setting was modified before reading, the system reads and discards, and then reads again. If no setting was changed, then the read and discard method does nothing, so we only sample the ADC once, So if you keep changing settings, you'd get ~ 1/2 your requested sampling rate, but if settings stays the same, you'd get your expected sample rate

PR #817, opened 2025-01-30
Updates all Tier 1 BSP examples to use RTIC v2 and the RTC monotonic where applicable

kyp44:
# Summary

Some Tier 1 BSP RTIC examples used RTIC v1 and others v2. When delays are required, they all used the inefficient SysTick monotonic.

This updates all Tier 1 BSP RTIC examples to use RTIC v2 and the recently released RTC monotonic instead. It also ensures that all Tier 1 BSPs feature at least one RTIC example, adding a `blinky_rtic` example as needed.

See the commit message for details on what was done for each BSP.

Note that I was only able to test the example changes for the `metro_m0` and `pygamer` BSPs. If others are able/willing to test on other boards that would definitely be a good thing.

# Checklist


jbeaurivage:
Good stuff, thanks @kyp44. Is there any more work to be done in this PR or is it ready to merge?

kyp44:
It's ready to merge as far as I am concerned.

PR #823, opened 2025-03-09
feat: Add new EHAL InputPin traits to EIC

kaidokert:
# Summary
Adding Ehal 1.0 InputPin trait implementations. I'm not sure if this is the cleanest / expected way to implement them. Can also put them back in `pin` modules.

# Checklist

### DIFF
--- a/hal/src/peripherals/eic/d11/mod.rs
+++ b/hal/src/peripherals/eic/d11/mod.rs
@@ -1,4 +1,5 @@
 mod pin;
+mod pin2;
 
 #[cfg(feature = "async")]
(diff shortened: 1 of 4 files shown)


ianrrees:
Thanks @kaidokert !  I was exploring this and wound up with a slightly different solution that is a bit tidier, basically just moved the Error impl out of the async-conditional module and added the new InputPin impl.

https://github.com/atsamd-rs/atsamd/pull/826 appears to supersede this so I'm going to close it, but please feel free to reopen (or chat on Matrix) if I've missed anything!

PR #825, opened 2025-03-11
embassy-time-driver using RTC interrupts

stillinbeta:
# Summary
These changes implement the [embassy-time-driver](https://crates.io/crates/embassy-time-driver) interface, used by Embassy for scheduling and a prerequisite for the Embassy runtime. 

These changes are confined to the `embassy-time` feature. 

I have only tested these changes on the `pyportal` board (see examples provided). 

# Checklist

## If Adding a new Board

#### Note


hal/src/rtc/mod.rs
```diff
@@ -14,12 +14,15 @@ use core::marker::PhantomData;
 #[cfg(feature = "sdmmc")]
 use embedded_sdmmc::{TimeSource, Timestamp};
 
-#[cfg(feature = "rtic")]
-mod modes;
+#[cfg(any(feature = "rtic", feature = "embassy-time"))]
+pub mod modes;
```
stillinbeta, line 18: just to make debugging easier, we can seal it back up when things work

hal/src/rtc/embassy/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+#[interrupt]
+fn RTC() {
+    critical_section::with(|cs| {
+        let rtc = unsafe { Rtc::steal() };
+        if RtcMode0::check_interrupt_flag::<Compare0>(&rtc) {
+            // Due to synchronization delay, the RTC may be slightly behind
+            // Assume the current time is the time the interrupt is set for
+            let now = RtcMode0::get_compare(&rtc, 0) as u64;
```
stillinbeta, line 43: @sjroe solved this by setting the prescaler to four, thus ensuring the resolution was coarse enough to not cause issues. I like this better, but am open to suggestions

boards/pyportal/examples/blinky_interrupt.rs
```diff
@@ -0,0 +1,163 @@
+//! Turn on and off with an LED
```
stillinbeta, line 1: This could potentially be moved to another PR. I wrote this while debugging my interrupt code, and thought it might be a useful example for others in the future.

hal/src/rtc/modes.rs
```diff
@@ -71,6 +71,14 @@ macro_rules! create_rtc_interrupt {
     };
 }
 
+// will be unused in boards that don't use syncbusy
```
stillinbeta, line 74: I only added the sync_busy() to Mode0, since that's all I'm using. I'm not even sure it's required, but it is Technically Correct.

boards/pyportal/.cargo/config.toml
```diff
@@ -1,7 +1,7 @@
 # vim:ft=toml:
 [target.thumbv7em-none-eabihf]
-runner = "hf2 elf"
-# runner = 'probe-rs run --chip ATSAMD51J20A'
```
jbeaurivage, line 4: Could you please revert this?

boards/pyportal/Cargo.toml
```diff
@@ -19,7 +19,8 @@ version = "0.7.5"
 optional = true
 
 [dependencies.atsamd-hal]
-version = "0.21.0"
+# version = "0.21.0"
+path = "../../hal"
```
jbeaurivage, line 23: Same here, we'd like to keep pyportal a Tier 2 boarrd

jbeaurivage (changes requested):
Thanks for the PR @stillinbeta! I intend to review this shortly. In the meantime, could you revert these two changes? Thanks!

boards/pyportal/Cargo.toml
stillinbeta: Not sure exactly how to manage this, then. Should I have the PR to add `embassy-time` support be a stand-alone PR, then send another PR once there's a HAL release with the examples?

hal/src/rtc/embassy.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn set_alarm(&self, _cs: &CriticalSection, at: u64, rtc: &Rtc) -> bool {
+        // Embassy uses u64::MAX as a "no upcoming interrupt" sentinel
+        let at = match u32::try_from(at) {
+            Ok(at) => at,
+            _ if at == u64::MAX => u32::MAX,
+            Err(_) => return false,
+        };
+
+        RtcMode0::set_compare(rtc, 0, at);
+        true
```
ritikmishra, line 57: i am experiencing race condition in this implementation where it is possible for the RTC counter to advance past `at` while COMP0 is being set

the [[embassy timer queue implementation for stm32 microcontrollers]](https://github.com/embassy-rs/embassy/blob/6186d111a5c150946ee5b7e9e68d987a38c1a463/embassy-stm32/src/time_driver.rs#L449-L514) tries to prevent this from happening by avoiding setting alarms for timestamps in the very near future:

kyp44:
As I'm interested in this getting merged, I had a good look at at and tested it. Here are some comments/questions:

1. The implementation is well done, and I appreciate that it utilizes the RTC modes abstraction layer instead of accessing the registers directly.
2. I tested this both on a [Metro M0 (SAMD21G)](https://github.com/kyp44/atsamd-tests/tree/embassy-tests/metro/embassy-stress) and the [PyGamer (SAMD51J)](https://github.com/kyp44/atsamd-tests/tree/embassy-tests/pygamer/embassy-stress) and it works really well.
3. For safety reasons, it seems like the driver `init` function should require the [`pac::Rtc`](https://docs.rs/atsamd51j/latest/atsamd51j/struct.Rtc.html) as not doing so allows the user to simultaneously use the [`Rtc`](https://docs.rs/atsamd-hal/latest/atsamd_hal/rtc/struct.Rtc.html), which would be bad. Doing so would also allow `init` to not need to be `unsafe`. Is there particular a reason why this was not done? 
4. `init` requires proof of the RTC clock being configured (e.g. [`RtcClock`](https://docs.rs/atsamd-hal/latest/atsamd_hal/clock/struct.RtcClock.html) for `thumbv6m` chips), there are a couple of concerns with this:
  a. For `thumbv7` chips, this requires v2 of the clock API, which is not well support yet throughout the rest of the HAL, making it difficult to actually use for most projects. Perhaps an alternative `init` function could be added (and later removed after full migration) that does not require proof so as to be compatible with v1 on `thumbv7` chips. Edit: Issue #912 tracks the progress of the clock v2 API migration.
  b. This allows the user to configure the RTC clock to be whatever clock rate they like. However, embassy will always assume that the rate is 32.768 kHz due to passing the [`tick-hz-32_768` flag to `embassy-time-drive`](https://github.com/stillinbeta/atsamd/blob/3d7a9612be2acecb91055961757072f3991803c0/hal/Cargo.toml#L74). Unfortunately, unlike RTIC, it seems like the only way to specify the time driver tick rate is via these feature flags. We could just allow the user to set this flag themselves in their `embassy-time` dependency, but, annoyingly, this defaults to 1 MHz if nothing is specified, which will result in odd problems if the user is not aware they need to pass this themselves. An alternative would to for the driver to initialize the Rtc clock itself, always to the same rate, but this does not really seem possible to do for `thumbv6` chips in a way that avoids at best a runtime panic. I believe with v2 (`thumbv7` targets only right now) this could be done with compile-time assurances by simply passing in the RTC clock token.
5. As the embassy time counter is inherently a `u64` while the RTC only supports `u32`, it would be good to enhance this with [half-period counting (HPC)](https://docs.rs/rtic-time/2.0.0/rtic_time/half_period_counter/index.html) as the RTIC monotonic uses, otherwise tasks could stall around the 36-hour mark. This is an enhancement that could be added later that would not affect the API at all. The correct way to do this would probably be a single abstraction that incorporates HPC and can be used by both the embassy time driver and the RTIC monotonic. This is something I am certainly willing to take a crack at after this gets merged.

Note: This PR makes a few changes to the RTC modes abstraction. I also made some changes and additions with my RTC fixes in PR #845. In that PR I've already merged the changes from this one, so that version should take precedence in the event of any conflicts.

Edit: For my own purposes, I'm maintaining [this branch in my fork](https://github.com/kyp44/atsamd/tree/embassy-interrupt) that I am keeping up to date with [`atsamd-rs/atsamd:master`](https://github.com/atsamd-rs/atsamd/tree/master) (via rebasing) and have correctly merged in the changes from PR #845 discussed above. Since this is rebased, I don't know that a PR against [stillinbeta/atsamd:embassy-interrupt](https://github.com/stillinbeta/atsamd/tree/embassy-interrupt) would work correctly.

boards/pyportal/Cargo.toml
stillinbeta: Opened secondary PR #1008

stillinbeta:
@kyp44 thanks for doing the hard work of rebasing this back onto master! I've gone ahead and removed all the pyportal changes (they're in #1008 now). I'm going to see what I can do about the race condition @ritikmishra found.

PR #833, opened 2025-03-23
Fix CI docs and publish to atsamd-rs/atsamd GitHub Pages

ianrrees:
# Summary
The first commit gets the docs builds going again.  The second is a bigger change which I'm less confident about; in the old scheme docs were published to a separate repo atsamd-rs/docs, but that required some auth shenanigans which this removes by just using a gh-pages branch in this repo.  We've got some very old docs in the gh-pages branch already, I'm not sure on the history there but there is some precedent.

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -17,7 +17,6 @@ jobs:
       - name: Install Rust
         run: |
           rustup set profile minimal
-          rustup override set ${{ matrix.toolchain }}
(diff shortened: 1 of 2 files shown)


ianrrees:
Cheers @aralhin, I snagged your script from https://github.com/arahlin/qpoint

ianrrees:
Thanks @jbeaurivage !  I triggered a build of the docs, seems to have worked

PR #834, opened 2025-04-01
Update wio terminal examples

ianrrees:
This is mainly aimed at resolving a CVE in heapless flagged by dependabot.  Will require a new release of seeed-erpc which should be shortly after https://github.com/twitchyliquid64/seeed-erpc-rs/pull/5 is merged - thanks @twitchyliquid64 !

### DIFF
--- a/boards/wio_terminal/examples/clock.rs
+++ b/boards/wio_terminal/examples/clock.rs
@@ -1,5 +1,6 @@
 #![no_std]
 #![no_main]
+#![allow(static_mut_refs)]
 
 /// Makes the wio_terminal appear as a USB serial port, and display
 /// the time on the screen.
--- a/boards/wio_terminal/examples/clock.rs
+++ b/boards/wio_terminal/examples/clock.rs
@@ -27,7 +27,6 @@ use wio::pac::{interrupt, CorePeripherals, Peripherals};
 use wio::prelude::*;
 
 use core::fmt::Write;
-use heapless::consts::U16;
 use heapless::String;
 use wio::hal::rtc;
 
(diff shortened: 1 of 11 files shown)


ianrrees:
That seeed-erpc release is live now, so I think this should be ready to go.  I've not tested it on real hardware, as I don't own a wio terminal.

Cooler1989 (commented):
LGTM, later I can give it a spin on my Wio Terminal board.

ianrrees:
Thanks @Cooler1989 !  I think it may have been a while since those examples were exercised, so I won't be surprised if there are some bugs.

ianrrees:
Just wanted to mention here that @Cooler1989 is looking at updating the HAL version used by the wio terminal BSP, discussed in #836.

Since the upgrade in seeed-erpc done for this work results in our CI builds failing (my bad), I'd suggest we go ahead and merge this PR.  If any of the wio terminal examples have issues, those could be addressed in Cooler1989's work

boards/wio_terminal/examples/clock.rs
```diff
@@ -1,5 +1,6 @@
 #![no_std]
 #![no_main]
+#![allow(static_mut_refs)]
```
jbeaurivage, line 3: Is this something we'd want to look into replacing with `UnsafeCell`? Recent consensus is that static mut refs are considered Very Bad™ and essentially instant UB. Otherwise I'm happy with the PR, thanks!

jbeaurivage: Or, maybe @Cooler1989 could take this on in their own PR.

ianrrees: I've not dug in to the static mut thing, but FWIW have been using that in all my Rust firmware and not seen issues (of course, being careful to not violate the usual rules etc).  Maybe `UnsafeCell` is a good replacement in our examples, but in any case I feel we should handle that separately from the wio terminal update

jbeaurivage: Fair enough! I think it's quite telling that it's now denied by default, but it can certainly be handled separately.

PR #837, opened 2025-04-10
EIC Fixes for D5X

rnd-ash:
# Summary

As part of my work for implementing the Event system for pulse counting, I noticed that there was some issues with the EIC implementation. This PR fixes them.

1. Document the fact that when using the V2 clocking API, the user must enable the OSC32K clock and wire it up to the EIC Clock, since the HAL assumes that the OSC32K clock will drive the EIC peripheral. Later, I will possibly create a new constructor for the EIC peripheral that can take in a much faster clock source if the user wishes for it.

2. Create a constructor for the EIC peripheral that does not assume that OSC32K clock source is used. This allows the EIC peripheral to be driven by a much faster clock source, thus, have a faster event detection frequency

3. The enable_event method on an EIC channel does not actually function, as the EIC peripheral is enabled when the function is called, but the evctrl register is locked due to the peripheral being enabled. So the method now disables the EIC before enabling the bit, and then re-enables the peripheral. This is noted in the function documentation

### DIFF
--- a/hal/src/peripherals/eic.rs
+++ b/hal/src/peripherals/eic.rs
@@ -218,6 +218,17 @@ impl Eic {
 
     /// Create and initialize a new [`Eic`], and wire it up to the
     /// ultra-low-power 32kHz clock source.
+    ///
+    /// Due to the way the EIC peripheral is implemented, the EIC Clock MUST
+    /// be driven by the OSC32K clock, this can be done with the V2 clocking API as follows:
+    ///
+    /// ```no_run
+    /// let (osc32k, base) = OscUlp32k::enable(tokens.osculp32k.osculp32k, clocks.osculp32k_base);
+    /// let (gclk2, _osc32k) = Gclk::from_source(tokens.gclks.gclk2, osc32k);
+    /// let gclk2_32k = gclk2.enable();
+    /// let (pclk_eic, gclk2_32k) = Pclk::enable(tokens.pclks.eic, gclk2_32k);
+    /// let eic = Eic::new(&mut mclk, pclk_eic.into(), cx.device.eic).split();
+    /// ```
     #[hal_cfg("eic-d5x")]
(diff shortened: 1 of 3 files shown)


rnd-ash:
@jbeaurivage @ianrrees I had some conversations today about how this API works, and I believe my initial interpretation is wrong.

In summary, the EIC actually works like this.

1. You can give it a GCLK or OSC32K clock source. GCLK is preferred for when the processor is active, however you would want to use the OSC32K clock source when you put the processor to sleep so that it can wake up.

2. Connecting to OSC32K is not done via GCLK, EIC directly hooks into the OSC32K clock source when `ctrla.cksel` bit is set to 1.

So, forward question to ask. How to implement this in the API? I was thinking of the EIC clock source being an enum
```rust
pub enum EicClkSrc<'a>{
   Gclk(&'a EicClock),
   Ulp(&'a UlpClock)
}
```

Then provide this enum to the new() function of EIC so the user can initially set the clocking source. But then also provide `switch_to_gclk` and ``switch_to_ulp` methods to switch the clock provider for EIC.

Trying to implement this now, but I find that the V1 clock API does not have a struct to represent the OSC32K clock, any suggestions on how to go from here?

jbeaurivage:
@rnd-ash, your suggestion does make sense. IMO the premise behind the v1 clock API is that peripherals are pretty much responsible for setting up the clocking according to their needs. Therefore my suggestion would be this:

* If you provide a constructor for clock v2, obviously you would use the type system to prove that the OSCULP32K output (or a GCLK) is enabled. Potentially two different constructors according the clock source.
* For v1, enable the OSCULP32K inside the constructor if the user selects this clock source. In the `switch_to_osculp32k`, turn it on if it isn't already.

jbeaurivage:
Actually I would expose 3 different constructors:

* One for v1, where the clock input can be selected using an enum
* One for v2 with OSCULP32K
* One for v2 with GCLK

PR #838, opened 2025-04-11
Fix evsys user[m] register locations

rnd-ash:
# Summary

This patch fixes the length of the SAM5x evsys user[m] channel registers. The registers are 1 byte long, whilst the SVD had them as 4 bytes long, causing numerous issues when dealing with the EVSYS multiplexer.


rnd-ash:
Closing the PR. 4 bytes wide is correct, and the datasheet is wrong.

ianrrees:
@rnd-ash have you reported the datasheet error to Microchip?

PR #840, opened 2025-04-12
fix(tc): wait for `syncbusy` bits to clear instead of `perbufv`

ritikmishra:
# Summary

The [ATSAM D5x/E5x datasheet](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ProductDocuments/DataSheets/SAM-D5x-E5x-Family-Data-Sheet-DS60001507.pdf) mentions that the PERBUFV bit is only available in 8-bit mode (and will always read back 0 in 16/32-bit modes). 

The prescribed way to check that the timer has finished enabling/resetting is through the SYNCBUSY register

<table>
<tr>
<td> </td>
<td>
</td>
</tr>
</table>

# Checklist

#### Note

### DIFF
--- a/hal/src/peripherals/timer/d5x.rs
+++ b/hal/src/peripherals/timer/d5x.rs
@@ -96,16 +96,12 @@ where
 
         // Disable the timer while we reconfigure it
         count.ctrla().modify(|_, w| w.enable().clear_bit());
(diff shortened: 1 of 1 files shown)


ianrrees:
Welcome, and thanks @ritikmishra !

Issue #819, opened 2025-02-07
`clock::v2`: DPLL bug, potential UB?


jbeaurivage:
I've been trying out the `clock::v2` module, and I've encountered some pretty strange behavior when trying to setup the DPLL to speed up the main CPU clock.

## My test setup

* Metro M4 (SAMD51)
* The following example code:
```rs
#![no_std]
#![no_main]

use defmt_rtt as _;
use panic_probe as _;

use atsamd_hal::{
    clock::v2::{
        clock_system_at_reset,
        dpll::Dpll,
        gclk::{Gclk, GclkDiv16},
        pclk::Pclk,
    },
    pac::Peripherals,
};

#[cortex_m_rt::entry]
fn main() -> ! {
    let mut peripherals = Peripherals::take().unwrap();

    let (_buses, clocks, tokens) = clock_system_at_reset(
        peripherals.oscctrl,
        peripherals.osc32kctrl,
        peripherals.gclk,
        peripherals.mclk,
        &mut peripherals.nvmctrl,
    );

    // Set GCLK1 to use the 48 MHz DFLL...
    let (gclk1, dfll) = Gclk::from_source(tokens.gclks.gclk1, clocks.dfll);
    // ...and divide it down to 2 MHz
    let gclk1 = gclk1.div(GclkDiv16::Div(24)).enable();
    // Feed GCLK1 to the DPLL
    let (pclk_dpll0, _gclk1) = Pclk::enable(tokens.pclks.dpll0, gclk1);
    // Now start the DPLL with a 50x multiplier, to get a 100 MHz clock,
    // which is the maximum speed the ADC may run at.
    let dpll0 = Dpll::from_pclk(tokens.dpll0, pclk_dpll0)
        .loop_div(50, 0)
        .enable();

    defmt::info!(
        "DPLL: Ready: {}, Locked: {}",
        dpll0.is_ready(),
        dpll0.is_locked()
    );

    // When the DPLL is ready, swap the GCLK0 source to run the CPU
    // on the 100 MHz DPLL
    while !(dpll0.is_ready() && dpll0.is_locked()) {}
    let (_gclk0, _dfll, dpll0) = clocks.gclk0.swap_sources(dfll, dpll0);

    defmt::info!(
        "DPLL: Ready: {}, Locked: {}",
        dpll0.is_ready(),
        dpll0.is_locked()
    );

    loop {
        // A long `nop` delay as a rough way of comparing CPU clock speeds
[...]

PR #841, opened 2025-04-14
Various issue 819 fixes

ianrrees:
# Summary
While investigating #819 I made a few documentation changes, added ability to tune the DPLL filters, and added a method to check the DFLLRDY status bit.  Although the actual issue in 819 turned out to not be in our code, these might be useful additions.

# Checklist

### DIFF
--- a/hal/src/peripherals/clock/d5x/v2/dfll.rs
+++ b/hal/src/peripherals/clock/d5x/v2/dfll.rs
@@ -1,21 +1,27 @@
 //! # Digital Frequency Locked Loop
 //!
 //! The `dfll` module provides access to the 48 MHz digital frequency locked
-//! loop (DFLL) within the `OSCCTRL` peripheral.
+//! loop (DFLL or DFLL48M) within the `OSCCTRL` peripheral.
 //!
 //! ## Operation modes
 //!
-//! The DFLL can operate in both open-loop and closed-loop modes. In open-loop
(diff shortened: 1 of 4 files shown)


jbeaurivage:
Thank you for this. Did we reach a conclusion wrt the chicken-egg problem of having the DFLL defaulting to being on-demand, and therefore won't be ready unless there's a GCLK requesting it before it's turned on? Is it just a matter of documenting that behavior, or there still more work to be done?

I think an alternative that has been suggested is to force the user to explicitly choose whether they want on_demand to be on or off. That would be a good place to document the implications of leaving it on.

ianrrees:
I think that the `on_demand` thing isn't too big of a footgun; it only shows up if a check is added that the PLL is running, which in most cases probably isn't necessary.  We could put it in an example code along with a comment about why it is where it is.

Which reminds me, I think I noticed some example code in clock v2 that wanted an update...  Will aim to look at that soon so let's not merge this quite yet.

jbeaurivage:
The "code that wanted an update" was most likely in the ADC PR (#814) if that's of any help!

jbeaurivage:
@ianrrees were there any discussion points still open about this PR? Any more work to be done?

ianrrees:
I've rebased to the current master, and updated that example code to use `Fugit` - think this should be good to go, thanks!

Issue #501, opened 2021-10-04
RTC broken on SAMD51G


ryankurte:
hey there,

i've been messing around with trying to get the RTC (and RTIC) working on a `SAMD51G19A` with (almost) no success. attempting to use `atsamd-hal@master` and `cortex-m-rtic@v0.6.0-alpha.5`. possibly related to: #488.

i started with the setup from [feather_m4/examples/sleeping_timer_rtc.rs](https://github.com/atsamd-rs/atsamd/blob/master/boards/feather_m4/examples/sleeping_timer_rtc.rs#L40) which doesn't seem to actually kick off the RTC. looking at the datasheet the `count32_mode` function seems to be missing some steps, and there's a significant delta between [`Rtc::count32_mode`](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/rtc.rs#L224) and [`Rtc::into_count32_mode`](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/rtc.rs#L160), in fact if one adds this line the RTC does start to function:

```rust
let rtc = Rtc::count32_mode(peripherals.RTC, 1024.hz(), &mut peripherals.MCLK);
let mut rtc = rtc.into_count32_mode();    // Adding this line starts the RTC ticking
rtc.enable_interrupt();
```

so it's possible that was just missed in a previous refactor.

from there i've been attempting to use the RTC as a [`Monotonic`](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/rtc.rs#L468) source for RTIC per [feather_m0/examples/blinky_rtic.rs](https://github.com/atsamd-rs/atsamd/blob/master/boards/feather_m0/examples/blinky_rtic.rs#L38). one can observe that the RTC is running via increasing values from `monotonics::now()`, scheduling immediately (via `::spawn()`) works so the _dispatcher_ would seem to be okay, but `::spawn_after()` does not. attaching a debugger i can see the first call, but can't seem to step into the scheduling to see what it _thinks_ it's doing.

![Screenshot from 2021-10-04 21-45-51](https://user-images.githubusercontent.com/860620/135821634-fa5f551e-4346-4c36-9732-ae34fd4eca9f.png)

it's likely i've missed or misunderstood something but, would definitely appreciate any advice y'all have ^_^

ryankurte:
swapped to `SysTick` based monotonic and things get scheduled, so, it appears it's something to do with the RTC and/or embedded-time.

sajattack:
[...]

PR #845, opened 2025-04-21
Fixes problems with the `rtc::Rtc` abstraction and refactors to use the new `rtc::modes` abstractions.

kyp44:
# Summary

It was discovered that the [`rtc::Rtc`](https://docs.rs/atsamd-hal/latest/atsamd_hal/rtc/struct.Rtc.html) abstraction does not work properly, at least on SAMx5x targets. This was previously reported and discussed in issue #501, but it looks like a cause was never really nailed down.

This PR effectively re-writes the RTC with the same API, expanding and leveraging the new private `rtc::modes` abstractions that were written for the [RTC-based RTIC monotonic](https://docs.rs/atsamd-hal/latest/atsamd_hal/rtc/rtic/index.html).

Refer the to commit message for some additional details.

I wanted to get this PR going, and the code is ostensibly complete. However, I have only tested all implemented RTC features/modes on the PyGamer (SAMD51) so far, see [here](https://github.com/kyp44/atsamd-tests/tree/master/pygamer/hal-rtc). I have a Metro M0 (SAMD21) on order with a display, so I will test on that as well and report back when it arrives.

# Checklist

### DIFF
--- a/hal/src/rtc/mod.rs
+++ b/hal/src/rtc/mod.rs
@@ -26,17 +26,11 @@ pub mod rtic;
 
 // SAMx5x imports
 #[hal_cfg("rtc-d5x")]
-use crate::pac::{
-    Mclk as Pm, rtc::mode0::Ctrla as Mode0CtrlA, rtc::mode0::ctrla::Prescalerselect,
-    rtc::mode2::Ctrla as Mode2CtrlA,
-};
+use crate::pac::{Mclk as Pm, rtc::mode0::ctrla::Prescalerselect};
 
 // SAMD11/SAMD21 imports
 #[hal_cfg(any("rtc-d11", "rtc-d21"))]
-use crate::pac::{
-    Pm, rtc::mode0::Ctrl as Mode0CtrlA, rtc::mode0::ctrl::Prescalerselect,
-    rtc::mode2::Ctrl as Mode2CtrlA,
-};
+use crate::pac::{Pm, rtc::mode0::ctrl::Prescalerselect};
 
 pub use modes::mode2::Datetime;
 


jbeaurivage:
@kyp44, I tried testing this with an SAMD21 board. It seems like there is something making the RTC code stall forever. Here is my test code:

```rs
#![no_std]
#![no_main]

use atsamd_hal::ehal::delay::DelayNs;
use atsamd_hal::rtc::Rtc;
use atsamd_hal::time::KiloHertz;
use defmt_rtt as _;
use panic_probe as _;

use bsp::hal;
use bsp::pac;
use feather_m0 as bsp;

use bsp::{entry, pin_alias};
use hal::delay::Delay;
use hal::prelude::*;
use pac::{CorePeripherals, Peripherals};

use hal::fugit::RateExtU32;

#[entry]
fn main() -> ! {
        peripherals.gclk,
        &mut peripherals.pm,
        &mut peripherals.sysctrl,
        &mut peripherals.nvmctrl,
    );
    let pins = bsp::Pins::new(peripherals.port);
    let mut red_led: bsp::RedLed = pin_alias!(pins.red_led).into();

    // STALLS HERE!
    let mut rtc = Rtc::count32_mode(peripherals.rtc, 10_000.kHz(), &mut peripherals.pm);
    //

    loop {
        rtc.delay_ms(200);
        red_led.set_high().unwrap();
        rtc.delay_ms(200);
        red_led.set_low().unwrap();
    }
}
```

Digging a little further, looks like the stall occurs when trying to read the SWRST bit of the RTC mode 0 ([rtc/modes.rs:173](https://github.com/atsamd-rs/atsamd/blob/40f119ade82016759f6f73c595bc8afecf6400ed/hal/src/rtc/modes.rs#L173)):

```rs
// rtc/modes.rs
// Wait for the reset to complete
// SYNC: Write (we just read though)
// STALLS HERE!
#[hal_cfg(any("rtc-d11", "rtc-d21"))]
while rtc.mode0().ctrl().read().swrst().bit_is_set() {}
//
```

kyp44:
Thanks for the detailed info! I've got my new kit in hand, and have been working porting my PyGamer code to test the RTC to the Metro M0. I don't have a debugger, but I bought a display with it, so I'll definitely be able to fully troubleshoot this very soon.

kyp44:
Ok, so I finally got this [tested](https://github.com/kyp44/atsamd-tests/tree/master/metro/hal-rtc) and everything is working correctly on the Metro M0. It took much longer than expected because I went on a bit of a wild goose chase with the display driver for the display I purchased until I realized that async just isn't really viable on the SAMD11/21 chips due to memory limitations.

I also immediately ran into the issue you described above, and the problem was that both of us had not setup the RTC clock so that the peripheral was basically not operating, hence it could not reset correctly. The following example code is needed to first setup a clock for the RTC on the SAMx11/21:

```rust
let rtc_clock_src = self
    .clocks
    .configure_gclk_divider_and_source(
        ClockGenId::Gclk3,
        // This divider results in a 1.024 kHz RTC clock.
        32,
        ClockSource::Xosc32k,
        false,
    )
    .unwrap();

self.clocks.configure_standby(ClockGenId::Gclk3, true);
// The value returned here is proof that the RTC clock has been configured.
let proof = self.clocks.rtc(&rtc_clock_src).unwrap();
```

The RTC clock is setup differently on the SAMx11/21 and SAMx51 chips: with the former it has to be setup as a normal generator clock using the generic controller, wheres with the latter it's kind of a special animal. With the SAMx51, proof of the RTC clock being configured is only available with the clock v2 API, but with the existing clock API for the SAMx11/21, [proof is avilable](https://docs.rs/atsamd-hal/latest/atsamd_hal/clock/struct.RtcClock.html), which is returned from [`clock::GenericClockController::rtc`](https://docs.rs/atsamd-hal/latest/atsamd_hal/clock/struct.GenericClockController.html#method.rtc). However, setting up the [RTC abstraction](https://docs.rs/atsamd-hal/latest/atsamd_hal/rtc/struct.Rtc.html) does not currently require this proof, potentially causing the situation we both ran into. We may want to consider requiring this when building for SAMx11/21 chips, though at some point I do still intend to port the clocks v2 API to SAMx11/21 and to make their use viable across the HAL.

In terms of this PR, though, all the features of the the fixed RTC abstraction are working correctly on both chip variants.

EDIT: Found an issue reporting the above issue with not requiring the RTC clock setup: #642.

hal/src/rtc/modes.rs
```diff
@@ -149,9 +172,31 @@ pub trait RtcMode {
...
     }
 
+    /// Sets the clock prescalar divider to lower the tick rate.
+    ///
+    /// # Safety
+    ///
+    /// Should be called only when the RTC is disabled.
+    #[inline]
+    #[hal_macro_helper]
+    fn set_prescalar(rtc: &Rtc, divider: Prescalerselect) {
```
jbeaurivage: I think this should be spelled "prescaler".

jbeaurivage (commented):
This looks good @kyp44, thank you! Apart from some minor spelling mistake, I don't really have anything to add - did you confirm that Mode 2 works correctly on both platforms? I'm asking because it seems like that part is new functionality rather than just a refactor.

hal/src/rtc/modes.rs
kyp44: Great catch on the mispelling, I just pushed another commit, correcting this everywhere.

Previously, it looks like [`ClockMode`](https://docs.rs/atsamd-hal/latest/atsamd_hal/rtc/enum.ClockMode.html) _was_ implmented for [`Rtc`](https://docs.rs/atsamd-hal/latest/atsamd_hal/rtc/struct.Rtc.html), but I do not know whether it actuallty worked or not, which is moot at this point. This PR does add a new [`rtc::modes::RtcMode2`](https://github.com/kyp44/atsamd/blob/4ce11ccc6977803c23de9abf48cb3267b7ba073d/hal/src/rtc/modes.rs#L508-L645) abstraction. I just discovered a couple of minor mistakes there that I corrected as well. I have retested clock mode on both platforms for this new commit, using [this code](https://github.com/kyp44/atsamd-tests/blob/f0d3dbbea5f46b5fa93bf2b56e3f446e1cdfb133/lib/shared/src/tests/rtc.rs#L158-L163).

PR #847, opened 2025-04-22
ci: Check for doc warnings

jbeaurivage:
# Summary
Add a CI check to ensure the docs pass without warning. Currently, two feature sets are checked:

* `samd21g,dma,async,usb,defmt,rtic`, and
* `samd51j,dma,async,usb,defmt,rtic,can`

Ideally this list should be expanded to encompass all supported chips, but this is at least a good start.

### DIFF
--- a/.github/workflows/check-docs.yml
+++ b/.github/workflows/check-docs.yml
@@ -1,0 +1,24 @@
+name: Check docs
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  CARGO_TERM_COLOR: always
+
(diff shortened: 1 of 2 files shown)


ianrrees:
Many of the warnings fixed in that other PR were about broken links, which were due to eg documentation for SERCOM linking to async api that wasn't enabled.  I'm not sure of the best approach is to detect that situation, because we obviously don't want to do every combination of features.

We should cover the features used by documentation that gets published on GH Pages though, and maybe also update the published docs when the master branch is updated.  I'm not sure why it's manual at the moment, but have just triggered the action to see if we're in good shape :).

PR #850, opened 2025-04-29
Fix bugs with EIC and allow clock provider switching (D5X)

rnd-ash:
# Summary

 A reattempt at #837 
 
 * BREAKING - D5X EIC - Take a reference to EIC clock in new function rather than consuming it (Examples have been updated accordingly)
 * EIC D5X - Add methods for switching the clock provider of EIC between OSC32K and GCLK
 * EIC D2x - Disable EIC around channel manipulation operations
 * EIC - Fix both implementations of `enable_event` to disable the EIC  in order to successfully perform the operation


jbeaurivage:
@rnd-ash, were you happy with this PR? I would like to merge it.

rnd-ash:
Should be OK in its current state, there is other things I want to add, but I think we can add them later.

PR #856, opened 2025-05-03
feat: add ATSAME51G19A.svd

nyurik:
# Summary
I downloaded the `1.1.139 (2019-11-05)` pack for [ATSAME51G19A](https://www.microchip.com/en-us/product/ATSAME51G19A). I also copy/pasted `atsame51g18a.xsl` as `atsame51g19a.xsl` without any changes.

**ATTENTION**: I ran the `update-pacs.sh` and for some reason it changed the dsu/did `RESET_VALUE` - I am no embed expert, and I tbh have no idea what this means and how that affects other PACs.  It seems wrong, but I don't know enough about this - any suggestions?

# Checklist

## If Adding a new Board

#### Note


jbeaurivage:
Hi @nyurik, thanks for your contribution. However there is no need to add an SVD for the ATSAME51G19A. It's exactly the same chip as the ATSAME51G18A, just with a little more flash/RAM. You can use the same SVD and PAC for both chips - the only thing that you would need to change is the linker script to make use of the extra memory.

nyurik:
@jbeaurivage thank you for the quick reply!  Yeah, I just realized it after diving a bit deeper into the embedded world :)  One thing though - each chip seems to have a different Did reset value - which i guess is how chip can identify itself to the software. Is there any need/plans to remove the reset value from the generated code?  This way there is no chance of accidentally overriding it (is it even possible?), nor any other confusion due to the mismatch of generated code and the hardware

jbeaurivage:
No, there's no need to remove the device ID reset value. It's hardcoded on the chip, and anyone wanting to find out the device ID should read the register itself, and not rely on the reset value provided by the PAC since its going to be different for each die anyway.

PR #859, opened 2025-05-05
fix(i2c): acknowledge last byte when continuing read

### LINKED ISSUE #858: I2C transaction continued read broken
The current implementation of multiple, consecutive I2C reads in one transaction skips the first byte of every continued read.

At least on my hardware here (SAMD51P), the first byte gets filled with the last byte of the previous read. For example, if an ascending sequence of bytes is read into multiple buffers, the following will actually be written into the buffers:

```text
1st read op: [00, 01, 02, 03, 04]
2nd read op: [04, 05, 06, 07]
3rd read op: [07, 08, 09, 10, 11, 12, 13]
4th read op: [13, 14, 15]
```

This bug seems to have been introduced with the new ACK handling in [`fill_buffer`](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/sercom/i2c/reg.rs#L431). After having read the complete buffer, a NACK is scheduled [here](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/sercom/i2c/reg.rs#L449) to be sent later. However, if the next action is another read, [`cmd_read`](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/sercom/i2c/reg.rs#L380) does not get called in between, thus causing a byte to be read twice.

In a short test, I injected a call to `cmd_read` into the [`continue_read`](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/sercom/i2c/reg.rs#L480) method right before the call to `fill_buffer`. This seems to work for me.

fko-kuptec:
Should fix #858

### DIFF
--- a/hal/src/sercom/i2c/reg.rs
+++ b/hal/src/sercom/i2c/reg.rs
@@ -476,8 +476,16 @@ impl<S: Sercom> Registers<S> {
     /// Continue a read operation that was issued before with
     /// [`do_read`](Self::do_read) or [`continue_read`](Self::continue_read)
     /// without a repeated start condition in between
+    ///
+    /// Note, that the previous read must have been non-empty, otherwise a byte
+    /// will be skipped!
     #[inline]
     pub(super) fn continue_read(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
+        // The last call to `fill_buffer` did not ACK the last received byte to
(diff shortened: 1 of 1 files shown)


fko-kuptec:
Maybe there is a nice solution, but this fixes the issue at least for my testing on SAMD51P

PR #860, opened 2025-05-06
docs: a few minor doc improvements

nyurik:
# Summary
Just a few cleanups - trying to make things a bit more readable

### DIFF
--- a/atsamd-hal-macros/src/lib.rs
+++ b/atsamd-hal-macros/src/lib.rs
@@ -4,7 +4,7 @@
 //!
 //! The main purpose of this crate is to separate the task of writing the code
 //! to support peripherals of the atsamd families from the task of figuring out
-//! which specific devices has those peripherals.
+//! which specific devices have those peripherals.
 //!
 //! The actual mapping of devices to peripherals is specified in the
 //! `devices.yaml` file. In the `atsamd-hal` crate you then only need to care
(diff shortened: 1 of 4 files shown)


atsamd-hal-macros/src/lib.rs
```diff
@@ -61,7 +61,7 @@ fn hal_cfg_impl(args: TokenStream) -> Result<Group, Error> {
     Ok(cfgs)
 }
 
-/// Macro which expands to a `mod foo;` item with different paths for each
+/// Macro, which expands to a `mod foo;` item with different paths for each
```
ianrrees, line 64: I think if we're going to bother with very small edits like this, it's worth knocking the two-line summary down to one line.

hal/src/gpio/mod.rs
```diff
@@ -16,7 +16,8 @@
 //!
 //! ## Features
 //!
-//! - Converting between pin modes no longer requires access to the `Port` type.
+//! - Converting between pin modes no longer requires access to the
```
ianrrees, line 19: Let's just get rid of this whole reference to the old GPIO module - it was useful when we had two GPIO versions in the HAL but now only of historical interest

hal/src/typelevel.rs
```diff
@@ -152,11 +152,7 @@
 //! parameters in place of struct fields to represent the unknown types.
 //!
 //! ```
-//! struct Container<A, B>
-//! where
-//!     A: EnumOne,
-//!     B: EnumTwo,
-//! {
+//! struct Container<A: EnumOne, B: EnumTwo> {
```
ianrrees, line 155: Are this and the other changes to eliminate the `where` clauses personal preference, or suggested by a tool?  I feel the old way is more illustrative, and since they are in docstrings this change seems to be moving in the wrong direction.

ianrrees (commented):
I appreciate your recent interest and PRs, @nyurik !  The currently open three PRs are for relatively small changes, it would help if you could put them in to one PR - maybe using different commits to separate the different type of change.

nyurik:
@ianrrees , thx for such a quick review!  The `where` clause was not suggested by any tool, but rather me reading through the docs and stumbling on this, which caused me to create this PR (which included a few minor doc changes). It is totally a personal preference that if the code can be expressed concisely, it should be, until it no longer fits on a single line - and only then it should be broken up into the `where` clause. I was reading all the docs carefully, and it took me a moment to get through the syntax - hence the reason for the PR - just trying to make it a bit easier to read.

I am happy to make a bigger single PR - some projects simply require smaller and easier to review changes - e.g. if i am editing the docs, i shouldn't change the cargo.toml - as that may hide a bug - but i will be happy to merge all PRs into one

nyurik:
closing in favor of #872

PR #874, opened 2025-05-11
Update Wio Terminal embedded-sdmmc dependency

Cooler1989:
Update from 0.3 to 0.8 version.
Sacrifice delay to embedded sdmmc.
Possibly missing SPI interface baud to higher speed after controller initialization.

# Summary
[describe your changes here]

# Checklist

## If Adding a new Board

#### Note

### DIFF
--- a/boards/wio_terminal/Cargo.toml
+++ b/boards/wio_terminal/Cargo.toml
@@ -38,7 +38,7 @@ display-interface-spi = "0.5"
 heapless = "0.8"
 ili9341 = "0.6.0"
 lis3dh = "0.3.0"
-embedded-sdmmc = "0.3.0"
+embedded-sdmmc = "0.8.0"
 usb-device = { version = "0.3", optional = true }
 nb = "1.0"
 bbqueue = { version = "0.5", optional = true }
--- a/boards/wio_terminal/examples/sdcard.rs
+++ b/boards/wio_terminal/examples/sdcard.rs
@@ -3,8 +3,9 @@
 
 /// Makes the wio_terminal read the SD card and print the filenames
 /// of the first few entries.
-/// In case of missing SD card, it will be stuck in an infinite loop inside of the embedded-sdmmc crate
-/// trying to initialize and not giving any feedback to the user.
+/// In case of missing SD card, it will be stuck in an infinite loop inside of
+/// the embedded-sdmmc crate trying to initialize and not giving any feedback to
+/// the user.
 use embedded_graphics as eg;
 use panic_halt as _;
 use wio_terminal as wio;
--- a/boards/wio_terminal/examples/sdcard.rs
+++ b/boards/wio_terminal/examples/sdcard.rs
@@ -3,6 +3,8 @@
 
 /// Makes the wio_terminal read the SD card and print the filenames
 /// of the first few entries.
+/// In case of missing SD card, it will be stuck in an infinite loop inside of the embedded-sdmmc crate
+/// trying to initialize and not giving any feedback to the user.
 use embedded_graphics as eg;
 use panic_halt as _;
 use wio_terminal as wio;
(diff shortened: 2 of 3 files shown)


jbeaurivage:
Thanks @Cooler1989. Did you test that the updated examples work?
It's unfortunate, had I known that you would update this dependency, I would've waited before releasing `wio_terminal`. It would have saved us a breaking release. No big deal though. :)

Cooler1989:
Before merging, let me check how the things are working in old version while you insert the microSD card later on.

jbeaurivage:
Regarding speeding up the SPI bus:

This example is based on `embedded-sdmmc` v0.5 (I think?) but hopefully it should work fine with the latest version.
```rs
use fugit::RateExtU32;
let sd_device = volume_mgr.device();
// Bring the SPI clock to a higher speed. This must be done after
// the card has been initialized.
sd_device.spi(|spi| spi.reconfigure(|config| config.set_baud(12.MHz())));
```

Cooler1989:
> Regarding speeding up the SPI bus:
> 
> This example is based on `embedded-sdmmc` v0.5 (I think?) but hopefully it should work fine with the latest version.
> 
> ```rust
> use fugit::RateExtU32;
> let sd_device = volume_mgr.device();
> // Bring the SPI clock to a higher speed. This must be done after
> // the card has been initialized.
> sd_device.spi(|spi| spi.reconfigure(|config| config.set_baud(12.MHz())));
> ```

@jbeaurivage, what do you think about leaving the clock at low speed for simplicity of the example? 
Otherwise, here it should be always  toggled before initialization and after initialization, because someone may take out and insert again the microSD, which then needs slow clocks for configuration.
I think this was not covered in previous version of the example.

jbeaurivage:
Really it's up to you. I have no issues leaving the bus at low speeds. Usually our bsp examples can have two purposes:
* Provide a minimal working code, which people can use as a starting point to write their own applications
* Showcase and test our APIs

I have no issues keeping it simple, it's still a good starting point for users to understand how to initialize the card.

Cooler1989:
I will add two more notes as comments for users to get this information and I am also fine with merging as it now i.e. far from ideal.

jbeaurivage:
Sounds good. And you said the example still works without providing a delay type to the SpiDevice? I think that's ok, if you used the NoDelay type from ehal-bus, I think it would panic if the library attempted to actually wait on the delay.

Cooler1989:
@jbeaurivage, the `delay` was not used by Spi neither it is used now. I am not even sure what scenarios it causes it to use. 
The sum up of the `delay` for this change is: 
Now: 
1) borrowed for display initialization, 
2) handed over to sdmmc driver for internal usage (I don't like the aproach. How can one clone it for other uses? Do we have any examples?)

Before: 
1) used for the example app purpose: 
   a) borrowed for display init
   b) used for delay between loop iteration (this delay is gone with this change)

jbeaurivage:
I would assume that the SD driver takes ownership of the Delay object to avoid having to deal with borrows and lifetimes, which ultimately makes it much easier to move around. I don't think it's a big deal to let the driver eat the delay in the example... we want to keep things simple. If you wanted a cloneable delay type, you'd need something like a timer queue. RTIC's `Monotonic` has something similar, which works well out of the box with either the SYSTICK or RTC timers, but the methods are async. In any case, probably too much complexity for what should be a simple example.

jbeaurivage:
If you want to rebase your branch and fix formatting, I'll be happy to merge this!

Cooler1989:
I would very much like to do that, but currently I am not able to tell what is wrong. Running rustfmt.sh locally do not show any changes for git-tracked files.

jbeaurivage:
You just had to run `cargo fmt` with nightly :)
https://github.com/Cooler1989/atsamd_wio/pull/1

Cooler1989:
Possibly, I had a lot of those: 
```
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
```

jbeaurivage:
Yup, that makes sense! If you're ready to merge, just mark the PR as ready for review and I'll take care of it. Thanks!

PR #875, opened 2025-05-12
Remove deprecated items and repo cleanup

### LINKED ISSUE #784: Remove deprecated code - embedded-hal 0.2

jbeaurivage:
# Summary
Since we have a breaking release lined up, it's a good opportunity to sneak in breaking changes to clean up the repo.

* Bump MSRV to 1.85.1 and Rust edition to 2024 (`atsamd-hal` and `atsamd-hal-macros`).
* `rustfmt` clean up, mostly imports in alphabetical order
* Change CI to use `cargo fmt` instead of  `rustfmt`. It will automatically respect the individual
    crates' Rust editions.
* Remove some unused dependencies: `void` and `vcell`
* Update `defmt`, `rng-core` and `embedded-sdmmc` dependencies
* Remove a bunch of deprecated code:
- [x] `spi_future`
- [x] Some items previously used for SERCOM+DMA. Now the `embedded-hal` implementations can be used with DMA instead.
- [ ] `embedded-hal v0.2`. This was long overdue!

Closes #784.

~~I'd like to wait on #814 before merging, because there's no point in touching the `adc` module - which has some old ehal 0.2 impls - here, since it underwent a major rewrite.~~ #814 is merged.

## Note to maintainers:

Please merge this PR using **rebase. Please do not squash the commits!**


nyurik:
would it make sense to rename `gen` to something that is not a keyword in 2024?  I love the overall cleaner feel!

ianrrees:
A little bit tangential, but what do you think about this - seems like we can reinforce what the comment says in code:
```diff
diff --git a/hal/src/async_hal/interrupts.rs b/hal/src/async_hal/interrupts.rs
index 677eba8aad..0e9a4b09a2 100644
--- a/hal/src/async_hal/interrupts.rs
+++ b/hal/src/async_hal/interrupts.rs
@@ -224,7 +224,7 @@ declare_interrupts!(ADC);
 /// once.
 ///
 /// May not be implemented outside of this HAL.
-pub trait InterruptSource: crate::typelevel::Sealed {
+pub(crate) trait InterruptSource: crate::typelevel::Sealed {
     /// Enable the interrupt.
     ///
     /// # Safety

```

ianrrees:
I've had a skim over this and think it looks like good progress.  There's a pattern that might warrant a bit more thought though: we've got a few instances where there's a method (often part of a trait) that was marked `unsafe` and now the implementation of that method also needs to contain `unsafe` blocks - so we're gaining a lot of `unsafe` keywords.

I suspect that some of those methods were declared `unsafe` because all/most implementations would be `unsafe`- IOW, it was more to make code tidy/pragmatic, not necessarily because the API is unsafe.  If that is the case, I think we should remove the `unsafe` from those APIs.  The `InterruptSource::enable()` trait method might be a good talking point on this: to me it looks like it's not really `unsafe`, wouldn't calling it in a critical section be a logic bug?

jbeaurivage:
> A little bit tangential, but what do you think about this - seems like we can reinforce what the comment says in code:
> 
> ```diff
> diff --git a/hal/src/async_hal/interrupts.rs b/hal/src/async_hal/interrupts.rs
> index 677eba8aad..0e9a4b09a2 100644
> --- a/hal/src/async_hal/interrupts.rs
> +++ b/hal/src/async_hal/interrupts.rs
> @@ -224,7 +224,7 @@ declare_interrupts!(ADC);
>  /// once.
>  ///
>  /// May not be implemented outside of this HAL.
> -pub trait InterruptSource: crate::typelevel::Sealed {
> +pub(crate) trait InterruptSource: crate::typelevel::Sealed {
>      /// Enable the interrupt.
>      ///
>      /// # Safety
> ```

I think it should remain public - the idea of `InterruptSource` is to have methods that act on all interrupt lines related to a single peripheral at once - eg `SERCOM0_0, SERCOM0_1, SERCOM0_2, SERCOM0_OTHER`. The trait is sealed so users of the crate can't actually implement it, but it may still be useful to them to be able to enable/disable/unpend/set the priority of all those interrupts at once.

jbeaurivage:
> There's a pattern that might warrant a bit more thought though: we've got a few instances where there's a method (often part of a trait) that was marked `unsafe` and now the implementation of that method also needs to contain `unsafe` blocks - so we're gaining a lot of `unsafe` keywords.
> 
> I suspect that some of those methods were declared `unsafe` because all/most implementations would be `unsafe`- IOW, it was more to make code tidy/pragmatic, not necessarily because the API is unsafe. If that is the case, I think we should remove the `unsafe` from those APIs. The `InterruptSource::enable()` trait method might be a good talking point on this: to me it looks like it's not really `unsafe`, wouldn't calling it in a critical section be a logic bug?

I'll go through the code and see if we can't mark some unsafe functions as safe. Though we still have to be careful to guarantee that all the invariants of the unsafe functions we're calling are still respected at all times, regardless of the inputs of our own (newly marked safe) functions.

In the specific case of `InterruptSource::enable()`, it can definitely cause UB by breaking the invariants of critical-section based `Mutex` if called inside a critical section. 

**EDIT**: Looking at the functions where the new `unsafe` blocks had to be added, it seems like for all of them, marking them as `unsafe` is still warranted.

ianrrees:
> > A little bit tangential, but what do you think about this - seems like we can reinforce what the comment says in code:
> > ```diff
> > diff --git a/hal/src/async_hal/interrupts.rs b/hal/src/async_hal/interrupts.rs
> > index 677eba8aad..0e9a4b09a2 100644
> > --- a/hal/src/async_hal/interrupts.rs
> > +++ b/hal/src/async_hal/interrupts.rs
> > @@ -224,7 +224,7 @@ declare_interrupts!(ADC);
> >  /// once.
> >  ///
> >  /// May not be implemented outside of this HAL.
> > -pub trait InterruptSource: crate::typelevel::Sealed {
> > +pub(crate) trait InterruptSource: crate::typelevel::Sealed {
> >      /// Enable the interrupt.
> >      ///
> >      /// # Safety
> > ```
> 
> I think it should remain public - the idea of `InterruptSource` is to have methods that act on all interrupt lines related to a single peripheral at once - eg `SERCOM0_0, SERCOM0_1, SERCOM0_2, SERCOM0_OTHER`. The trait is sealed so users of the crate can't actually implement it, but it may still be useful to them to be able to enable/disable/unpend/set the priority of all those interrupts at once.

Got it, that makes sense, thanks!

ianrrees:
> In the specific case of InterruptSource::enable(), it can definitely cause UB by breaking the invariants of critical-section based Mutex if called inside a critical section.

Hmm, sorry I'm taking the conversation a bit outside the scope of this PR, but I think the usual critical section mutex disables interrupts in the ARM core where this method is about the peripheral interrupt sources.  Without digging in to it, my hunch is that if calling that method in a certain context can cause UB, that's actually indicating a bug elsewhere - like in the mutex implementation.

But anyway, this is off-topic for the PR!  It looks good to me - shall we merge it?

jbeaurivage:
You're right that the critical section impl disables all interrupts in the core. I actually took most of the interrupt-handling stuff from [`embassy`](https://github.com/embassy-rs/embassy/blob/206a324cf4d612122356fb350b4a3b56391d6f20/embassy-hal-internal/src/interrupt.rs#L48). Feel free to open an issue if you feel this should be discussed further!

I think this is ready to merge. I wanted to also remove [`ehal v0.2`](https://github.com/jbeaurivage/atsamd/tree/remove-ehal-02), but I think it should just go into a follow up PR at this point.

ianrrees:
I've used this PR to run [an experiment](https://github.com/atsamd-rs/atsamd/wiki/June-2025-release%E2%80%90plz-experiment) around our release process.  For that, I modified the commit messages locally then merged in to master from my machine - unfortunately GitHub didn't understand that workflow so I'm closing this PR manually.

ianrrees:
> You're right that the critical section impl disables all interrupts in the core. I actually took most of the interrupt-handling stuff from [`embassy`](https://github.com/embassy-rs/embassy/blob/206a324cf4d612122356fb350b4a3b56391d6f20/embassy-hal-internal/src/interrupt.rs#L48). Feel free to open an issue if you feel this should be discussed further!

I guess in this instance it's a bit academic, since the trait can only be implemented in the HAL anyway.  Maybe just something to keep in mind next time we're in that area.

PR #878, opened 2025-05-19
Arduino MKR1000: update to hal v0.16

ojob:
Trying to use async samples on my Arduino MKR1000, I propose to bring code and examples from arduino_mkrzero based on same hal version (v0.16). Then I will try to update up to current stable version (0.21?).

My plan is then to add documentation towards the [existing WiFi lib for ATWINC1500](https://github.com/kaidokert/winc-rs/tree/main) in the README.

Rationale for the update: I'm quite new to embedded-rust, and found difficult to identify that I could use `arduino_mkrzero` directly on `arduino_mkr1000`, then to understand how to adapt it to my use case beyond the `blinky_basic` example. Having a directly-applicable starter examples would have saved me quite a lot of hours.

# Summary
- update `boards/arduino_mkr1000/Cargo.toml` to match `arduino_mkrzero`, in particular ` atsamd-hal=0.16`
- structurize code according to `arduino_mkrzero`, with creation of modules `pins.rs`, `sercom.rs` (but without the SD card support), `usb.rs`.
- bring examples from `arduino_mkrzero` to `boards/arduino_mkr1000/examples/`

# Checklist
  - [x] All new or modified code is well documented, especially public items : taken as-is from existing `arduino_mkrzero` (apart from SD functionality, that is not present on MKR1000)
  - [ ] All examples run correctly on real hardware

### DIFF
--- a/boards/arduino_mkr1000/Cargo.toml
+++ b/boards/arduino_mkr1000/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "arduino_mkr1000"
-version = "0.5.1"
+version = "0.6.0"
 authors = [
     "Eric Rushing <rushinge@gmail.com>",
     "Joël Bourgault <ojob-info@ouvaton.org>",
--- a/boards/arduino_mkr1000/Cargo.toml
+++ b/boards/arduino_mkr1000/Cargo.toml
@@ -1,7 +1,10 @@
 [package]
 name = "arduino_mkr1000"
 version = "0.5.1"
-authors = ["Eric Rushing <rushinge@gmail.com>"]
+authors = [
+    "Eric Rushing <rushinge@gmail.com>",
+    "Joël Bourgault <ojob-info@ouvaton.org>",
+]
 description = "Board Support crate for the Arduino MKR 1000 WiFi"
 keywords = ["no-std", "arm", "cortex-m", "embedded-hal"]
 categories = ["embedded", "hardware-support", "no-std"]
@@ -15,7 +18,7 @@ version = "0.7"
 optional = true
 
 [dependencies.atsamd-hal]
-version = "0.14"
+version = "0.16"
 default-features = false
 
 [dependencies.usb-device]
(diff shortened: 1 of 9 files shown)


ojob:
About following checklist item :

```md

```

I'm not sure that I have something to do, considering that `/crates.json` uses a `--all-features` flag to build examples :

```json

    "arduino_mkr1000": {
      "tier": 2,
      "build": "cargo build --examples --all-features",
      "target": "thumbv6m-none-eabi"
    },
```

ianrrees:
Thanks @ojob !

The item about features applies to the HAL, where you're updating a BSP, so you're correct that it doesn't apply to this Pull Request.  The top level [README.md](https://github.com/atsamd-rs/atsamd/blob/master/README.md) has a little bit about what the PACs, HAL, BSPs are - though it is somewhat terse.

Have you tested all the examples on your hardware with this branch?

ojob:
@ianrrees thanks for the feedback!

I did not test yet, but I must obviously do so 😉  (It seems that I put much confidence in the compiler not complaining!) Maybe this should be a step added to the merge request template?

ianrrees:
The compiler is very good, but we do see problems that it doesn't catch.  Please update here when you're done testing, thanks!

ojob:
Pins were not correct, for the led and the PWM sample, and now `blinky_rtic` and `pwm` examples are running with hal `v0.16`.

I found out that mapping PWM to pins is not trivial: datasheet for SAMD21 §7 "multiplexing" exposes names like `TCC0/
WO[5]`, but trying to use `hal::pwm::Channel::_5` fails. Said otherwise, it seems that some PWMs may not be reachable.

This `Channel` enum lies in `/hal/src/peripherals/pwm/` in either `d5x.rs` or `d11.rs`, with a conditional use in `/hal/src/peripherals/mod.rs`, but the relation though `hal_module` to the `arduino_mkr1000` board and components is quite obscure to me. Any ideas?

ianrrees:
I'll have a closer look later, but the high-level idea is that the BSP's dependency on the HAL specifies the target chip via a feature flag for the family of chip (in this PR's Cargo.toml diff, this is `default = ["rt", "atsamd-hal/samd21g", "usb"]`).  In the HAL (from memory, mostly in /hal/Cargo.toml), we break the chip family down in to finer-grained configurations to pick eg the d11.rs you referred to, including pin multiplexing information.  This allows us to use the Rust type system to throw compile errors for code that tries to use invalid combinations of pins, for instance.

So, could you share what you tried, and what the error was?  If there is a pin mapping mismatch between what the HAL supports and what the datasheet says, then that's likely a bug in the HAL.  But I should say there is a lot of complexity in this area, and only some of that is captured int hat multiplexing table from the datasheet.

ianrrees:
Hi @ojob, just checking that you're happy for us to rebase then merge this PR?

As for using PWM wave output 5, there are 4 capture/compare channel in the TCC peripheral, and that's what the "Channel" enum selects between - not the waveform outputs.  It may be possible to somehow make the TCC output PWM on the other channels, and our docs could be clearer, but this is outside the scope of this PR so I think should be handled separately.

ojob:
Hi @ianrrees, yes indeed, basic functionalities work, even if the mapping is incomplete.

So I just rebased the branch, and if you're happy with merging, I wouldn't be happier! However, I see many failures with `nightly`, maybe this first needs a bit more attention… :disappointed: 

This summer I'll push the work with more complete pin mapping and update to more recent versions of the HAL.

ianrrees:
Cheers!  Yes, I believe there's a fix in the pipeline for those CI failures - they're just a minor warning elevated to error:

```
   Compiling atsamd-hal-macros v0.2.5 (/home/runner/work/atsamd/atsamd/atsamd-hal-macros)
    Checking atsamd-hal v0.22.0 (/home/runner/work/atsamd/atsamd/hal)
error: enum `Compare0` is never used
   --> hal/src/rtc/modes.rs:290:34
    |
290 |     create_rtc_interrupt!(mode0, Compare0, cmp0);
    |                                  ^^^^^^^^
    |
```

PR #879, opened 2025-06-02
refactor(sercom): clean up underlying pad data structures

FraFrieFa:
This change lays the groundwork for implementing synchronous and half-duplex UART by cleaning the pad modules. Each commit compiles on its own.

- Introduce `hal/src/sercom/pads.rs`
  - A four-slot, type-level `Pads<P0, P1, P2, P3>` container, representing GPIO pins for one Sercom (each Sercom has up to 4 Pads: Pad0, Pad1, Pad2, Pad3 which can only be assigned to one Pad each -> suitable struct for storing underlying pins)
    - Holds up to 4 pads, Sercom is automatically determined
  - A macro `roles_type` to combine Pads with protocol roles (Rx, Tx, ...)
  - `ValidPads` trait
    - checks all pads share one Sercom (`ShareSercom`)
    - thumbv7em: also checks a common IO set (`ShareIoSet`)
- Replaced the implementation for Sercom protocols with the new `Pads` + `Roles` structures
- Pad access traits ```PadSet``` and ```ValidPads``` have been reduced to ```ValidPads```
- Delete all thumbv6m/thumbv7em “pad” files for I²C, SPI, and UART
- Update all board BSPs and examples
  - Remove `SercomX` arguments in pad aliases (e.g. `uart::Pads<Sercom0, …>`)
  - All examples now have to convert all raw pins via `.into()` to pick the correct alternate function

**TODO:**
- [x] Verify `python3 build-all.py` runs for each commit
- [ ] Verify that the new `Pads` behave like the old ones (no invalid configurations)
- [ ] Fix Clippy lints and ensure CI passes on all chips
- [ ] Consider ergonomic constructors to avoid manual `.into()` calls (the old code automatically converted a `Pin` into the correct `Pad`, but replicating that behavior requires specifying a Sercom) A `DormantPads` struct might hold GPIO pins until it is activated by claiming a Sercom slot (for example in the `Uart` constructor), coverting them to their alternate functions

This intermediate refactoring state is shared for review and to help the team become familiar with the changes.


ianrrees:
Not directly about this PR (which, sorry, I've only skimmed over but so far it looks good!), but I've been thinking we should use tests to cover typelevel code, maybe using a technique like one of the answers to https://stackoverflow.com/questions/74989492/how-to-assert-that-code-wont-compile-in-a-test

PR #880, opened 2025-06-03
Implements the `ehal::delay::DelayNs` trait for `timer::TimerCounter`

kyp44:
# Summary

In doing some other work, I noticed that [`TimerCounter`](https://docs.rs/atsamd-hal/latest/atsamd_hal/timer/struct.TimerCounter.html) implements the `embedded-hal` v0.2 traits ([`CountDown`](https://docs.rs/embedded-hal/0.2.7/embedded_hal/timer/trait.CountDown.html) and [`Periodic`](https://docs.rs/embedded-hal/0.2.7/embedded_hal/timer/trait.Periodic.html)) but not the `embedded-hal` v1.0 [`DelayNs`](https://docs.rs/embedded-hal/1.0.0/embedded_hal/delay/trait.DelayNs.html) trait, noting that this is implemented by [`delay::Delay`](https://docs.rs/atsamd-hal/latest/atsamd_hal/delay/struct.Delay.html) and [`rtc::Rtc<Count32Mode>`](https://docs.rs/atsamd-hal/latest/atsamd_hal/rtc/struct.Rtc.html) with PR #845.

It was also noticed that there was quite a bit of code duplication between the different two main chip variants for `TimerCounter`.

This implements `DelayNs` for `TimerCounter` and also does a bit of refactoring to pull out common code between the two variants, reducing code duplication.

# Checklist

### DIFF
--- a/hal/src/timer_params.rs
+++ b/hal/src/timer_params.rs
@@ -38,10 +38,6 @@ impl TimerParams {
 
         let cycles: u32 = ticks / divider;
 
-        if cycles > u16::MAX as u32 {
-            panic!("cycles {} is out of range for a 16 bit counter", cycles);
-        }
-
         TimerParams {
             divider: divider as u16,
             cycles,
--- a/hal/src/timer_params.rs
+++ b/hal/src/timer_params.rs
@@ -9,20 +9,30 @@ pub struct TimerParams {
 }
 
 impl TimerParams {
-    /// calculates TimerParams from a given frequency based timeout.
+    /// Calculates the [`TimerParams`] from a given frequency based timeout.
+    ///
+    /// Panics if the combination of `timeout` and `src_freq` cannot be done
+    /// with a 16-bit timer.
     pub fn new(timeout: Hertz, src_freq: Hertz) -> Self {
         let ticks: u32 = src_freq.to_Hz() / timeout.to_Hz().max(1);
-        Self::new_from_ticks(ticks)
+        let ret = Self::new_from_ticks(ticks);
+        ret.check_cycles_u16();
+        ret
     }
 
-    /// calculates TimerParams from a given period based timeout.
+    /// Calculates the [`TimerParams`] from a given period based timeout.
+    ///
+    /// Panics if the combination of `timeout` and `src_freq` cannot be done
+    /// with a 16-bit timer.
     pub fn new_ns(timeout: Nanoseconds, src_freq: Hertz) -> Self {
         let ticks: u32 =
             (timeout.to_nanos() as u64 * src_freq.to_Hz() as u64 / 1_000_000_000_u64) as u32;
-        Self::new_from_ticks(ticks)
+        let ret = Self::new_from_ticks(ticks);
+        ret.check_cycles_u16();
+        ret
     }
 
-    fn new_from_ticks(ticks: u32) -> Self {
+    pub(crate) fn new_from_ticks(ticks: u32) -> Self {
         let divider = ((ticks >> 16) + 1).next_power_of_two();
         let divider = match divider {
             1 | 2 | 4 | 8 | 16 | 64 | 256 | 1024 => divider,
@@ -46,7 +56,7 @@ impl TimerParams {
 
     /// Returns the number of required `cycles` as a `u16` and panics if the
     /// number is too high to fit.
-    pub fn check_cycles_u16(&self) -> u16 {
+    pub(crate) fn check_cycles_u16(&self) -> u16 {
         match u16::try_from(self.cycles) {
             Ok(c) => c,
             Err(_) => panic!(
--- a/hal/src/timer_params.rs
+++ b/hal/src/timer_params.rs
@@ -43,6 +43,18 @@ impl TimerParams {
             cycles,
         }
     }
+
+    /// Returns the number of required `cycles` as a `u16` and panics if the
+    /// number is too high to fit.
+    pub fn check_cycles_u16(&self) -> u16 {
+        match u16::try_from(self.cycles) {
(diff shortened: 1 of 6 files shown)


ianrrees:
Thanks for keeping these PRs updated @kyp44 , I'm hoping to have a closer look and merge soon.

hal/src/timer_params.rs
```diff
@@ -38,10 +38,6 @@ impl TimerParams {
 
         let cycles: u32 = ticks / divider;
 
-        if cycles > u16::MAX as u32 {
```
ianrrees, line 41: I'm curious why remove this check?  I think the code above could possibly be written more clearly using range-based match and dropping the call to `next_power_of_two()`, but it's possible to hit the error condition (eg, if ticks = 0x4000000).

kyp44: So this is a bit of a subtle issue, but this check that panics was taken out of `TimerParams` and moved out to the users of `TimerParams`, for example [here](https://github.com/kyp44/atsamd/blob/fa1fe6c5dadbb596cb87715b64e82dba7cb0c2f7/hal/src/peripherals/timer/common.rs#L32-L46).

The reason is that, for the `DelayNs` implementation in `TimerCounter`, because this blocks, we want to be able to handle longer times (i.e. time in nanoseconds at full `u32` width) without panicking, but instead by delaying using the timer at `u16::MAX` multiple times as needed, as done [here](https://github.com/kyp44/atsamd/blob/fa1fe6c5dadbb596cb87715b64e82dba7cb0c2f7/hal/src/peripherals/timer/common.rs#L91-L116).

However, it does look like I neglected to assess the impact of this change to `TimerParams` on the only other thing that seems to use it, which is the [PWM driver](https://github.com/kyp44/atsamd/tree/fa1fe6c5dadbb596cb87715b64e82dba7cb0c2f7/hal/src/peripherals/pwm). I just pushed a new commit that addresses this in what I think is a good way.

ianrrees: Gotcha, thanks!

Since `TimerParams` is public and the provided ctors call `new_from_ticks()`, I think we should continue to check that the ctors return cycles that will fit in 16b and panic otherwise.  I don't think that behaviour is great, but changing it is outside the scope of this PR (maybe a good compromise is to add a note in the docs for the ctors).

Down the line, we might think about making the `TimerParams` count field generic, so that the check would be appropriate for the timer it's going to be used with.

kyp44: I just pushed a new commit that reverts `TimerParams` and the `pwm` drivers back. Unfortunately, doing so does require a partial re-implementation of `TimerParams::new_from_ticks` in the [`DelayNs` implementation](https://github.com/kyp44/atsamd/blob/cb4918c3e18ec82c060288080837a5250441ad93/hal/src/peripherals/timer/common.rs#L87-L122). I'm not really seeing a good way for this and `TimerParams` to be able to share the same code here.

It's not clear to me why `TimerParams` was left public. As far as I can see nothing in the HAL requires or even allows dealing directly with ticks/cycles, everything is (rightfully) in terms of times and frequencies. It is what is it though.

I do think **Panic** sections should be added to the ctors documentation for `TimerParams`, but I'm not not sure whether you think that's appropriate for this PR or not. If so, I will go ahead and add them.

ianrrees: Cheers!  Rather than reverting all that good work, I wonder if we couldn't start from 56ef6056257673547f8c2f64ca55ad83dc4b0a72 and use the `check_cycles_u16()` in those ctors - something like:

```diff
diff --git a/hal/src/timer_params.rs b/hal/src/timer_params.rs
index 36a4807ce9..c33564b64c 100644
--- a/hal/src/timer_params.rs
+++ b/hal/src/timer_params.rs
@@ -9,17 +9,29 @@ pub struct TimerParams {
 }
 
 impl TimerParams {
-    /// calculates TimerParams from a given frequency based timeout.
+    /// Calculates TimerParams from a given frequency based timeout.
+    /// 
+    /// # Panics
+    /// Panics if the combination of timeout and src_freq can't be done with a
+    /// 16b timer.
     pub fn new(timeout: Hertz, src_freq: Hertz) -> Self {
         let ticks: u32 = src_freq.to_Hz() / timeout.to_Hz().max(1);
-        Self::new_from_ticks(ticks)
+        let ret = Self::new_from_ticks(ticks);
+        ret.check_cycles_u16();
+        ret
     }
 
-    /// calculates TimerParams from a given period based timeout.
+    /// Calculates TimerParams from a given period based timeout.
+    /// 
+    /// # Panics
+    /// Panics if the combination of timeout and src_freq can't be done with a
+    /// 16b timer.
     pub fn new_ns(timeout: Nanoseconds, src_freq: Hertz) -> Self {
         let ticks: u32 =
             (timeout.to_nanos() as u64 * src_freq.to_Hz() as u64 / 1_000_000_000_u64) as u32;
-        Self::new_from_ticks(ticks)
+        let ret = Self::new_from_ticks(ticks);
+        ret.check_cycles_u16();
+        ret
     }
 
     fn new_from_ticks(ticks: u32) -> Self {

```

Yeah, I think improving documentation is always good!  Agree that `pub (crate)` seems more appropriate too.  I'd just rather not make a PR that is nominally about adding an impl, also make a somewhat subtle change to an existing API - good chance it wouldn't matter, but it's better to play it safe.

kyp44: That is a good idea to maximize re-use! I rolled back the latest commit (but saved it as another branch in case it's ever needed again) and pushed a new one that implements this, as well as improves the documentation. Let me know what you think.

PR #885, opened 2025-06-09
feat(adc): Load factory calibration values

QuartzShard:
# Summary
The samd21g17 ADC is not very usable, in my experience, without loading the factory calibration values. The hal ADC constructor doesn't do this, but does cause a software reset, so the only way to actually use the factory calibration values is to steal the pac::ADC after creating the hal struct, which I don't think is egrgonomic.
```rust
let adc = Adc::adc(ctx.device.adc, &mut ctx.device.pm, &mut clocks);
// SAFETY: Discard this reference before using the ADC. The adc init performs a software
// reset, so we have to do this to actually configure it.
let adc_pac = unsafe { &mut pac::Adc::steal() };
load_adc_factory_calibration(adc_pac);
```
This PR moves the calibration into the Adc::adc() constructor, avoiding this issue. I do not see many use cases where this would be undesirable; however, an additional `calib()` method could be added to the Adc to facilitate cleaner modification of these values in these cases.

Happy to take feedback, this is my first contribution to a project like this. Particularly, I was torn between adding a constant for the NVM address, or editing the one at `crate::calibration::ADDR` to be pub(crate). I opted to not edit the existing one.

# Checklist

#### Note

### DIFF
--- a/hal/src/peripherals/adc/d11.rs
+++ b/hal/src/peripherals/adc/d11.rs
@@ -25,6 +25,7 @@ pub struct Adc<ADC> {
     adc: ADC,
 }
 
+const CALIB_ADDR: *const u32 = 0x806020u32 as *const u32;
 impl Adc<pac::Adc> {
     /// Create a new `Adc` instance. The default configuration is:
     /// * 1/32 prescaler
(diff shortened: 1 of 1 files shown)


rnd-ash:
@QuartzShard FYI: this is already done in the ADC V2 branch, which should be merged relatively soon hopefully.

QuartzShard:
I hadn't spotted that, fair enough!

ianrrees:
Thanks @QuartzShard !  The PR looks good, I like the explanation and think the code changes are a good size/scope.  As Ashcon wrote, we've got this functionality covered in a pending (quite large) PR [here](https://github.com/atsamd-rs/atsamd/pull/814/files#diff-5f94d427df3904957855fdc8ab198c4bc55f341192ab144a1af690a263e5fe54R39-R44) (link is likely to break sometime in the future) so I'm going to close this one out but please don't hesitate to open PRs/issues in the future!

PR #887, opened 2025-06-12
ci: Temporarily allow `mismatched_lifetime_syntaxes` lint in PACs

jbeaurivage:
# Summary
`mismatched_lifetime_syntaxes` is a new lint introduced in rustc 1.84. PACs currently fail CI on nightly because of this.
Temporarily allow the lint. The `#![allow]` statement should be removed from `update-pacs.sh` when we start generating PACs with `svd2rust >= 0.36.2`: see https://github.com/rust-embedded/svd2rust/pull/936 (currently unreleased).


ianrrees:
@jbeaurivage I saw your messages in Rust Embedded about svd2rust - are you looking in to using a newer one?

jbeaurivage:
@ianrrees ideally yes, I'm in no rush, but I felt like updating the pacs would be better than adding another lint exception. Not if it breaks the HAL though!

Looks like on nightly, we have some new `dead_code` warnings introduced - but those don't show up on stable. Weird, but we should probably figure out why to make ci green again

PR #892, opened 2025-06-18
Clock v2 thumbv6m

ianrrees:
Resurrecting @bradleyharden's work on clockv2 for thumbv6m. Will begin by rebasing on to current master, then try to finish things up!


ianrrees:
Note to self: https://github.com/atsamd-rs/atsamd/pull/908

ianrrees:
I've spent some more time on this today, and have moved the thumbv6 clock v2 support up to just before #728 .  From a not-very-close inspection of the newer history of the project, this seems like the last major snag to resurrect this work.

It seems like using diff/patch (IOW Git rebase) isn't a good approach to update the thumbv6 clock v2 work on to the new config scheme, there are tons of small changes spread through the codebase.  I'm thinking that manually replicating the changes from the thumbv6 clock v2 work is the way forward, directly on to the current master.

If anyone else feels like taking a crack at this, please feel free!

ianrrees:
I've saved the branch that attempted to do this with rebase as [clock_v2_thumbv6m_rebase](https://github.com/ianrrees/atsamd/tree/clock_v2_thumbv6m_rebase) and started manually transcribing the work on top of current master here.

ianrrees:
I've advanced to somewhere between Bradley's second and third "stash" commits, the code at the current head (commit message "WIP stash 2 enable v2 for thumbv6m") comments out some important stuff but builds much of the clock v2 code for thumbv6m (with some warnings that should be addressed), and the commit before that cleanly builds for thumbv7em (though isn't tested to still work).  Some of the changes I've made are a little bit hacky; I'm trying to create a finer-grained history to keep things compiling in intermediate steps so that it'll be easier to go back and clean up that stuff once the bulk of the work is done.

ianrrees:
- [x] Need to build with stable Rust - currently one of the earlier commits here introduces a couple nightly features just to get a compile.
- [ ] There's a pattern in the new work, where a type alias is used to gain access to the peripheral register block.  We should standardize that across all peripherals.
- [ ] In several places, I've made fairly crude alternate implementations for thumbv6m and thumb7em chips, where Bradley's original work had finer conditionals on eg one a member of a struct.  Use `#[hal_macro_helper]` to deal with this instead
- [x] The DPLL in smaller chips doesn't support DCO filtering options, have put a hacky conditional in to get a build but need to tidy this up - remove the API for changing DCO filter in smaller chips, etc.
- [ ] There are quite a few things gated on eg `clock-d5x` that should instead use more specific `hal_cfg` arguments.
- [ ] I2S clocking needs attention - just turned it off
- [ ] Macro `define_apb_types` takes two booleans, but there are only two valid combinations of input - look in to cleaning that up.
- [x] Think up a naming scheme/structure to replace the `(old|new)-clock-system`, if it's not OK to keep using `sysctrl` and `mclk-oscctrl`.  This is an internal-only thing, so probably not too important, but old/new is a bit ambiguous in light of #915 (which describes new chips, some of which use SYSCTRL) and of course we might come across a third scheme!
- [ ] The GCLK implementation currently is a bit repetitive, perhaps `atsamd-hal-macros` provides (or could be extended to provide) a variable like `gclk-max` so that we could use `seq!()`.
- [ ] Update T1 BSPs
- [ ] Ensure flash wait states are (or, can be) managed correctly
- [ ] How to handle TC/TCC instances sharing clocks?  Maybe split this out to a separate issue
- [x] Should the DPLL `enable()` method call `wait_enabled()` at the end?

ianrrees:
Have made a bit more progress on this today, largely splitting up the last few commits in [clock_v2_thumbv6m_rebase](https://github.com/ianrrees/atsamd/tree/clock_v2_thumbv6m_rebase) to make them a bit easier to digest

ianrrees:
The bulk of the clock v2 support is now in place for SAMD21, I've not been exercising SAMD11 builds and those currently error.  There are a couple of Bradley's changes that still need to be ported, and the above list of TODOs, I'll try to chip away at it more tomorrow.

ianrrees:
Changes to clock v2 API for thumbv7 introduced by this PR:

- Removed nvmctrl argument from `clock_system_at_reset()`
- Changed field in `reset::Clocks` from `osculp32k_base: Enabled<osculp32k::OscUlp32kBase>` to `pub osculp: OscUlpClocks,`.  `OscUlpClocks` has `pub base: osculp32k::EnabledOscUlp32kBase,`

ianrrees:
I've done some basic testing on both Metro M0 and Metro M4, seems to work OK.  Will chip away at updating the tier-1 BSPs for thumbv6m chips, but this seems ready for wider testing and feedback if anyone's got spare cycles to work on it.

ianrrees:
Woohoo!  Green tick from the CI.

hal/src/clock/v2/pclk.rs
```diff
@@ -292,6 +356,95 @@ macro_rules! with_pclk_types_ids {
...
+            (Tc6Tc7 = 29, tc6_tc7)
+            (Adc0 = 30, adc)
+            (AcDig = 31, ac_dig)
+            (AcAna = 32, ac_ana)
+            (Dac = 33, dac)
+            (Ptc = 34, ptc)
+            #[hal_cfg("i2s")]
+            (I2S0 = 35, i2s0)
+            #[hal_cfg("i2s")]
+            (I2S1 = 36, i2s1)
```
rnd-ash, line 408: Missing GCLK_TCC3 (0x25)

rnd-ash:
@ianrrees this looks really good! After a long read of the PR over the past couple hours, I couldn't see anything drastically wrong with this, Ill get to testing later this week I hope, then I can hunt for some bugs

hal/src/clock/v2/apb.rs
```diff
@@ -348,81 +376,235 @@ macro_rules! define_apb_types {
...
+        Tc5 = (13, any, all)
+        Adc0 = (16, any, all)
+        Ac = (17, any, all)
+        Dac = (18, any, all)
+        Ptc = (19, any, all)
         #[hal_cfg("i2s")]
-        I2S = 10,
-        Pcc = 11,
+        I2S = (20, any, all)
+        Ac1 = (21, any, all)
```
rnd-ash, line 567: Missing TCC3 (bit 24)

https://ww1.microchip.com/downloads/en/DeviceDoc/SAM-D21DA1-Family-Data-Sheet-DS40001882G.pdf

rustfmt.sh
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
```
wt: A version of this fix has already been applied. Can you please check out the HEAD version of the file and either break this line out into another PR or remove it from the diff?

wt, line 1: #964 updated all the bash scripts to have a consistent shebang. This line can be removed if you want.

Issue #830, opened 2025-03-16
Consolidating BSPs and/or their examples


ianrrees:
As we gain more BSPs and examples in them, the maintenance burden is growing and my feeling is that this distracts from HAL development which is probably more valuable.  This issue explores possible options to improve the situation.

There are two almost-orthogonal concerns here: one is about how many examples/BSPs we maintain, and the other is about the approach to maintaining them.  The aim here is to work on the approach:

1. The quantity question is inherently subjective, and seems unresolvable without a rather dictatorial maintenance policy.
2. With a better approach to examples/BSPs, we might find that the maintenance burden eases enough that the quantity ceases to be a significant problem.
3. If we run in to problems with a different approach, we can undo that attempt without losing ground (presuming we maintain examples/BSPs throughout the attempt).  On the other hand, if we drop examples/BSPs then an undo would require dealing with bit rot in dropped examples/BSPs, and there's a real risk of irritating contributors and users.

## Project management program / templated examples
Move the examples from BSPs to a separate directory and transform them in to templates, so that eg `use metro_m0 as bsp;` becomes generic `use {{bsp}} as bsp;`, and add a utility that can render these templates out to the BSPs.  This is prototyped in https://github.com/ianrrees/atsamd/tree/manage

### Pro
 * Conceptually simple
 * Easy to integrate in to existing CI
 * No change from the perspective of a BSP end-user
   - It remains clear which examples are available for which board

### Challenges
 * The utility might need to update BSPs' Cargo.toml to meet examples' dependencies

### Con
 * Bespoke code generation makes the overall project less approachable, seems a bit dirty

## Crate per-example
Rather than providing examples using Cargo's functionality, provide them as stand-alone crates which select a target board using a Cargo feature.  To improve developer ergonomics, we probably want to transform/consolidate the BSPs (for each target chip?), so that the BSP provides little more than the chip-board pin mappings.  Doing this would remove most of the `#[cfg()]` logic from the example source code and Cargo.toml .

`build.rs` can be leveraged to ensure that exactly one board is selected, and populate the correct `memory.x` and `.cargo/config.toml` files.

This is prototyped in https://github.com/ianrrees/atsamd/tree/crate-per-example

### Pro
 * Less Cargo friction to turn an example in to a custom firmware
 * It's clear which dependencies are required by any particular example, as they're not all combined in to a single `[dev-dependencies]` section in a BSP.

### Challenges
[...]

PR #898, opened 2025-06-20
removed(hal)!: Remove `embedded-hal` v0.2 implementations

jbeaurivage:
# Summary
Finally, this was long overdue!

# Checklist


ianrrees:
This looks like exactly the sort of PR that would be good for testing #830 .  I won't have much time to work on it in the next few days, but I'll try to rebase that branch that adds the example management program on to the current master, then maybe we try putting this work on top of it to see what the example update process would look like.

Issue #810, opened 2025-01-10
Provide defmt feature in BSPs to enable atsamd-hal/defmt


ianrrees:
In a firmware that depends on a BSP, it's a bit awkward to use defmt for errors that originate in the HAL.  One needs to add the same version of the HAL as the BSP uses to Cargo.toml, to enable the defmt feature.

To resolve this, we should have all the BSPs (at least, that use a new-enough HAL) support a defmt feature themselves.

sajattack:
I think part of the difficulty here is that defmt requires an addition to the linker script?

ianrrees:
I'm not sure I follow - yes using defmt in a firmware requires modifying the linker script, but even with that in place it's not straightforward to use defmt in a firmware that depends on a BSP directly.

For example, if my firmware's Cargo.toml has something like this:

```toml
[dependencies]
metro_m4 = { version = "0.16.1", features = ["usb"] }
```

and in my firmware, I want to use something like:
```Rust
#[derive(Error, defmt::Format, Debug)]
enum Error {
    #[error("UART error:")]
    UartError(hal::sercom::uart::UartError),
   // ...more variants..
}
```

Currently, the only way AFAICT to do this, is to modify Cargo.toml to add a dependency on the HAL, using the same version as the BSP:
```toml
[dependencies.atsamd-hal]
version = "0.20.2"
default-features = false
features = ["defmt"]
```

What I'm proposing is that it would be nice to be able to just do:
```toml
[dependencies]
metro_m4 = { version = "0.16.1", features = ["usb", "defmt"] }
```

sajattack:
Oh, yeah, I know, I'm just saying adding the dep to the default linker script might be slightly annoying and I don't think it can be feature gated. But it's probably not an issue. Just something to consider.

ianrrees:
Ah, you're a step ahead of me: if the feature is enabled, but the linker script isn't modified, then the build fails.  This means that the `--all-features` builds as are done by CI would require us to modify the default linker script if we added this defmt feature to BSPs...

So yes, probably deserves a bit more thought, I'll remove the "good first issue" tag!

ianrrees:
I stumbled on [this](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-arg-bins) today - looks feasible to use a build script to set the linker flag when the defmt feature is selected.  By default, build scripts run when features change.

PR #902, opened 2025-06-20
example(grand_central_m4): Adding async dmac example

supersimple33:
# Summary
I added the async dmac example from the metro m4. 

# Checklist

### DIFF
--- a/boards/grand_central_m4/Cargo.toml
+++ b/boards/grand_central_m4/Cargo.toml
@@ -30,9 +30,17 @@ features = ["critical-section-single-core"]
 cortex-m = "0.7"
 usbd-serial = "0.2"
 panic-halt = "1.0.0"
+panic-probe = "1.0.0"
 panic-semihosting = "0.6"
 smart-leds = "0.3"
 ws2812-timer-delay = "0.3"
+embassy-executor = { version = "0.7.0", features = [
+	"arch-cortex-m",
+	"executor-thread",
+	"task-arena-size-8192",
+] }
+defmt = "1.0.1"
+defmt-rtt = "1.0.0"
 
 [features]
 default = ["rt", "atsamd-hal/samd51p"]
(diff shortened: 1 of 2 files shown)


ianrrees:
Thanks @supersimple33 !  In #830 we're currently discussing some options to make it a bit easier for us to maintain the BSPs and examples (currently one of the bigger challenges of maintaining this project) - it strikes me that this PR might make a good test case for a new scheme we're considering.  Would you be up for trying the proposed approach?  I think this would look like:

1. Add my fork as a remote of your local atsamd-rs repo and checkout the `manage` branch - LMK if you'd like the Git incantations to do that.
2. Change the filename of `/examples/metro_m4-async_dmac.rs` to `/examples/generic-async_dmac.rs`.
3. Edit `examples/examples.toml` to make a new section for the async_dmac example, specifying it's used by both metro_m4 and grand_central_m4.
4. cd in to `/manage` and do `cargo run -- example distribute ../examples ../boards`
5. Edit the /boards/grand_central_m4/Cargo.toml just like you've done here (I'd do this via `git checkout AsyncDMAC -- boards/grand_central_m4/Cargo.toml`)
6. Open a PR against my fork.

Since this is all work-in-progress, there's not much documentation yet and things might be a little rough around the edges.  Just curious to get some feedback on what the process looks like (noting that some of this is more involved as the work is in a different fork), and see what snags we encounter.

Also, apologies that this is a bit of a comment-and-run as I'm going to be AFK for a couple days starting shortly.

supersimple33:
Got it will do

ianrrees:
Sorry I've dropped the ball on this - we've decided to go with a different approach to deduplicating examples, and I started porting this work to the new scheme [here](https://github.com/ianrrees/atsamd/tree/async_dmac_examples) but realised it also means introducing defmt to boards that don't currently use it.

The defmt stuff seems like a separate can of worms; a while back I opened #810 to make it an optional feature of BSPs, so I'm thinking to resolve that using metro_m4 as a test case (partly as I have a Metro M4 to test on), then finally we can use your work in porting the async_dmac examples to metro_m4 and grand_central_m4.

PR #696, opened 2023-07-19
Dac

zakary-bt:
# Summary
[describe your changes here]

# Checklist
  - [ ] `CHANGELOG.md` for the BSP or HAL updated
  - [ ] All new or modified code is well documented, especially public items
  - [ ] No new warnings or clippy suggestions have been introduced (see CI or check locally)

## If Adding a new Board
  - [ ] Board CI added to `crates.json`
  - [ ] Board is properly following "Tier 2" conventions, unless otherwise decided to be "Tier 1"

## If Adding a new cargo `feature` to the HAL
  - [ ] Feature is added to the test matrix for applicable boards / PACs in `crates.json`

Issue #39, opened 2019-01-22
ATSAMD51 DAC support


tarcieri:
This is a tracking issue for supporting the ATSAMD51's DAC. I'm primarily interested in supporting the AdaFruit NeoTrellis M4 (see https://github.com/rust-embedded/wg/issues/286). Wiring it into an ADC abstraction (#35) sounds cool, but for now I'd be happy just getting it working with the [low-level API](https://docs.rs/samd21g18a/0.2.1/samd21g18a/dac/index.html).

The main thing I have to go off of is AdaFruit's port of the Teensy audio engine to these platforms. Here are a few relevant commits:

- "DM: working w/ dual dac only" https://github.com/adafruit/Audio/commit/6d1ee585adb0e39ef34abf6f55544f6e18fc1db4
- "DM: interrupts fix and update synthesis examples for SAMD51" https://github.com/adafruit/Audio/commit/4a8bc7144effb58905f3c445422e14dd928b6516
- "DM: fix DMA interrupt structure for DAC output, other SAMD51 updates" https://github.com/adafruit/Audio/commit/8f4325807503a655e30ef304da7d75c3d89dda6f

Is DMA going to be an issue?

For what it's worth, I'd like to attempt to port this PureData synthesis engine to this platform:

https://github.com/mhroth/ZenGarden

I started a project to do that here:

https://github.com/neobirth/zengarden.rs

This project supports NEON acceleration, although I'm not sure if it's compatible with the subset of NEON (DSP instructions) supported by the ATSAMD51(G19A?). Otherwise it appears to be an easy-to-embed PureData engine that runs inside of an external event loop written in C++ that presents a C interface.

sajattack:
The low-level API portion is auto-generated and will be part of my SAMD51 PR.

tgross35:
Since there is now ADC support in the HAL https://docs.rs/atsamd-hal/latest/atsamd_hal/adc/index.html and the PAC https://docs.rs/atsamd51g/latest/atsamd51g/adc0/index.html, could this be closed?

ianrrees:
I think so!

sajattack:
ADC != DAC?

tgross35:
Misread, thanks

ianrrees:
Whoops, thanks for the catch @sajattack!

PR #904, opened 2025-06-24
DAC Support (Sync + Async) for both D5x and D11/D21

rnd-ash:
# Summary
Adds DAC support for D5x and D11

Both implementations contain oneshot write functions (Just write a voltage and keep the DAC output going), and a DMA functions (Both blocking and async) for writing buffers to the DAC.

It is also possible to monitor the DAC output voltage via the ADC now with `Adc::read_dac0_output` (Both sync and async functions provided)

Closes #696 and #39

### DIFF
--- a/hal/src/peripherals/adc/d11/mod.rs
+++ b/hal/src/peripherals/adc/d11/mod.rs
@@ -6,6 +6,7 @@ use super::{
 #[cfg(feature = "async")]
 use super::{FutureAdc, async_api};
 
+use crate::dac::DacWriteHandle;
 use crate::{calibration, pac};
 use pac::Peripherals;
 use pac::Sysctrl;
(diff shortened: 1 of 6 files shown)


kyp44:
@rnd-ash I see that you've now taken full ownership of the `Pclk`, which is great. Can you change the name of the `release` method to `free`? I think this name should be consistent across the HAL and numerous other peripherals already use `free` (e.g. `Adc`, `Delay`, `Qspi`), so I think it makes sense to stick with that.

rnd-ash:
@kyp44 method name should be consistent now

sajattack (approved):
looks great, what's missing for d21 support?

rnd-ash:
I thought I supported it with this as well?...

sajattack:
> I thought I supported it with this as well?...

You did, I forgot the d11 file applied to both.

Issue #907, opened 2025-06-26
Clock API v2 DPLL Output Frequency Invalid Through Integer Division


fko-kuptec:
While integrating the clock API v2 into our firmware, I stumbled across the following issue:

The [code](https://github.com/atsamd-rs/atsamd/blob/b18cc12e2f5dc881a9b1789c0b0487227033f5ad/hal/src/peripherals/clock/d5x/v2/dpll.rs#L992) used to compute the output frequency for the DPLL is, as far as I understand, flawed. Because the fractional denomintor is only dividing the fractional part, it is truncating the fractional precision: 

```rust
self.input_freq() * (self.settings.mult as u32 + self.settings.frac as u32 / 32)
```

I wrote a script to compute the multiplier and fractional part needed to boost the external 32,768 kHz crystal frequency up to a usable `10^x` frequency. The resulting computation done by the PLL to boost up to even 160 MHz is:

```rust
32768 * (4882 + 26 / 32)
```

If I type that into a normal calculator, I get exactly 160,000,000 MHz. If I run this computation in Rust with integer precision, however, it results in 159,973,376 MHz.

This issue could be solved by restructuring the computation:

```
y = x * (a + b / 32)                 |  multiply an artifical 1 = 32 / 32
y = x * (a + b / 32) * 32 / 32       |  integrate the factor 32 into the parenthesis
y = x * (32 * a + 32 * b / 32) / 32  |  simply the fraction
y = x * (32 * a + b) / 32
```

Translated into rust, it would look like the following. Unfortunately, `u64` is needed for a part of the computation to avoid overflows:

```rust
let input = self.input_freq().to_Hz() as u64;
let multiplier_x32 = (32 * self.settings.mult as u32 + self.settings.frac as u32) as u64;
let output = (input * multiplier_x32 / 32) as u32;
output.Hz()
```

At least for my specific example,  this gives the expected output: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1ae2e8add9b31439921fb86565326cab)

I will create a PR to fix this. :)

bradleyharden:
@fko-kuptec, thanks for this. I don't remember what the largest multiplier is, but we may want to convert to `u64` first, before doing any math.

```rust
let input = self.input_freq().to_Hz() as u64;
let multiplier = (32 * self.settings.mult as u64 + self.settings.frac as u64);
let output = (input * multiplier / 32) as u32;
```

[...]

PR #908, opened 2025-06-26
fix: Maintain accuracy in DPLL frequency math

fko-kuptec:
Explained in issue #907


ianrrees:
One of the rustfmt changes is flagged by the nightly rustfmt, but it passes CI.  YOLO!

Issue #912, opened 2025-06-27
`clock::v2` API migration discussion and tracking


kyp44:
# Summary

This issue is meant to track and discuss issues related to migration to the clock v2 API.

The clock v2 API "...provides a simple, ergonomic, and most of all **safe** API to create and manage the clock tree in ATSAMD5x and E5x devices. It uses [type-level programming techniques](https://docs.rs/atsamd-hal/latest/atsamd_hal/typelevel/index.html) to prevent users from creating invalid or unsound clocking configurations."

The key feature compared to the [clock v1 API](https://docs.rs/atsamd-hal/latest/atsamd_hal/clock/index.html) is that the clock tree configuration is validated at compile time using type-level programming, which is better for most use cases in which the clock tree is set up once at the beginning of the program and then not changed.

# Status

Currently, the v2 API is only implemented for `thumbv7` targets, and is not widely supported throughout the HAL, making it currently largely unusable for most projects.

# Tasks

The following are pending identified tasks at the highest level:

1. Add clock v2 API support throughout the entire HAL for `thumbv7` targets.
2. Port the clock v2 API to `thumbv6m` targets.
3. Add clock v2 API support throughout the entire HAL for `thumbv6m` targets.

# Task 1: HAL support for `thumbv7` targets

Discussion is needed on the approach to take on this, but it should be uniform across the HAL for all chip variants. Some peripherals do already have support, but the way the support is implmeented does not seem to be consistent. Some discussion points, that will be converted to implementation guidelines as they are resolved:

1. Initially, should both clock v1 and v2 be supported for all peripherals?
2. Should the user select which API version to use via a cargo feature, causing the user-facing peripheral constructors to be different?
4. If not, should v2 support be added as an alternate constructor to preserve backward compatibility (as was done in [`Delay`](https://github.com/atsamd-rs/atsamd/pull/803)), or should the primary constructor be changed to use v2 and an alternate v1 constructor created, which breaks backward compatibility but would better faciliate eventually full migration as the v1 constructors can be deprecated.

Here are various PRs and issues related to this:

- PR #803 (Merged) adds alternate support to [`Delay`](https://docs.rs/atsamd-hal/latest/atsamd_hal/delay/struct.Delay.html).
- PR #814 (Merged) adds support for the [`Adc`](https://docs.rs/atsamd-hal/latest/atsamd_hal/adc/index.html), which seems to be mandatory.
[...]

PR #916, opened 2025-07-02
feat!: USB for thumbv7 using `clock::v2`

supersimple33:
# Summary

I have migrated the UsbBus to Clocks V2. That said I don't have a lot of experience with this stuff and I wasn't sure what I should consume and what I should hold onto or if this is even the right way to go about this. That said I tested my changes on the gc4 and they worked. 

See #912


kyp44:
This looks really good in terms of how it uses the `clock::v2` API. The only thing I would add is that, according to the [documentation](https://onlinedocs.microchip.com/oxy/GUID-F5813793-E016-46F5-A9E2-718D8BCED496-en-US-14/GUID-99666338-E59C-4F8B-BB43-26CBEF83A451.html), the `Pclk` is required to be at 48 MHz, so I would recommend we check for this, which can only be done at runtime. I'm not sure whether it makes sense to panic if the rate is not 48 MHz, or to have `usb::UsbBus::new` return a `Result` instead. This should probably involve a `UsbError` enum similar to `adc::BuilderError` for example.

rnd-ash:
@supersimple33 @kyp44 the USB Clock check for 48Mhz is something i opened in #884 . So maybe look at that PR for inspiration should it not get merged

kyp44:
@supersimple33 I would recommended checking the clock rate and returning a `Result` as in PR #884, since it was decided there that that was desired when migrating to `clock::v2`. However I would rename the `UsbBusErr` enum maybe to `UsbBusError`, noting that it is also private in the PR and it needs to be public. Be sure to also grab the updated documentation for `UsbBus::new` mentioning the change.

boards/grand_central_m4/src/lib.rs
```diff
@@ -595,18 +609,22 @@ pub fn uart(
...
-    use pac::gclk::{genctrl::Srcselect, pchctrl::Genselect};
+    dfll: Enabled<Dfll, U1>,
+    gclk_token: GclkToken<G>,
+    pclk_token: PclkToken<Usb>,
+    ahb_clk: AhbClk<clock::types::Usb>,
+    apb_clk: ApbClk<clock::types::Usb>,
+    dm: impl AnyPin<Id = PA24>,
+    dp: impl AnyPin<Id = PA25>,
+) -> UsbBusAllocator<UsbBus<G>> {
+    // SWITCH DFLL INTO USB CLOCK RECOVERY MODE - THIS ALLOWS FOR MORE ACCURATE USB
```
ianrrees: I'd suggest moving the comment about clock recovery mode in to the docstring for the method, so that it shows up in docs (or Rust Analyzer etc).  Also, I'm pretty sure it's required to use USBCRM if running USB with the DFLL clock source, so not sure "more accurate" is the right way to describe it.

kyp44:
A number of the USB examples for BSPs or their examples are now broken. This can be checked locally by running the `build-all.py` script in the root. I'm not sure what the policy is, but I'm guessing this PR should update them all to use the new API wherever possible. Partial v2 implementation if other v1-only peripherals are also used in examples may make this difficult or impossible. Not sure how to handle these cases.

ianrrees:
Yes, the examples need to be updated for the T1 BSPs, but I don't think we need to keep them all building for every commit in the history.  IOW, if some examples need peripheral A and B both updated to clock v2, we could leave the examples broken in the peripheral A PR, base the peripheral B PR on to the A branch, and fix the examples there.

The USB examples are problematic in their use of static mut references as well; we should fix that independently of the clock v2 work.  It would probably be a little easier to do that first, rebasing this on top of the static mut ref fix.

kyp44:
@ianrrees How can you base a PR on another branch since the branch only exists in the forked repository? Asking because I could base my PR #925 on this branch instead so that I can fix the `nvm_dsu` example in the `feather_m4` BSP in my PR.

ianrrees:
@kyp44 I'd suggest that while working the Git stuff out, you keep one copy of your work based off master as usual, so you're a bit freer to experiment with other copies.  To make the backup, with your branch checked out just do:

`git branch backup-mybranch`

You'll then need to make your local repo aware of the repo that has the work you want to rebase yours on to.  In this case, I guess that'll be supersimple33's GH - add that as a remote like:

```
git remote add supersimple33 git@github.com:supersimple33/atsamd.git
git fetch supersimple33
```

Finally, you can rebase your work on to a commit from that branch.  Keep in mind that in Git, a branch is just a named pointer to a commit (if you haven't come across [Git Internals](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects), I feel like it's well worth the time to read it closely and work through the examples).

`git rebase supersimple33/UsbV2`

kyp44:
@ianrrees But then my PR will include changes from the other PR as well. Is that not a problem? Does it mean that the other PR has to be merged first?

ianrrees:
@kyp44 You're right that if one PRed branch (for feature A) is based on another PRed branch (for feature B) that the first PR will look a lot bigger (the diff between master and A+B).  I don't think that's inherently a problem, but it does require a bit more Git-fu to work through.  I guess there are a few related things going on here:

If you're wanting a context that has both sets of changes, e.g. to update examples, then you could use this approach to update the examples but PR just your set of changes.  This would involve maintaining two branches for `A` and `A+B`.  `A` would be PRed, and the examples in it would not be expected to build (until `B` has been merged in to `master` and `A` is rebased on to it).  In `A+B` you could get the examples to build, then use eg cherry-pick to bring that work in to `A`.

There are multiple approaches we can consider to manage the PRs that will be involved in the clock v2 migration.  One option is to proceed as usual, with each PR being a self-contained change to `master` and updating the T1 BSP examples.  Another is to have a `clockv2` branch that we PR changes to for each peripheral or whatever, and maybe in that we don't worry about keeping the examples in good shape as we go.  Then, once that work is done, we just update master to that.  Personally, I think that there's maybe a little more overall work in going with the first approach, but it's more robust and a safer option given our intermittent maintenance resources.  Longer-lived feature branches (like `clockv2` might be) can take on a life of their own, as work in `master` will continue.  If we had a large amount of code that really couldn't be updated piecemeal, that would push toward a `clockv2` branch, but I don't think the examples are really at that level so I'm inclined to stick with basically the normal approach of updating one peripheral at a time in `master`.

Third is the question of updating peripherals to `clockv2` when they also need to keep working with eg thumbv6.  I think the answer here is as you've suggested: support `clockv2` on thumbv6 before working on the peripherals that need to work on both architectures.

kyp44:
Gotcha, I think for the examples going forward (including the one discussed above for the DSU/USB migrations) I'll just make a note in the breaking PR and make a note to fix it once both have been merged. I think it would have been good to have done a `clockv2` branch if it was done from the beginning as this would have avoided the current problematic state of partial support, and the full migration of the entire HAL could have been merged into master at once. At this point though, a number of peripheral migrations have already been merged to master, so probably not worth trying to do that now. Hindsight is 20/20.

I agree that it's it's worth waiting for the `thumbv6` v2 port before tackling the affected peripherals instead of the half-migration done by the `Adc`. I'll note this in the tracking issue and mark the applicable peripherals. In the meantime, I'm going to continue working on the `thumbv7` only peripherals, noting example breakages as I go, which I'll also note in the tracking issue.

rnd-ash:
Is there any update on this PR? - If not I might take ownership of it to move USB to clock V2

supersimple33:
@rnd-ash  Yeah I was waiting for thumbv6 before continuing. I just saw the talk on #912 I'll update the target branch and merge this into new v2 dev branch later tonight

boards/grand_central_m4/src/lib.rs
supersimple33: got it

boards/grand_central_m4/src/lib.rs
```diff
@@ -595,18 +609,22 @@ pub fn uart(
 
 #[cfg(feature = "usb")]
 /// Convenience function for setting up USB
-pub fn usb_allocator(
+/// Switches the DFLL into USB clock recovery mode
+pub fn usb_allocator<G: GclkId>(
```
rnd-ash: I think here, the arguments are way too complex. We should just reduce this down to Pclk,Ahb,Apb, and check for 48Mhz at runtime. 

The dfll fromusb is optional for the USB system (If the user has a more accurate clock  source on their board). So we shouldn't enforce it, but we could add documentation to state that it can lead to better results if USB appears unstable.

supersimple33: For checking for 48mhz at runtime what do you think is the best way to go about that.

supersimple33: Should I turn the return type into `Result`? like @kyp44 suggested. 

*wait I just looked at #884 I'll just follow that implementation

supersimple33:
@rnd-ash If I just take in pclk ahb apb then the `usb_allocator` function becomes just a feed forward: 

```
#[cfg(feature = "usb")]
/// Convenience function for setting up USB
pub fn usb_allocator<G: GclkId>(
    usb: pac::Usb,
    pclk_usb: Pclk<clock::types::Usb, G>,
    ahb_clk: AhbClk<clock::types::Usb>,
    apb_clk: ApbClk<clock::types::Usb>,
    dm: impl AnyPin<Id = PA24>,
    dp: impl AnyPin<Id = PA25>,
) -> UsbBusAllocator<UsbBus<G>> {
    return UsbBusAllocator::new(UsbBus::new(pclk_usb, ahb_clk, apb_clk, dm, dp, usb).unwrap());
}
```

Given this fact I feel like it's kinda redundant to even have the function then. Should I just remove it? Or we could repurpose it to maybe take in the clock/token state at the start as I was kinda doing right now but just better and better documented.

rnd-ash:
Personally, I think that the USB allocator in the BSPs is redundant. Since there is no change board to board that requires special handling for USB. Ill see if the others want to comment on this

wt:
I definitely second the idea of removing that function.

boards/grand_central_m4/examples/usb_serial.rs
```diff
@@ -14,67 +14,63 @@
...
-use hal::time::Hertz;
-use hal::timer::TimerCounter;
-use hal::timer_traits::InterruptDrivenTimer;
-use hal::usb::UsbBus;
-use smart_leds::{colors, hsv::RGB8, SmartLedsWrite};
 use usb_device::bus::UsbBusAllocator;
 use usb_device::prelude::*;
 use usbd_serial::{SerialPort, USB_CLASS_CDC};
-use ws2812_timer_delay as ws2812;
+
```
wt: Is there a real reason to separate the uses? If not please remove this line.

boards/grand_central_m4/examples/usb_serial.rs
```diff
@@ -14,67 +14,63 @@
...
         USB_ALLOCATOR.as_ref().unwrap()
     };
 
     unsafe {
-        USB_SERIAL = Some(SerialPort::new(&bus_allocator));
+        USB_SERIAL = Some(SerialPort::new(bus_allocator));
         USB_BUS = Some(
-            UsbDeviceBuilder::new(&bus_allocator, UsbVidPid(0x16c0, 0x27dd))
-                .strings(&[StringDescriptors::new(LangID::EN_US)
+            UsbDeviceBuilder::new(bus_allocator, UsbVidPid(0x2222, 0x3333))
```
wt: Where does the different vid/pid come from?

boards/grand_central_m4/examples/usb_serial.rs
```diff
@@ -85,66 +81,66 @@ fn main() -> ! {
...
+        // Turn off interrupts so we don't fight with the interrupt
+        cortex_m::interrupt::free(|_| unsafe {
+            if USB_BUS.as_mut().is_some() {
+                if let Some(serial) = USB_SERIAL.as_mut() {
+                    let _ = serial.write("Hello USB\n".as_bytes());
+                }
+            }
         });
-        if let Some(color) = pending {
-            let _ = neopixel.write((0..5).map(|_| color));
```
wt: I guess this is the same as many other usb serial examples, but it's such a shame to lose the neopixel part of the example.

boards/grand_central_m4/examples/usb_serial.rs
```diff
@@ -85,66 +81,66 @@ fn main() -> ! {
...
     }
 }
 
-static mut USB_ALLOCATOR: Option<UsbBusAllocator<UsbBus>> = None;
-static mut USB_BUS: Option<UsbDevice<UsbBus>> = None;
-static mut USB_SERIAL: Option<SerialPort<UsbBus>> = None;
-static mut PENDING_COLOR: Option<RGB8> = None;
+static mut USB_ALLOCATOR: Option<UsbBusAllocator<UsbBus<Gclk1Id>>> = None;
+static mut USB_BUS: Option<UsbDevice<UsbBus<Gclk1Id>>> = None;
+static mut USB_SERIAL: Option<SerialPort<UsbBus<Gclk1Id>>> = None;
```
wt: I wonder if a LazyCell could be used to improve the look of this part. This is not something that I would recommend for this change, just something to think about for the future.

boards/grand_central_m4/examples/usb_serial.rs
```diff
@@ -14,67 +14,63 @@
...
+            buses.apb.enable(tokens.apbs.usb),
             pins.usb_dm,
             pins.usb_dp,
         ));
         USB_ALLOCATOR.as_ref().unwrap()
     };
 
     unsafe {
-        USB_SERIAL = Some(SerialPort::new(&bus_allocator));
+        USB_SERIAL = Some(SerialPort::new(bus_allocator));
```
wt: This appears to be a safe function. If so, please move it outside of the unsafe block.

hal/src/peripherals/usb/d5x/bus.rs
```diff
@@ -543,15 +583,38 @@ impl UsbBus {
...
+            ahb_clk,
+            apb_clk,
         };
 
         Self {
             inner: Mutex::new(RefCell::new(inner)),
         }
     }
+
+    pub fn free(
```
wt: Should this api be renamed to `into_inner` to follow other rust api patterns. E.g. https://doc.rust-lang.org/stable/core/cell/struct.LazyCell.html#method.into_inner

boards/grand_central_m4/Cargo.toml
```diff
@@ -17,6 +17,7 @@ optional = true
 [dependencies.atsamd-hal]
 default-features = false
 version = "0.22.0"
+path = "../../hal"
```
wt: Is the intention to move this to a tier 1 bsp crate? I think the paths are reserved for that purpose.

FWIW, I see the changes to the hal and get the need. I just wanted to call this out.

wt (commented):
It is my understanding that changes should avoid touching multiple crates. Also, I see a bunch of different themes, like reformatting, hal changes, board changes, etc. Is it possible to break these different concerns into smaller changes that are a bit more isolated?

I am just a rando, so I hope my feedback makes sense.

boards/grand_central_m4/examples/usb_serial.rs
supersimple33: wait i'll keep it. Was just doing some testing with the metro m4 example

boards/grand_central_m4/Cargo.toml
supersimple33: I currently have this just to stay up to date with the clockv2 changes but would switch back before merging to the main

boards/grand_central_m4/examples/usb_serial.rs
supersimple33: Got it

boards/grand_central_m4/examples/usb_serial.rs
supersimple33: ok

hal/src/peripherals/usb/d5x/bus.rs
supersimple33: yeah thats probably a better name you're right

supersimple33:
This will now handle #884

boards/grand_central_m4/examples/usb_serial.rs
supersimple33: the metro m4 example, shouldn't matter

boards/grand_central_m4/src/lib.rs
supersimple33: ok this is done

supersimple33:
@ianrrees @wt @rnd-ash @kyp44 Flagging for you guys. I wanted to get a review on my PR at this point for modifications to the grand central m4 before I end up applying it to all the other boards. Let me know your thoughts. I can confirm the example works on my GC m4!

boards/grand_central_m4/examples/usb_serial.rs
```diff
@@ -85,66 +90,66 @@ fn main() -> ! {
...
 
 #[interrupt]
-fn USB_TRCPT0() {
+fn USB_TRCPT1() {
     poll_usb();
 }
 
 #[interrupt]
-fn USB_TRCPT1() {
+fn USB_SOF_HSOF() {
```
rnd-ash: I don't think we need an interrupt for HSOF. This is just timing for the USB system, no data should come in here (I think)

hal/src/peripherals/usb/d5x/bus.rs
```diff
@@ -524,17 +533,49 @@ impl Inner {
...
+        ahb_clk: AhbClk<clock::types::Usb>,
+        apb_clk: ApbClk<clock::types::Usb>,
         dm_pad: impl AnyPin<Id = PA24>,
         dp_pad: impl AnyPin<Id = PA25>,
         _usb: Usb,
-    ) -> Self {
-        mclk.ahbmask().modify(|_, w| w.usb_().set_bit());
-        mclk.apbbmask().modify(|_, w| w.usb_().set_bit());
+    ) -> Result<Self, UsbBusErr> {
+        if pclk.freq().to_Hz() != 48_000_000 {
```
rnd-ash: Maybe here we can use `48.Mhz()`? - Might be more readable than just the number in Hz

rnd-ash:
@supersimple33 Just finished a review for you. I only have a couple minor points, otherwise, looks good!

hal/src/peripherals/usb/d5x/bus.rs
supersimple33: Yeah good point

boards/grand_central_m4/examples/usb_serial.rs
supersimple33: @rnd-ash I grabbed it from the metro m4 example where they included the interrupt. Since this example is standardized across most board should I just remove it going forward?

rnd-ash:
Awesome. @supersimple33 did you want to convert this PR now from a draft?

boards/grand_central_m4/examples/usb_serial.rs
wt: Perhaps with should get an assigned one instead of squatting on another vendor's vid/pid.

Apparently, there is a project to get them for free to open source projects here: https://pid.codes/

supersimple33: I guess I'll use 0x1209, 0x0001 for now until we get a better PID.

boards/grand_central_m4/examples/usb_serial.rs
supersimple33: Ok I added it back

supersimple33:
@rnd-ash Im gonna set it ready for review once I distribute these updates to all the other thumbv7 boards. That should be done within the next few days.

PR #917, opened 2025-07-06
bsp(wio_terminal): Bump version 0.8.1 -> 0.9.0

ianrrees:
I missed in the recent PR, that the crate re-exports seeed-erpc types, so the previous change was breaking.


jbeaurivage:
@ianrrees, could you please update the readme as well?

PR #920, opened 2025-07-11
Updates the AES peripheral to use the `clock::v2` API and make it safer.

kyp44:
# Summary
As part of the `clock::v2` effort tracked in Issue #912, this PR updates the `aes` to use the `clock::v2` API as well as some other refactors. Note that the AES module is only available on `thumbv7` targets so there was no need to maintain `clock::v1` compatibility for `thumbv6` targets.

The module was out of date, so some dependencies were updated, which now allows the ciphers to be used in a safe way, so this was refactored to make things safer, removing the now-unneeded `enable_unsafe_aes_newblock_cipher` feature.

See the commit messages for more details.

# Checklist


ianrrees:
Love that this both improves functionality and deletes more source code than it adds, thanks!

PR #922, opened 2025-07-13
bsp: Add defmt feature

### LINKED ISSUE #810: Provide defmt feature in BSPs to enable atsamd-hal/defmt

ianrrees:
This adds a `defmt` feature to our T1 BSPs (plus Grand Central M4, as groundwork for @supersimple33 's #902) which enables the HAL's `defmt` feature and sets the linker argument using `build.rs`.

I'm a little surprised this seems compatible with the `use_rtt` feature, inasmuch as the examples still build.  @kyp44 if you have a minute, could you please have a look and see if this looks problematic?

Resolves #810


sajattack:
Great idea with the build.rs!

ianrrees:
~~Before merging this, I'd like to add a comment about using `DEFMT_LOG` near to the `defmt` feature in `Cargo.toml`~~

kyp44:
@ianrrees is there something specific you'd like me to look at? I don't use a debugger or `defmt`, so I am not really tracking any of this stuff.

ianrrees:
> @ianrrees is there something specific you'd like me to look at? I don't use a debugger or `defmt`, so I am not really tracking any of this stuff.

Nothing specific, just noticed that git blame showed you as touching some lines to do with `use_rtt`

kyp44:
@ianrrees I did not see this in the git blame, though I have never used that command before until now. I did touch lines related to the `rtic` feature since I worked on that. in any case, I haven't done any actual work on the `use_rtt` explicitly so I'm not familiar with it, though maybe at one point I had to touch something for some other reason.

rnd-ash:
@ianrrees I can't find anything wrong with this PR that might need correcting, it all looks good. 

Potentially, yes, using RTT Serial and DEFMT could cause issues if used together, I might be able to test this later today and check

PR #923, opened 2025-07-13
Map all mkr1000 pins

ojob:
# Summary

Goal is to document all Arduino MKR1000 pins, and give hints on using them.

For instance, PWM configuration, with the TCC and Channel to use.

# Checklist

## If Adding a new Board

n/a

no new feature

### DIFF
--- a/boards/arduino_mkr1000/Cargo.toml
+++ b/boards/arduino_mkr1000/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "arduino_mkr1000"
-version = "0.5.1"
+version = "0.6.0"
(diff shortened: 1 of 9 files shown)


ojob:
Following work on #878, here's a new branch to map all pins.

Still a question on PWM, that I have to explore some more: TCCx are trivial to use, TCx not so.

ianrrees:
> Still a question on PWM, that I have to explore some more: TCCx are trivial to use, TCx not so.

@ojob - I'm going to mark this as a draft for now, but let me know when the PWM questions are resolved and this is ready to go.  Thanks!

PR #925, opened 2025-07-14
Updates the DSU peripheral to use the `clock::v2` API

kyp44:
# Summary
As part of the `clock::v2` effort tracked in Issue #912, this PR updates the `dsu` to use the `clock::v2` API by requiring ownership of its `AhbClk` and `ApbClk`. Note that this peripheral is only on `thumbv7` targets.

- `feather_m4/nvm_dsu` (`usb::UsbBus`)

# Checklist

### DIFF
--- a/hal/src/peripherals/dsu.rs
+++ b/hal/src/peripherals/dsu.rs
@@ -7,10 +7,17 @@
 
 use crate::pac::{self, Pac};
 
+type DsuAhbClk = crate::clock::v2::ahb::AhbClk<crate::clock::v2::types::Dsu>;
+type DsuApbClk = crate::clock::v2::apb::ApbClk<crate::clock::v2::types::Dsu>;
+
 /// Device Service Unit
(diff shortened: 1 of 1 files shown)


rnd-ash:
@kyp44 regarding the broken example, I believe this can be worked around as clock::V2 clocks have a .into() method to convert them into their V1 counterparts

kyp44:
@rnd-ash The conversion to the v1 types does not help here because [`usb::UsbBus::new`](https://docs.rs/atsamd-hal/latest/atsamd_hal/usb/struct.UsbBus.html#method.new) requires a `&mut Mclk`, which is not possible to pass when using v2 because the `Mclk` is consumed by `clock::v2::clock_system_at_reset` to use the v2 API.

rnd-ash:
Ah right, in this case I personally was calling unsafe to steal mclk .... But you are right, it's probably best to avoid an unsafe block in our examples

PR #927, opened 2025-07-19
Updates the ICM peripheral to use the `clock::v2` API

kyp44:
# Summary
As part of the `clock::v2` effort tracked in Issue #912, this PR updates the `icm` to use the `clock::v2` API by requiring ownership of its `AhbClk` and `ApbClk`. Note that this peripheral is only on `thumbv7` targets.

No examples on any Tier 1 BSPs use the ICM so none were affected.

# Checklist

### DIFF
--- a/hal/src/peripherals/icm.rs
+++ b/hal/src/peripherals/icm.rs
@@ -910,8 +910,8 @@ impl<I: RegionNum> Region<I> {
 pub struct Icm {
     /// ICM pac register providing hardware access
     icm: crate::pac::Icm,
-    _ahb_clk: IcmAhbClk,
-    _apb_clk: IcmApbClk,
+    ahb_clk: IcmAhbClk,
+    apb_clk: IcmApbClk,
 }
 
 impl Icm {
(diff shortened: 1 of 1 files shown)


rnd-ash:
Looks good to me! Appreciate the documentation updates as well.

hal/src/peripherals/icm.rs
```diff
@@ -877,23 +910,19 @@ impl<I: RegionNum> Region<I> {
 pub struct Icm {
     /// ICM pac register providing hardware access
     icm: crate::pac::Icm,
+    _ahb_clk: IcmAhbClk,
```
ianrrees: Let's remove the leading underscore from these, they do get used in `free()`.

hal/src/peripherals/icm.rs
```diff
@@ -877,23 +910,19 @@ impl<I: RegionNum> Region<I> {
...
-    ///
-    /// Clock::v2
-    /// `tokens.apbs.icm.enable();`
     #[inline]
-    pub fn new(icm: crate::pac::Icm) -> Self {
-        Self { icm }
+    pub fn new(icm: crate::pac::Icm, ahb_clk: IcmAhbClk, apb_clk: IcmApbClk) -> Self {
+        Self {
+            icm,
+            _ahb_clk: ahb_clk,
```
ianrrees: With the underscore removed as above, clippy will want this to look like `ahb_clk,` r/t `ahb_clk: ahb_clk,`

ianrrees (commented):
Just a minor naming suggestion, but otherwise LGTM

kyp44:
@ianrrees Removed the underscores in the latest commit, and also did so in PR #925 for the same reason.

PR #930, opened 2025-07-28
Re-add GCLK0 for CAN dependencies

rnd-ash:
# Summary

Reverts the changes done in #919, whilst also ensuring that the provided GCLK is EnabledGclk0, rather than potentially any GCLK source like it was prior.  The way GCLK0 is provided is the same as in the QSPI rework PR (#926), since both peripherals require proof of GCLK0, so both peripherals should have a consistent way of doing this. 

@glaeqen would I be able to ask for your input on this?

### DIFF
--- a/boards/atsame54_xpro/examples/mcan.rs
+++ b/boards/atsame54_xpro/examples/mcan.rs
@@ -161,8 +161,14 @@ mod app {
 
         let (pclk_can1, gclk0) = clock::pclk::Pclk::enable(tokens.pclks.can1, gclk0);
 
-        let dependencies =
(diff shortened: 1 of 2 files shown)


glaeqen:
Looks fine to me. One could bikeshed that the order of elements in the tuple does not match the one from before the revert but types have changed anyway so probably it is not the big deal.

ianrrees:
@rnd-ash - gentle reminder that we need to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).  I believe we'll also need to manually mark this as a breaking change, since it doesn't seem to detect those in feature-gated APIs.

PR #931, opened 2025-08-04
Updates the NVM peripheral to use the `clock::v2` API

kyp44:
# Summary
As part of the `clock::v2` effort tracked in Issue #912, this PR updates the `nvm` to use the `clock::v2` API by requiring ownership of its `AhbClk`s and `ApbClk`. Note that this peripheral is only on `thumbv7` targets.

The NVM has three AHB bus clocks: `CLK_NVMCTRL_AHB`, `CLK_NVMCTRL_CACHE`, and `CLK_NVMCTRL_SMEEPROM`. It seems that `CLK_NVMCTRL_AHB` is needed for basic operation, but the [documentation](https://onlinedocs.microchip.com/oxy/GUID-F5813793-E016-46F5-A9E2-718D8BCED496-en-US-14/GUID-2696AA45-C209-41AD-AF05-458B806A1A00.html) does not talk about exactly what `CLK_NVMCTRL_CACHE` and `CLK_NVMCTRL_SMEEPROM` are needed for at all. As the the `nvm::Nvm` does not seem to do anything with the cache at all, it was assumed that `AhbClk<NvmCtrlCache>` is not needed. However, it is assumed that `AhbClk<NvmCtrlSmeeProm>` is needed for SmartEEPROM functionality, and so this is now required when calling `nvm::Nvm::smart_eeprom`, and the clock can be freed with `nvm::smart_eeprom::SmartEeprom::free`.

Note that automatic wait states are enabled when `clock::v2::clock_system_at_reset` is called so that the notes [here](https://onlinedocs.microchip.com/oxy/GUID-F5813793-E016-46F5-A9E2-718D8BCED496-en-US-14/GUID-2696AA45-C209-41AD-AF05-458B806A1A00.html) about wait states should not be an issue.

- `feather_m4/nvm_dsu` (`dsu::Dsu`, `usb::UsbBus`)
- `feather_m4/smart_eeprom` (`usb::UsbBus`)

# Checklist


hal/src/peripherals/nvm/mod.rs
```diff
@@ -727,8 +746,8 @@ impl Nvm {
 
     /// Retrieve SmartEEPROM
```
rnd-ash, line 747: We should probably document why the AhbClk is required here

kyp44: Do you mean why this NVMCTRL clock is needed as opposed to the NVMCTRL_SMEEPROM (which is required to create a `nvm::smart_eeprom::SmartEeprom` via `nvm::Nvm::smart_eeprom`) or the NVMCTRL_CACHE? I don't have the answer to that, and the datasheet does not explain these other bus clocks at all.

rnd-ash: Ah, I just read the datasheet a bit more, I thought AHBClk wasn't required for Smart eeprom to function, thats all. My mistake, apologies. It is fine as it is now then

rnd-ash:
Apologies for the very long delay in reviewing this!

It looks good, however the only major drawback I see here is that having the SmartEEPROM clock passed down to the SmartEEPROM might make passing around NVM a lot harder, since SmartEEPROM takes a reference to NVM, so when SmartEEPROM is in use, you can't move or use NVM. So now you need to keep track of both the clock token and NVM when sharing it in order to make use of the SmartEEPROM.

That said, it is still a viable option, but in an RTIC app that uses NVM+SmartEEPROM for example, I can see some shared NVM resource going from this:
```
Shared {
   shared_nvm: Nvm
}
```

to 

```
Shared {
   shared_nvm: (Nvm, Option<AhbClk<NvmCtrlSmeeProm>>)
}
```

and then some redundant extra code to access the SmartEEPROM every time to pull the clock token out of the option, use it, and then place it back.

Maybe, the simpler solution would be to just put the NvmCtrlSmeeProm clock into NVM itself? - I know this would mean the ahb is enabled when maybe a user will never use the SmartEEPROM functionality....but I think makes using it a lot easier.

kyp44:
@rnd-ash I've been away from this stuff for a few months, but wanted to address your points. I see the inconvenience of having to carry around the `AhbClk<NvmCtrlSmeeProm>` so that the `SmartEeprom` can be recreated using `Nvm::smart_eeprom` whenever the `Nvm` is moved (while also needing to call `SmartEeprom::free` each time to get it back).

Certainly the simplest solution to avoid this inconvenience is to just require the `AhbClk<NvmCtrlSmeeProm>` in `Nvm::new` (and return it with `Nvm::free`), but this can be inefficient since the AHB clock would have to be enabled even if the user has no intention  of using the Smart EEPROM, as you pointed out.

While I haven't thought it through completely, it seems like this inefficiency would be eliminated if desired by using a type-level enum parameter on `Nvm`. Something like:

```rust
trait UseEeprom {}

struct NvmNoEeeprom {}
impl UseEeeprom for NvmNoEeeprom {}

struct NvmEeprom {}
impl UseEeeprom for NvmEeeprom {}

struct Nvm<E: UseEeprom> {
  ...
}
```

Then the `new` and `free` methods of `Nvm` would have different versions for the `E` parameter, which would or would not take/return the `AhbClk<NvmCtrlSmeeProm>`, and the `smart_eeprom` method would only be implemented for `Nvm<NvmEeprom>`.

PR #932, opened 2025-08-08
Updates the DMAC peripheral to use the `clock::v2` API

kyp44:
# Summary

As part of the `clock::v2` effort tracked in Issue #912, this PR updates the `dmac` to use the `clock::v2` API by requiring ownership of its `AhbClk` for `thumbv7` targets.

See the commit message for details.

The following Tier 1 BSP examples were also updated to include the change:
- `samd11_bare/i2c`
- `feather_m0/async_dmac`
- `feather_m0/async_i2c`
- `feather_m0/async_spi`
- `feather_m0/async_uart`
- `feather_m0/dmac`
- `feather_m0/i2c`
- `feather_m0/spi`
- `feather_m0/uart_dma_blocking`
- `feather_m0/uart_dma_nonblocking`
- `feather_m4/dmac`
- `metro_m0/i2c`
- `metro_m0/spi`
- `metro_m4/async_dmac`

- `feather_m4/i2c` (`sercom::i2c::Config`)
- `feather_m4/spi` (`sercom::spi::Config`)
- `feather_m4/uart_dma_blocking` (`sercom::uart::Config`)
- `feather_m4/uart_dma_nonblocking` (`sercom::uart::Config`)
- `metro_m4/async_i2c` (`sercom::i2c::Config`)
- `metro_m4/async_spi` (`sercom::spi::Config`)
- `metro_m4/async_uart` (`sercom::uart::Config`)
- `metro_m4/i2c` (`sercom::i2c::Config`)
- `metro_m4/spi` (`sercom::spi::Config`)

# Checklist


hal/src/dmac/dma_controller.rs
```diff
@@ -150,28 +152,18 @@ macro_rules! define_channels_struct_future {
 with_num_channels!(define_channels_struct_future);
 
 /// Initialized DMA Controller
+#[hal_macro_helper]
```
rnd-ash, line 155: We should probably decide if in the HAL in general, it makes more sense to use this macro to feature gate D5x/D21 clocking in the same peripheral structure, or, better to do it like in other peripherals, where D21 and D5x have completely different structures (Look at the ADC for example)

rnd-ash:
Apart from the discussion question, I'm happy for this to be merged into our Clock V2 branch, then we can fix the examples once the other peripherals are migrated

kyp44:
A merge is good with me. I don't think we need to track the broken examples in the branch as we go. I have [a script](https://github.com/kyp44/atsamd-tools/blob/master/compile-bsp-examples.py) to go build all the Tier 1 BSPs so they will be easy to assess the broken examples at any point, plus we can refer back to the PRs to know when requisite peripherals have been merged.

rnd-ash:
Exactly! The examples will be fixed as we go :)

hal/src/dmac/dma_controller.rs
kyp44: Taking a look, it looks like the ADC is different enough to justify this, but I worry that many peripherals could potential have a lot of duplicated code if this is done. This is the whole reason why many of the peripherals don't make sense to migrate until the `thumbv6` clock v2 API is done, see note 5 in the Migration Guidelines of #912. As disclaimer, I have not thought very deeply about how much code duplication might be able to be avoided by taking this approach with more similar peripherals.

EDIT: Additionally the general differences between the structs, namely the `thumbv7` version needing to own an `AhbClk`, for example while the `thumbv6` version does not, will likely go away for a lot of peripherals once the `thumbv6` clock v2 API is done since then they will both need to own these clock proofs.

PR #938, opened 2025-09-18
fix(dmac): DMAC handler now waits for channel to report as disabled

ritikmishra:
# Summary

I was running into an issue with using multiple SPI+DMA transactions in a very tight control loop: sometimes the `spi.transfer` would hang indefinitely because the RX DMA channel was enabled, but somehow the TX DMA channel was disabled. 

I managed to figure out that this was happening because the DMAC interrupt handler does not wait for the channel to finish disabling itself, meaning that sometimes the disable doesn't finish until _after_ my loop's next SPI enable is dispatched. 

The fix is for the DMAC interrupt handler to wait for the channel to finish disabling before waking the future up again.

--

Note that `TransferFuture` is not a public item, so it should be OK for its return type to change.

# Checklist

### DIFF
--- a/hal/src/dmac/async_api.rs
+++ b/hal/src/dmac/async_api.rs
@@ -1,6 +1,7 @@
 //! APIs for async DMAC operations.
 
 use atsamd_hal_macros::hal_cfg;
+use core::sync::atomic;
 
 use crate::{
     async_hal::interrupts::{DMAC, Handler},
@@ -96,6 +97,16 @@ impl Handler<DMAC> for InterruptHandler {
                     w.enable().clear_bit();
                     w.trigsrc().variant(TriggerSource::Disable)
                 });
+
+                while dmac.channel(channel).chctrla().read().enable().bit_is_set() {
+                    core::hint::spin_loop();
+                }
+
+                // Prevent the compiler from re-ordering read/write
+                // operations beyond this fence.
+                // (see https://docs.rust-embedded.org/embedonomicon/dma.html#compiler-misoptimizations)
+                atomic::fence(atomic::Ordering::Acquire); // ▼
+
                 WAKERS[channel].wake();
             }
         }
--- a/hal/src/dmac/channel/mod.rs
+++ b/hal/src/dmac/channel/mod.rs
@@ -719,17 +719,13 @@ impl<Id: ChId> Channel<Id, ReadyFuture> {
 
         self.configure_trigger(trig_src, trig_act);
 
-        transfer_future::TransferFuture::new(self, trig_src).await;
+        transfer_future::TransferFuture::new(self, trig_src).await
 
         // No need to defensively disable channel here; it's automatically stopped when
         // TransferFuture is dropped. Even though `stop()` is implicitly called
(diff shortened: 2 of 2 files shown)


hal/src/dmac/async_api.rs
```diff
@@ -96,6 +97,16 @@ impl Handler<DMAC> for InterruptHandler {
                     w.enable().clear_bit();
                     w.trigsrc().variant(TriggerSource::Disable)
                 });
+
+                while dmac.channel(channel).chctrla().read().enable().bit_is_set() {
+                    core::hint::spin_loop();
```
bradleyharden, line 112: Is `spin_loop` necessary? I thought that bug had been fixed. Don't we use an empty `while` body in tons of other places?

hal/src/dmac/async_api.rs
```diff
@@ -96,6 +97,16 @@ impl Handler<DMAC> for InterruptHandler {
...
+                }
+
+                // Prevent the compiler from re-ordering read/write
+                // operations beyond this fence.
+                // (see https://docs.rust-embedded.org/embedonomicon/dma.html#compiler-misoptimizations)
+                atomic::fence(atomic::Ordering::Acquire); // ▼
```
bradleyharden, line 118: Is there another `atomic::fence` with `Release` ordering somewhere else? I thought fences like this were only meaningful when paired across two concurrent functions? I think it's already guaranteed by the language that the enable bit will read true before `wake()` is called. The reads are volatile, so they can't be removed, and the read is observed, so you can't reorder `wake()` before it.

jbeaurivage:
Nice, thank you! One small comment, I think this fix should also extend to D11/D21 targets. This should also take care of the clippy warning by the same occasion.

hal/src/dmac/async_api.rs
ritikmishra: there was a bug where spin_loop was necessary? it seems like sometimes there's an empty body, sometimes `spin_loop` is used, and sometimes `cortex_m::asm::nop()` is used

hal/src/dmac/async_api.rs
jbeaurivage: Yeah, this seems more or less unnecessary. I think the intention was to protect reordering reads/writes to the memory chunk the DMA is accessing, although we already have fences for that at a lower level.

That being said, we had a discussion a while ago about how `atomic::fence` isn't even enough to guarantee no reordering will take place... IIRC the verdict was that we should use an empty `asm!` black box instead.

ritikmishra: `Channel::_enable_private` has the corresponding `Release` fence + is called in the `TransferFuture::poll` implementation that this ISR interacts with

ritikmishra:
oh good catch @jbeaurivage , added in 2d466f1e1099142b458c8319318db1be2161fcd4

hal/src/dmac/async_api.rs
jbeaurivage: Indeed maybe we should standardize at some point, but it's not a source of bugs or anything so I'm not super concerned about it for now

PR #941, opened 2025-10-03
Makes USB examples in Tier 1 BSPs safe and suppresses compiler warnings

kyp44:
# Summary

NOTE: Marking as draft because this depends on a merge of PR #935.

A number of Tier 1 BSP examples that use USB as a serial device emit compiler warnings related to the necessary use of `static mut`. Additionally, some are not used in a safe way.

This updates these examples to suppress the warnings and make things safe by accessing the USB device in the main program (i.e. not in the USB polling interrupt) only in crticial sections that disable applicable USB interrupts.

The usability issue with USB in embedded applications is a [known](https://github.com/rust-embedded-community/usb-device/issues/165) [issue](https://github.com/rust-embedded-community/usb-device/issues/9), and the best solution for these examples was discussed in Matrix.

See the commit log for details about which BSP examples were changed, and not that each BSP crate was changed in a separate commit per the guideline.

# Checklist


rnd-ash:
I had a look at this, overall quite happy with things,

However, I think the `usb_get` and `usb_free` methods that are used are a little bit over complex for an example (Especially making the user have to manually create a critical section).

In some other projects I have done with USB, I have gotten around the static mut like so


Incomplete example:
```rust

pub static USB_BUS_ALLOC: StaticCell<UsbBusAllocator<UsbBus>> = StaticCell::new();

// Data to be accessed whenever an ISR occurs on USB
pub struct UsbIsrInfo<'a> {
    pub dev: UsbDevice<'a, UsbBus>,
    pub class: SomeClassLikeSerial<'a, UsbBus>,
}

pub static USB: Mutex<RefCell<Option<UsbIsrInfo<'static>>>> = Mutex::new(RefCell::new(None));

#[inline(always)]
pub fn with_cs_mutex<R, T, F: FnOnce(&CriticalSection, &mut T) -> R>(
    m: &Mutex<RefCell<Option<T>>>,
    f: F,
) -> Option<R> {
    free(|cs| m.borrow(cs).borrow_mut().as_mut().map(|x| f(cs, x)))
}

/// Example usage for ISR:

#[interrupt]
fn USB_OTHER() {
    poll_usb();
}

#[interrupt]
fn USB_TRCPT0() {
    poll_usb();
}

#[interrupt]
fn USB_TRCPT1() {
    poll_usb();
}

fn poll_usb() {
    free(|cs| {
        if let Some(inf) = USB.borrow(cs).borrow_mut().as_mut() {
            if inf.dev.poll(&mut [&mut inf.can_usb]) {
                inf.class.poll(cs);
            }
        }
    });
}


```

PR #944, opened 2025-10-18
Fix and improve EIC Event control

rnd-ash:
# Summary
The EIC Evctrl register works differently to intset/intclear. There is just one register, and the user must set the bit for each channel in it according to if the event system should be enabled or not per channel. 

Before, this was done by writing to the register directly, rather than ORing it with whats already in it. This would cause the last value of the register to be overwritten, meaning, it would only ever be possible to have 1 event system channel output enabled at once.

For completion, I've also added a `disable_event` method, which turns off a channels Event system output.

# Checklist

### DIFF
--- a/hal/src/peripherals/eic/d11/pin.rs
+++ b/hal/src/peripherals/eic/d11/pin.rs
@@ -62,6 +62,15 @@ where
                 .modify(|r, w| unsafe { w.bits(r.bits() | 1 << P::ChId::ID) });
         });
     }
+
+    /// Disables the event output of the channel for the event system.
+    ///
+    /// Note that whilst this function is executed, the EIC peripheral is disabled
+    /// in order to write to the evctrl register
+    pub fn disable_event(&mut self) {
+        self.chan.with_disable(|e| {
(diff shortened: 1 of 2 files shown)


sajattack:
https://rust-training.ferrous-systems.com/latest/book/pac-svd2rust#the-svd2rust-generated-api-2

> Any un-written Fields are left alone

My understanding was it sort of did an or by default? But I could be out of date or totally wrong.

rnd-ash:
The problem here is that all the EIC channels share 1 field with this register 

Interestingly, on the D21/D11 version of the evctrl register, each channel has its own field, bit, there is no iterator method so it's impossible to do channel(n).set_bit, so for consistency, I have both versions write directly to the register using bit operations.


## SAM5x
image

## SAMD2/1 series
image

Looking at the samd21 PAC (https://atsamd-rs.github.io/atsamd/samd21g/thumbv6m-none-eabi/doc/atsamd21g/eic/evctrl/type.W.html), there is no iterator, so one would have to manually select which extinteo field in a switch statement based on ChId.

Issue #948, opened 2025-10-30
Pwm: TC/TCC period off-by-one


codepainters:
While experimenting with `ATSAMD21G18A` board I've noticed that TC and TCC period is not what I'd expect, even though the frequency I wanted to obtain is lower than expected, even though  the theoretical period is an integer number of cycles.

I believe the problem is here:

https://github.com/atsamd-rs/atsamd/blob/2d890f0167cf325dc1361b413a150c21d97c8bde/hal/src/peripherals/pwm/d11.rs#L60-L82

* `TimerParams::new()` called in line 62 correctly calculates the number of ticks by dividing the frequencies
* code in line 81 applies the calculated period:
  ```rs
     count.cc(0).write(|w| unsafe { w.cc().bits(params.cycles as u16) }); 
   ```

However, looking at the "SAM D21 Family Data Sheet" section 30.6.1, definition of `TOP`:
> The counter reaches TOP when it becomes equal to the highest value in
the count sequence. The TOP value can be the same as Period (PER)
or the Compare Channel 0 (CC0) register value depending on the
waveform generator mode in Waveform Output Operations.

My understanding is that the timer counts from 0 to `CC0` value (inclusive), so in fact it goes through `n + 1` distinct values. To generate `1/n` frequency, the value loaded to `CC0` should be equal to `n - 1`, and not `n`.

Changing the line 81 (and similar code in line 49) to
```rs
count.cc(0).write(|w| unsafe { w.cc().bits(params.cycles as u16 - 1) });
```
solves the  problem for me (verified with a frequency counter).

The code for TCC updating `PER` register suffers from the same issue.

codepainters:
I've just added a PR fixing the period setting.

Perhaps it might be a good idea to disallow `params.cycles` equal 0, as subtracting 1 from it doesn't make any sense.  
Currently `check_cycles_u16()` panics, if `cycles` value exceeds `u16`, perhaps it should also panic on `cycles == 0`?

Also, I only modified `hal/src/peripherals/pwm/d11.rs`, I guess that `d5x.rs` might require a similar fix (I can update the PR).

codepainters:
Checked the ATSAMD51 datasheet and it seems exactly the same (no surprise):

<img width="1098" height="290" alt="Image" src="https://github.com/user-attachments/assets/65971781-4ab8-4df3-9526-ac984ee32cb7" />

and

[...]

PR #949, opened 2025-10-30
samd11/21: fix off-by-one Pwm period

codepainters:
# Summary

This change fixes the `Pwm` period off-by-one issue on SAMD11/21, as described in #948.

### DIFF
--- a/hal/src/peripherals/pwm/d11.rs
+++ b/hal/src/peripherals/pwm/d11.rs
@@ -46,7 +46,7 @@ impl $TYPE {
             }
         });
         count.ctrla().write(|w| w.wavegen().mpwm());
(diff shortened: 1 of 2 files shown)


rnd-ash:
Good catch, I'd be happy for this to be merged, although, I'd like to see first if the D5x family is affected by the same bug

codepainters:
I can do it, but I'm not sure I have any chance to verify with real hardware, I don't think I have any D5x board in my collection. I need to check that.

codepainters:
PR updated, see #948 issue for details.

rnd-ash:
Just tested it, works well on my SAME54 setup, I'd be happy to have this merged!

Issue #915, opened 2025-07-01
PIC32C series discussion thread


rnd-ash:
# Summary
This issue is to track and discuss the potential of adding support the PIC32C series to the crate.

The PIC32C series is Microchips successor to the ATSAM series of processors.

The PIC32C can be broken down into the following product lines:

* PIC32CZ CA - Cortex M7 - Meant to replace the SAMx7x series of chips (Inclusion of 5V tolerant digital IO)
* PIC32CK - Cortex M33 - Next generation of the PIC32CX SG line (See below)
* PIC32CX SG - Cortex M4F - Direct replacement to SAMx5x line (Same peripherals as ATSAMx5x). Binaries compiled for the SAMx5x series of chips does run on the PIC32CX SG.
* PIC32CM JH - Cortex M0+ - Direct upgrade to the SAM C21 series of chips - Pin compatible with SAMC21 series.
* PIC32CM LX - Cortex M23 - Upgrade to SAML10/11 series
* PIC32CM MC - Cortex M0+ - Feature compatible with SAM C20 chips
* PIC32CM GV00 - Cortex M0+ - SAMD21 replacement

## Compatibility

From comparing datasheets, it appears Microchip was nice enough to either copy the previous generation SAMx peripheral sets, or, create 1 version of a new peripheral and use it across the lineup, meaning worst case, only 1 extra peripheral implementation would be needed for each peripheral (Like we have D5x and D11 implementations already in the crate).

As a glance, it appears the peripherals are split as follows (Microchip new denotes the new peripheral set that microchip has created, rather than peripheral sets present in the previous SAMx lineup, denoted as D5x or D11):

|Chip|Peripheral impl|Migration guide from SAMx chip|
|:-:|:-:|:-:|
|PIC32CZ CA|Microchip new|[From SAMx7x](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ApplicationNotes/ApplicationNotes/AN5816-Migration-Guide-from-SAM-E70-S70-V70-V71-to-PIC32CZ-CA70-MC70-Microcontroller-Units-DS00005816.pdf)|
|PIC32CK|Microchip new|[From SAMD/E5x](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ApplicationNotes/ApplicationNotes/AN5893-Migration-Guide-from-SAM-D5x-E5x-to-PIC32CK-SG01-SG00-GC01-GC00-DS00005893.pdf)|
[...]

PR #950, opened 2025-11-11
Add support for PIC32CX SG series

rnd-ash:
## Summary
As mentioned in #915, this specific PIC32C series is essentially a rebrand of SAME54, with the optional integration of a HSM Security module, and an extra register for NVM that allows for immutable boot.

I have used Microchips latest SVDs to generate the following new PACs, and applied a couple of XSL patches where necessary:

SG41 series - Immutable boot only

* `PIC32CXSG41064`
* `PIC32CXSG41080`
* `PIC32CXSG41100`
* `PIC32CXSG41128`

SG60 series - HSM Module only

* `PIC32CXSG60100`
* `PIC32CXSG60128`

SG61 series - Immutable boot and HSM

* `PIC32CXSG61100`
* `PIC32CXSG61128`

There is also now an additional family added to the devices.yaml file called `pic32cxsg` - This is used mostly with `hal_cfg(any(`d5x`, `pic32cxsg`))` in the crate as most of the peripheral logic is the same between both, but there is the occasional difference, so this lets us better control logic where necessary (For example, on the ADC, TSEN temperature sensor is deprecated on the PIC32CXSG.

 ### Things to finalize

* [ ] - HSM pad configuration (The HSM is hard wired to Sercom2 - So Sercom2 on chips with HSM should not allow any GPIOs to be configured for it (Maybe a special pads group will be the best solution?). And restrict Sercom2 to SPI only
* [ ] - Immutable boot logic - Add functions to NVM to allow the user to configure and check the status of immutable boot


hal/Cargo.toml
```diff
@@ -31,45 +31,51 @@ features = ["samd21g", "samd21g-rt", "usb", "dma", "async", "rtic"]
 
 [dependencies]
 aes = "0.8.4"
-atsamd-hal-macros = {version = "0.3.0", path = "../atsamd-hal-macros"}
```
ianrrees, line 34: The formatting changes make this PR hard to review - if they're necessary then please open a separate PR for them

README.md
```diff
@@ -1,6 +1,6 @@
 # atsamd & atsame support for Rust
 
-This repository holds various crates that support/enable working with Microchip (nee Atmel) `samd11`, `samd21`, `samd51` and `same5x` based devices using Rust.
+This repository holds various crates that support/enable working with Microchip (nee Atmel) `samd11`, `samd21`, `samd51` and `same5x` based devices using Rust, and also the newer Microchip `PIC32CX SG` series
```
ianrrees, line 3: This language makes the PIC32CX SG feel really "bolted on", maybe something like "working with Microchip SAMD family of devices including SAMD11, SAMD21...PIC32CX SG."

hal/src/sercom/pad/impl_pad_thumbv7em.rs
```diff
@@ -429,28 +429,28 @@ pad_table!(
...
-        #[hal_cfg("sercom4")]
+        #[hal_cfg(all("sercom4", not("hsm")))]
         D: (Sercom4, Pad2, IoSet4),
     }
     #[hal_cfg("pb29")]
     PB29 {
         #[hal_cfg("sercom2")]
         C: (Sercom2, Pad3, IoSet2),
-        #[hal_cfg("sercom4")]
+        #[hal_cfg(all("sercom4", not("hsm")))]
```
ianrrees, line 491: It's not immediately clear to me what this code actually means - if the HSM is present, does the chip have SERCOM4?

ianrrees (commented):
I feel like the value/cost this PR provides is not what the project needs at this time - the value seems to essentially be that it gives compile errors if a chip with HSM is targeted and the wrong SERCOM is used(?), but the cost is an increased support burden.  If a person wants to make some firmware for these PIC32CS SG chips - can't they simply tell this crate to target the equivalent SAM chip, and avoid using the wrong SERCOM?

At a high level, our issue is that maintainers are overextended; we have foundational work (#892 in particular) that is bitrotting, and bugs that need fixing

hal/src/sercom/pad/impl_pad_thumbv7em.rs
rnd-ash: Chips which contain the HSM module have these IOs hard wired to sercom 2 , thus cannot be used by any other peripheral (and the IOs are not even connected to the legs of the chip, so are not externally exposed)

rnd-ash:
Originally, this is what I wanted to do, just add a note to the readme, but then noticed this chip series has a new 80 pin variant that SAM5x doesn't have, and also slightly different peripheral counts (all chips have ethernet+2x can for example), so I then added the definitions to the devices yaml

The stuff about HSM though I am on the fence about. It would make sense to these extra compile error checks to avoid someone doing something wrong, but equally, it may be easier to just add a note somewhere documenting it (as the chips that have HSM are about 4x as expensive, it may be fair to assume a user has one for the specific reason to use the HSM, so they know what they are doing)

README.md
rnd-ash: would 

"This repository holds various crates that support/enable working with Atmel (Now Microchip) devices (`samd11`, `samd21`, `samd51` and `same5x`), as well as Microchip's own `PIC32CX SG` series using Rust "

sound better then?

(We could change this into a list, to better future proof this if we eventually get around to 1 day adding the next generation PIC32C series of ICs)

rnd-ash:
Probe-rs PR is merged, converting this from a draft

PR #952, opened 2025-11-12
fix: Update feather_m0 usb-device crate.

wt:
This crate needs to be the same as the atsamd-hal, which uses version 0.3.2.

# Summary
I updated the usb-device dependency of the feather_m0 support crate to match the version used in atsamd-hal.

# Checklist

### DIFF
--- a/boards/feather_m0/Cargo.toml
+++ b/boards/feather_m0/Cargo.toml
@@ -35,7 +35,7 @@ version = "2.1.1"
 
 [dependencies.usb-device]
 optional = true
-version = "0.3.1"
+version = "0.3.2"
 
 [dependencies.embedded-sdmmc]
 optional = true


rnd-ash:
Why would this  be necessary? Cargo should auto-pull the latest patch version of any dependency...or did USB-Device break the API in a patch version?

jbeaurivage:
> Why would this be necessary? Cargo should auto-pull the latest patch version of any dependency...or did USB-Device break the API in a patch version?

Yeah, I have this question as well

wt:
When I was building stuff I saw warnings about different versions of deps in the dependency tree. It might be better to make the dev deps for the examples use a * version for most of their deps, I think they would then just match whatever is used in other parts of the dep tree.

Again, I was just trying to get rid of the warning. I know rust will not see the data structures from different versions of crates as different. I just figured that probably wasn't good for binary size.

rnd-ash:
@wt if you run `cargo update` from within your project, it should pull the latest patch versions of every dependency in  the tree.

Out of interest,  what kind of warnings do you get? - I haven't seen warnings before related to different crate versions...

wt:
I was matching the dep from the hal crate's Cargo.toml:

```
usb-device = {version = "0.3.2", optional = true}
```

jbeaurivage:
Just checking whether this is still an issue. I don't think this change is necessary, unless `usb-device` somehow broke semver between 0.3.1 and 0.3.2, the Cargo resolver will deduplicate and select the most appropriate dependency version.

jbeaurivage:
Will close this for now, but feel free to reopen or discuss further if it causes issue.

Issue #951, opened 2025-11-12
circuit-playground-express crate has very old dependencies


wt:
I was playing around with upgrading the circuit-playground-express. I didn't get very far with it before I gave up for the time being. I have posted a branch at https://github.com/wt/atsamd/tree/upgrade_cpe to share what I did.

I am running into an issue where I cannot seem into a compile error with the i2c support. I was hoping someone with a little more knowledge could take a look.

I was just trying to build the blinky example with this command:

    cargo build --example blinky_basic

I get a compiler error that I am not sure how to fix.

wt:
I took another stab at it. I got it working. Will get it cleaned up as a PR.

wt:
#953 There you go.

PR #953, opened 2025-11-12
build: Update circuit-playground-express dependencies.

wt:
# Summary
I updated to the latest version of the atsamd-hal. A number of other deps needed to be updated to get there.

I also updated all the examples to compile. Examples for uart and usb serial needed substantial updates so I just copied the version from the feather_m0 crate and ported them.

I tested the following examples:
* blinky_basic
* neopixel_rainbow
* usb_serial

I wasn't able to test the uart example, but it does compile.

# Checklist
  - [ ] Clean up accelerometer example

Closes: #951

### DIFF
--- a/boards/circuit_playground_express/examples/neopixel_rainbow.rs
+++ b/boards/circuit_playground_express/examples/neopixel_rainbow.rs
@@ -26,8 +26,8 @@ use hal::timer_traits::InterruptDrivenTimer;
 use pac::{CorePeripherals, Peripherals};
 
 use smart_leds::{
-    hsv::{hsv2rgb, Hsv},
     SmartLedsWrite,
+    hsv::{Hsv, hsv2rgb},
 };
 use ws2812_timer_delay as ws2812;
 
--- a/boards/circuit_playground_express/src/lib.rs
+++ b/boards/circuit_playground_express/src/lib.rs
@@ -17,7 +17,7 @@ use hal::sercom::{Sercom0, Sercom3, Sercom4};
 use hal::time::Hertz;
 
 #[cfg(feature = "usb")]
-use hal::usb::{usb_device::bus::UsbBusAllocator, UsbBus};
+use hal::usb::{UsbBus, usb_device::bus::UsbBusAllocator};
 
 hal::bsp_peripherals!(
     Sercom0 { SpiSercom }


boards/circuit_playground_express/Cargo.toml
```diff
@@ -1,31 +1,36 @@
 [package]
 name = "circuit_playground_express"
-version = "0.11.1"
+categories = ["embedded", "hardware-support", "no-std"]
+version = "0.12.0"
```
jbeaurivage: Please leave the crate version as-is. Our versions are automatically updated by CI upon release :)

boards/circuit_playground_express/Cargo.toml
```diff
@@ -1,31 +1,36 @@
...
 
 [dependencies.cortex-m-rt]
 version = "0.7"
 optional = true
 
 [dependencies.atsamd-hal]
-version = "0.14"
+version = "0.22.2"
 default-features = false
+path = "../../hal"
```
jbeaurivage: Also please remove this path dependency. Only Tier 1 boards depend on the `master` branch - you can read about board tiers [here](https://github.com/atsamd-rs/atsamd?tab=readme-ov-file#pac-and-bsp---peripheral-access-crate-and-board-support-package).

hal/src/sercom/pad/impl_pad_thumbv6m.rs
```diff
@@ -427,12 +427,12 @@ pad_table!(
...
     PB02 {
         #[hal_cfg("sercom5")]
-        D: (Sercom5, Pad0),
+        D: (Sercom5, Pad0) + I2C,
     }
     #[hal_cfg("pb03")]
     PB03 {
         #[hal_cfg("sercom5")]
-        D: (Sercom5, Pad1),
+        D: (Sercom5, Pad1) + I2C,
```
jbeaurivage, line 456: Did you test the I2C functionality for these pins on hardware? It seems like the datasheet doesn't indicate these pins as I2C capable, but the wouldn't be the first time that we'd see something like that, ie capability hidden/not guaranteed by the datasheet.

wt: I'm basing this on the diagram here: https://cdn-learn.adafruit.com/assets/assets/000/139/941/original/makecode_circuit_playground_Adafruit_Circuit_Playground_Express_Pinout_2.png?1758740487

I was trying to enable support for the accelerometer.

It looks like the accelerometer is hooked up to PA00 and PA01. It also looks like PB03/02 have SDA/SCL designations. I don't know enough for more at this point. Can you please help me understand what needs to be done? I can take off the I2C traits if need be.

boards/circuit_playground_express/Cargo.toml
wt: Will do.

boards/circuit_playground_express/Cargo.toml
wt: k

hal/src/sercom/pad/impl_pad_thumbv6m.rs
jbeaurivage: Hmm that's really weird. Seems like the recognize the existence of i2c capable vs non-i2c capable pins as evidenced elsewhere on the diagram, yet the chose to name non-i2c pins SDA and SCL?

Anyways, in your case you could probably name the PA00 and PA01 pins `accel_sda` and `accel_scl` or something similar in the pins declaration macro. Then in `bsp_peripherals!`, define a new peripheral instance named `AccelSercom` or similar, tied to SERCOM1-the diagram tells us the PA00 and PA01 pins are pads for SERCOM1. Weirdly enough it doesn't seem like there are any user-exposed pins capable of i2c.

So, long story short: I would remove those `+ I2C` declarations from the HAL.

wt: I have updated the PR with an example communicates with the accelerometer via those pins using i2c. If I remove the `+ I2c` from those pins, the program does not work. I am trying to figure out how to do what you said.

In the latest version of the patch I included the `+ I2c` change, I have included the path in the Cargo.toml since I use changes to the other crate. I respect that i need to change something so that either I don't modify the dependency (samd21g crate) and this crate at the same time. However, I want you to see this in context.

Given that, I know that this change needs some zhuzhing before landing. I just wanted to make sure to call it out.

Thanks for your help.

wt: > Anyways, in your case you could probably name the PA00 and PA01 pins `accel_sda` and `accel_scl` or something similar in the pins declaration macro. Then in `bsp_peripherals!`, define a new peripheral instance named `AccelSercom` or similar, tied to SERCOM1-the diagram tells us the PA00 and PA01 pins are pads for SERCOM1. Weirdly enough it doesn't seem like there are any user-exposed pins capable of i2c.

The `accel_{sda/scl}` pins are PB03/02. They use sercom5. Please look at my new example to see something that actually works.

> So, long story short: I would remove those `+ I2C` declarations from the HAL.

I am not sure how to remove the + I2C declarations from the HAL (at least for PB03/02). Can I add those declarations only in this bsp crate somehow? If so, how?

Should I create a separate change that only introduces the `+ I2C` bits to the hal pad table definitions?

wt: Page 28-30 of https://cdn.sparkfun.com/assets/4/c/9/e/f/SAMD21-Family-DataSheet-DS40001882D.pdf appear to suggest that PA00/01 are on Sercom1 and PB03/03 are on Sercom5. And it looks like all of the sercom interfaces can be programmed to talk over i2c.

I don't know if I am interpreting that correctly, but, if I am, does that mean that these pads should be marked as I2C compatible for the d21 family?

sajattack: It seems like the way the HAL does this has changed since I last looked at it. Or maybe some difference between thumbv6m and thumbv7em. I'm confused beyond the scope of this PR now. All sercom padsets should support I2C I think. But the padset stuff might only be thumbv7em. I'm lost.

wt: @jbeaurivage Do you have any guidance here? I believe there are a number of pads in the d21 pad table that need to have the `+ I2C` annotation since they actually can be used to connect to an i2c bus. It looks like anything that is connected to a sercom can be programmed to speak i2c.

wt:
Oops. Forgot to push the lib.sh change.

wt:
I published #955 to just add the I2C trait to all sercom pads. If that change lands, I can remove that part from this change.

wt:
I updated this to depend on the `undoc-features` feature of the hal.

hal/src/sercom/pad/impl_pad_thumbv6m.rs
wt: I am going to resolve this conversation since I updated this to use the new undoc-features feature of the hal.

wt:
attn @jbeaurivage:

Other than the accelerometer example, I think it might be in decent shape. Maybe the accelerometer example should be pushed to another diff, as I was just using it to prove that I could talk to the accelerometer.

It does use a brand new feature of the hal, and I know you want me to remove the path dep to the hal, which I can only do after a new release of the hal since it uses this new feature. How do you want to handle this?

boards/circuit_playground_express/Cargo.toml
```diff
@@ -1,42 +1,49 @@
...
+categories = ["embedded", "hardware-support", "no-std"]
 version = "0.11.1"
 authors = ["Paul Sajna <paulsajna@gmail.com>"]
 description = "Board Support crate for the Adafruit Circuit Playground Express"
 keywords = ["no-std", "arm", "cortex-m", "embedded-hal"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/atsamd-rs/atsamd"
 readme = "README.md"
-edition = "2018"
+edition = "2021"
```
jbeaurivage: Any reason this couldn't be the 2024 edition instead?

boards/circuit_playground_express/Cargo.toml
```diff
@@ -1,42 +1,49 @@
...
 
 [dependencies.cortex-m-rt]
 version = "0.7"
 optional = true
 
 [dependencies.atsamd-hal]
-version = "0.14"
+version = "0.22.2"
 default-features = false
+path = "../../hal"
```
jbeaurivage: As discussed, can you remove this path dependency?

boards/circuit_playground_express/Cargo.toml
```diff
@@ -1,42 +1,49 @@
...
 smart-leds = "0.3.0"
+heapless = "0.9.2"
 
 [dev-dependencies.ws2812-timer-delay]
 features = ["slow"]
 version = "0.3.0"
 
 [features]
 # ask the HAL to enable atsamd21g support
+# Also, the accelerometer is connected via an undocumented i2c pad.
```
jbeaurivage: This comment should probably be removed until the accelerometer stuff gets merged in a later PR

hal/src/sercom/pad/impl_pad_thumbv6m.rs
```diff
@@ -438,12 +438,12 @@ pad_table!(
     #[hal_cfg("pb02")]
     PB02 {
         #[hal_cfg("sercom5")]
-        D: (Sercom5, Pad0),
+        D: (Sercom5, Pad0) + I2C,
```
jbeaurivage, line 446: This should be reverted in favor of #956.

hal/src/sercom/pad/impl_pad_thumbv6m.rs
```diff
@@ -438,12 +438,12 @@ pad_table!(
...
     }
     #[hal_cfg("pb03")]
     PB03 {
         #[hal_cfg("sercom5")]
-        D: (Sercom5, Pad1),
+        D: (Sercom5, Pad1) + I2C,
```
jbeaurivage, line 456: Same here

hal/src/sercom/pad/impl_pad_thumbv6m.rs
jbeaurivage: Or at least, be hidden behind the `undoc-features` flag

boards/circuit_playground_express/Cargo.toml
wt: Yeah, good catch.

boards/circuit_playground_express/Cargo.toml
wt: Sorry, meant to remove. I intended to add it back locally since it depends on unreleased changes to the hal. If we can't land it with this, then this cannot land until release of the hal.

hal/src/sercom/pad/impl_pad_thumbv6m.rs
wt: I just filed #958.

hal/src/sercom/pad/impl_pad_thumbv6m.rs
wt: also in #958

boards/circuit_playground_express/Cargo.toml
wt: I changed it.

wt:
Had to move back to edition 2021 because the usb example is broken in 2024. The global static muts are errors in 2024.

wt:
Pushed without the path dep. Realized that this will need to be modified to mention the new version if you want to be able to land shortly after cutting the new version.

wt:
I just pushed a second commit in this stack that moves the circuit-playground-express crate to the 2024 edition of rust. I can remove it if you like. The compromise here is that the usb examples are not able to be compiled with 2024 edition, so it must explicitly be held back.

The usb examples for all boards have this same issue. I could do something similar for all board crates en masse as a different PR if this move doesn't cause any issues.

If you want me to break this edition move into a separate PR, let me know. I am happy to do so.

PR #954, opened 2025-11-14
Switch itsybitsy_m4 BSP to latest HAL release.

codepainters:
# Summary

Changes necessary to make `itsybitsy_m4` BSP/examples build properly with HAL version `0.22`.
I had to make this change to test #948 on my ItsyBitsy M4 Express board.

BTW, I can see some inconsistencies when it comes to using HAL from repo vs pulling a crate. Some BSPs (e.g. `feather_m4`) use `path` to reference the repo:
```
[dependencies.atsamd-hal]
default-features = false
path = "../../hal"
version = "0.22.2"
```
while others (like `itsybitsy_m4`) simply pull the crate as a dependency. 
Is there any policy in place?

### DIFF
--- a/boards/itsybitsy_m4/Cargo.toml
+++ b/boards/itsybitsy_m4/Cargo.toml
@@ -24,7 +24,7 @@ version = "0.7"
 optional = true
 
 [dependencies.atsamd-hal]
-version = "0.21"
+version = "0.22.2"
 default-features = false
 
 [dependencies.usb-device]
--- a/boards/itsybitsy_m4/examples/sercom_interrupt.rs
+++ b/boards/itsybitsy_m4/examples/sercom_interrupt.rs
@@ -42,12 +42,12 @@ use bsp::hal::{
     prelude::*,
     sercom::{
         uart::{self, BaudMode, Flags, Oversampling},
-        IoSet3, Sercom0,
+        Sercom0,
     },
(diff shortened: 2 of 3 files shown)


codepainters:
Hmm, it looks like `sercom_interrupt`  example fails compiling, let me check.

jbeaurivage:
Hey @codepainters, this isn't a proper review (yet), but I just wanted to chime in on the path vs crates.io dependencies. We use a tier system for our BSPs, where Tier 1 are always pulling the HAL from the latest master, while Tier 2 specify a crates.io version, which may or may not be the latest release. This helps alleviate maintainer burden when breaking changes happen in the HAL, while still having some boards to test against. You can find the boards' tiers in the readme or in crates.json.

This board is a Tier 2, so it's correct to leave the dep to the crates.io release.

codepainters:
Thanks for the clarification! Now it makes much more sense.

boards/itsybitsy_m4/Cargo.toml
```diff
@@ -24,7 +24,7 @@ version = "0.7"
 optional = true
 
 [dependencies.atsamd-hal]
-version = "0.21"
+version = "0.22.2"
```
jbeaurivage, line 27: This could probably be 0.23 now.

PR #955, opened 2025-11-15
Fix: add I2C trait on all sercom pads

wt:
# Summary
All sercom pads can be used to communicate with i2c devices. This change removes just makes all sercom pads capable of being used for i2c. I looked in both the samd11 and samd21 datasheets. They both say that sercom pads ban be used to talk to i2c devices.

# Checklist


jbeaurivage:
Thanks @wt, but after our Matrix discussions, we will instead go with the solution proposed in #956.

PR #956, opened 2025-11-16
changed!: Gate undocumented chip features behind the `undoc-features` Cargo feature

jbeaurivage:
As discussed on Matrix, we would like to make it clear to our users that we optionally support pin configurations that aren't supported or documented in the datasheets. 

Historically this has been `UndocSercom1` and `UndocSercom2`. We have recently discovered that the Circuit Playground Express uses PA00 and PA01 to talk to an I2C accelerometer, despite these pins not being explicitly I2C-capable according to the datasheet.

With this breaking change, the users and BSPs will have to opt into using these undocumented features

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,8 +162,24 @@ provide any guarantees with respect to those.
 
 Currently, we provide these features undocumented features:
 
-* `UndocIoSet1` and `UndocIoSet2` for SAMx5x devices
-* SAMD21: PA00 and PA01 may be used for SERCOM I2C
+### SAMD21:
+
+* Mark `PA00` as I2C-capable according to `circuit_playground_express`.
+
+* Mark `PA01` as I2C-capable according to `circuit_playground_express`.
+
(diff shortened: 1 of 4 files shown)


rnd-ash:
Looks good @jbeaurivage ! I'd be happy for this to be merged. Only thing is, do we also add a feature flag to the BSPs themselves as well to pass this feature flag up to the HAL?

jbeaurivage:
> Looks good @jbeaurivage ! I'd be happy for this to be merged. Only thing is, do we also add a feature flag to the BSPs themselves as well to pass this feature flag up to the HAL?

IMO only the BSPs that require the flag to compile should enable it. Then, if needed the users can always enable it by bringing `atsamd-hal` in their dependencies, and enabling the feature there.

rnd-ash:
Fair point. Then in that case, LGTM unless anyone else has some issues with this

Issue #911, opened 2025-06-27
TccPwm::get_max_duty() is -1 too small


fko-kuptec:
The maximum duty cycle returned by the `Tcc?Pwm` peripherals via [`Pwm::get_max_duty`](https://github.com/atsamd-rs/atsamd/blob/d3aa0e1b0365c9e6349c84370071cbdbcb9248bd/hal/src/peripherals/pwm/d5x.rs#L638) seems to be one too small.

At least when setting exactly this duty cycle, the resulting signal (checked via oscilloscope) is not a fully 100% high signal - it is still interrupted by short spikes to low. However, if I set the returned maximum duty cycle + 1, the resulting signal is actually a duty cycle of 100% - a permanently high signal.

To be honest, I don't really understand the driver code. I didn't spend much time with these peripherals, yet. Therefore, I cannot tell, whether just adding `+1` is a good fix.

rnd-ash:
Can you provide some sample code with your clocking setup? It is likely some division error flooring the result which would cause this, might not be apparent on all clocking setups

codepainters:
Just spotted this issue - see my #948, it might be related. It might be that the timer period is 1 too high.

codepainters:
I've checked with the following code:
```rs
    // Gclk3 as a 1MHz source for the timer
    let timer_clk_gen = clocks.configure_gclk_divider_and_source(Gclk3, 48, Dfll48m, true)
        .unwrap();

    let timer_clk =clocks.tcc0_tcc1(&timer_clk_gen).unwrap();
    let mut pwm = Pwm1::new(&timer_clk, 250000.Hz(), peripherals.tcc1, &mut peripherals.pm);
    pwm.set_duty(Channel::_0, 0);
```
This gives 4 ticks period.

With my  https://github.com/atsamd-rs/atsamd/pull/949 PR I get (checked with an oscilloscope):
* 0 -> 0% (0 ticks)
* 3 -> 75% (3 ticks)
* 4 -> 100%
* `get_max_duty()` -> 75%

Looking at the code:
```rs
    fn get_max_duty(&self) -> Self::Duty {
        let top = self.tcc.per().read().bits();
        top
    }
```
It seems to suffer from off-by-one, too. `PER` register contains the top value, i.e., period - 1. The name is misleading.

fko-kuptec:
[...]

PR #959, opened 2025-11-18
Fix for off-by-one in get_max_duty()

codepainters:
# Summary

This PR fixes off-by-one issue in TC/TCC `get_max_duty()` function, for details see #911.

Change verified with Adafruit ItsiBitsy M4 and Sparkfun ATSAMD21G Mini boards.

### DIFF
--- a/hal/src/peripherals/pwm/d11.rs
+++ b/hal/src/peripherals/pwm/d11.rs
@@ -97,7 +97,7 @@ impl $crate::ehal::pwm::SetDutyCycle for $TYPE {
     fn max_duty_cycle(&self) -> u16 {
         let count = self.tc.count16();
         let top = count.cc(0).read().cc().bits();
-        top + 1
+        top.saturating_add(1)
     }
 
     fn set_duty_cycle(&mut self, duty: u16) -> Result<(), Self::Error> {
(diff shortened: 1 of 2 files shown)


jbeaurivage:
Thanks @codepainters. However I would suggest, if we want the functions to keep returning `u16`s, to use `saturating_add()` instead of the regular `+` operator to cover the case where `CC == u16::MAX`. What do you think?

codepainters:
Hmm, fair point. Actually it's quite interesting what the chip actually do if 0 is used as a duty cycle value, let me check the datasheet (and perhaps the hardware).

codepainters:
Finally found a moment to take a look, I've done some experiments with ItsyBits M4. 

I've tried setting maximum period (2^16 1) and try various duty cycle values (`CC1` register):
* 0 - output constantly low
* 2^16 - 1 - gives 100% - 1 tick (as expected)

So, with maximum period value it is not possible to get 100% PWM.

All that said, I guess that saturating add is a good approach here (the configuration with maximum period is somewhat pathological anyway).

jbeaurivage:
Sounds good, if you just want to make that change, I can get this merged.

codepainters:
Just added `saturating_add()`  (and rebased on master branch). 

Note: I only added `saturating_add()` for TC timer, where `CC0` register is 16-bit and return type is `u16`.  For TCC, PER is 24-bit and return type is `u32`, so there's no overflow possibility.

PR #960, opened 2025-11-19
docs: Document newly undoc I2C-capable pins (PB02,PB03)

jbeaurivage:
The documentation for these 2 newly undocumented I2C-capable pins was missing.


rnd-ash:
Great! Let's merge

PR #962, opened 2025-11-20
Improve cpe examples

wt:
# Summary
* Make the cpe usb_serial example use rtic. This allows the removal of the static muts, which are not legal in edition 2024.
* Add an blinky example that uses rtic.

# Checklist


wt:
I isolated the feature change that adds the rtic support for this board, so this should only land after #966.

sajattack:
Would you mind cleaning up these clippy lints while you're here?

https://github.com/atsamd-rs/atsamd/actions/runs/20494027875/job/58890997512?pr=962#step:6:86

sajattack (approved):

wt:
I added a second commit to fix the clippy errors. Those were from a previous commit. I guess I didn't see them then. Sorry about that.

PR #963, opened 2025-11-20
USB allocation rework and simplification

rnd-ash:
# Summary
Our current USB Subsystem has a few issues with buffer allocation. This PR attempts to fix them.

## Issues
1. All endpoints share a fixed 2048 byte buffer which cannot be adjusted
2. Max packet size is tied to allocation size, which means we never utilize SAMx's multi-packet transmit feature, and allows the user to setup invalid packet sizes for the FS USB when simply just trying to have a larger USB endpoint buffer.
3. Buffer allocation does not handle overflow correctly, resulting in a panic if we attempt to allocate more than 2048 bytes in total.

## Addressing the issues

To address (1), I have now added in some feature flags for the HAL, which allows the user to set the global USB endpoint allocation buffer to either 1KB, 2KB, 4KB or 8KB. 2KB remains the default to maintain compatibility (The 2KB feature flag is not required to compile the crate)

To address 2 and 3, and to simplify the USB buffer allocation method, This global buffer is now divided equally by 16 (Number of endpoints on our chips), rather than attempting to be split dynamically, it is also stored in a structure that is 4 byte aligned, to simplify the allocation process.

This PR also makes the USB connection more stable under high loads, since we now utilize the multi-packet RX/TX feature properly, which allows a user to send large bulk data, and the hardware will send it over smaller USB packets, before generating an interrupt once sending is complete. Before, we were limited by the USB packet size to the endpoint buffer size, meaning realistically, an interrupt was generated every 64 bytes being sent or received. 

##  Old implementation

|Global USB buffer size|2048|
|:-:|:-:|
|Buffer size per endpoint|Dynamic|
|Number  of packets before Tx/Rx interrupt fired|1|

##  New implementation
|Feature flag|Global USB buffer size|Buffer size per endpoint|Number of 64 byte packets Tx/Rx before interrupt fires|
|:-:|:-:|:-:|:-:|
|`usb-buffer-1k`|1024 bytes|64 bytes|1|
|`usb-buffer-2k` (Default)|2048 bytes|128 bytes|2|
|`usb-buffer-4k`|4096 bytes|256 bytes|4|
|`usb-buffer-8k`|8192 bytes|512 bytes|8|

### DIFF
--- a/hal/Cargo.toml
+++ b/hal/Cargo.toml
@@ -221,12 +221,3 @@ usb-buffer-8k = ["usb"] # 8192 byte buffer shared between all USB endpoints
 # The `device` feature tells the HAL that a device has been selected from the
 # feature list. It exists mostly to provide better error messages.
 device = []
-
-
-[package.metadata.feature-groups]
-usb-buffer.exactly-one = { features = [
-  "usb-buffer-1k",
-  "usb-buffer-2k",
-  "usb-buffer-4k",
-  "usb-buffer-8k",
-] }


rnd-ash:
@jbeaurivage if this PR is too much to merge at once, I am happy to split it so that at least I can fix the buffer allocater bug which causes the HAL to panic rather than return an error (Which the function says it would do) should you require more than 2KB for USB endpoints

hal/Cargo.toml
```diff
@@ -210,3 +221,12 @@ undoc-features = []
...
 device = []
+
+
+[package.metadata.feature-groups]
+usb-buffer.exactly-one = { features = [
+  "usb-buffer-1k",
+  "usb-buffer-2k",
+  "usb-buffer-4k",
+  "usb-buffer-8k",
+] }
```
jbeaurivage: I think this is meant to be used with [`cfg_feature_groups`](https://docs.rs/cfg_feature_groups/latest/cfg_feature_groups/). Although I don't think it really hurts anything either.

hal/Cargo.toml
```diff
@@ -195,10 +195,21 @@ dma = []
...
+
+# USB buffer sizes
+# These denote the size of the global buffer for all USB endpoints.
+# The buffer is split between all device endpoints equally
+
+usb-buffer-1k = [] # 1024 byte buffer shared between all USB endpoints
+usb-buffer-2k = [] # 2048 byte buffer shared between all USB endpoints (Default)
+usb-buffer-4k = [] # 4096 byte buffer shared between all USB endpoints
+usb-buffer-8k = [] # 8192 byte buffer shared between all USB endpoints
+
```
jbeaurivage, line 212: How do you feel about this:

```diff
- usb-buffer-1k = [] # 1024 byte buffer shared between all USB endpoints
- usb-buffer-2k = [] # 2048 byte buffer shared between all USB endpoints (Default)
- usb-buffer-4k = [] # 4096 byte buffer shared between all USB endpoints
- usb-buffer-8k = [] # 8192 byte buffer shared between all USB endpoints
+ usb-buffer-1k = ["usb"] # 1024 byte buffer shared between all USB endpoints
+ usb-buffer-2k = ["usb"] # 2048 byte buffer shared between all USB endpoints (Default)
+ usb-buffer-4k = ["usb"] # 4096 byte buffer shared between all USB endpoints
+ usb-buffer-8k = ["usb"] # 8192 byte buffer shared between all USB endpoints
```

hal/Cargo.toml
rnd-ash: It was just something i saw @kyp44 do in his USB Serial crate that had a similar compile time buffer size. I didn't actually realize this was something that wasn't part of the standard TOML file. Should we keep it then?

hal/Cargo.toml
rnd-ash: Sure!

rnd-ash:
@jbeaurivage awesome! I've updated the TOML with your suggestion. Just one last thing, should I  remove the "FIXME" comments regarding potential heap allocator before we merge? (Since this code now allows for variable buffer sizes, I don't think a heap allocator is necessary now)

hal/Cargo.toml
jbeaurivage: Eh, let's remove it for now, if we ever decide to use that crate we can always add it back later.

jbeaurivage:
> @jbeaurivage awesome! I've updated the TOML with your suggestion. Just one last thing, should I remove the "FIXME" comments regarding potential heap allocator before we merge? (Since this code now allows for variable buffer sizes, I don't think a heap allocator is necessary now)

Thanks. Yeah, I don't think we need to leave the fixme comment if we're happy with this, and the fact that we are always doing static allocation.

rnd-ash:
Right all done. I think this is ready now!

PR #966, opened 2025-11-25
Feature: Add RTIC support for CPE

wt:
# Summary
This change simply allows rtic to be used with the circuit-playground-express board.

# Checklist


wt:
I have isolated this from my change that adds some RTIC examples from this board.

PR #971, opened 2025-12-10
Update nano33iot hal

Dox5:
# Summary
Update the `arduino_nano33iot` to the latest HAL (0.23.0). I had to add I2C to the pins `PB08` and `PB09` for HAL to build - those are the labelled I2C pins on this board, I'm new to embedded so I don't know if/how that effects other boards...

I have built the hal and all the examples, and have programmed and verified `blinky_basic` and `usb_logging` work as expected, I don't have a way to test the other examples on my hardware though.

Happy to make any changes as needed! :)

# Checklist

### DIFF
--- a/boards/arduino_nano33iot/Cargo.toml
+++ b/boards/arduino_nano33iot/Cargo.toml
@@ -17,6 +17,7 @@ optional = true
 [dependencies.atsamd-hal]
 version = "0.23.0"
 default-features = false
+features = ["undoc-features"]
 
 [dependencies.usb-device]
 version = "0.3.2"


Dox5:
Ah - will I have to do this as two PRs, one to update the pins and then to update the HAL, the build failure looks like might've been missing the pin change?

DCRichards:
> Ah - will I have to do this as two PRs, one to update the pins and then to update the HAL, the build failure looks like might've been missing the pin change?

@Dox5 You could use a local path like the feather_mo board does https://github.com/atsamd-rs/atsamd/blob/master/boards/feather_m0/Cargo.toml#L24. Otherwise will need to do the hal work separately and have someone merge and tag. Up to the maintainers what they would prefer here of course

DCRichards:
FWIW I forked this last night to pick up my Nano 33 IOT project from some years ago and my changes are pretty much identical to this, including the need to add the I2C trait to the pins

Dox5:
based on a comment in another PR I've split this one, this needs #975 to be merged for the build

PR #975, opened 2025-12-24
feat: add I2C capabilities to labelled pins for arduino_nano33iot

Dox5:
# Summary
Splitting out the HAL change from #971 - this adds the I2C trait to the pins labelled on the Arduino nano 33 iot board.

# Checklist

### DIFF
--- a/hal/src/sercom/pad/impl_pad_thumbv6m.rs
+++ b/hal/src/sercom/pad/impl_pad_thumbv6m.rs
@@ -457,13 +457,23 @@ pad_table!(
     }
     #[hal_cfg("pb08")]
     PB08 {
+        #[cfg(not(feature = "undoc-features"))]
         #[hal_cfg("sercom4")]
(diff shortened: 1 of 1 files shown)


jbeaurivage:
Thanks @Dox5. Can you gate this added I2C function behind the `undoc-features` Cargo feature? See #956.

Dox5:
Updated - out of curiosity, what makes this undocumented? I was looking at the table in 7.1 in https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ProductDocuments/DataSheets/SAM-D21-DA1-Family-Data-Sheet-DS40001882.pdf which says function D for PB08 and PB09 are SERCOM-ALT? It it because that is only for the SAMD2xG and SAMD2xJ?

jbeaurivage:
Thanks! No, seems like Atmel has only validated that certain SERCOM pins are actually I2C-capable. Table 7-5 tells us which ones.

PR #983, opened 2026-01-25
fix(uart): Fix embedded_io::Read::read impl

### LINKED ISSUE #982: Help with UART RX
Trying to get uart working on my atsam board with this crate. I am using the Atmel SAM D21 Xplained kit (ATSAMD21J18A) on sercom0. I am able to transmit from the board, but having a difficult time receiving. 

If I call 
```
let mut rx_buffer = [0x00; 50];
uart_rx.read(rx_buffer )
```

The call to read blocks forever. However, if I make an unsafe block and use ```let bytes = uart_rx.read_data()``` , I can see that the bytes are being received by the board. 

I may have missed something very silly, if that is the case  I apologize!

jbeaurivage:
Fix embedded_io::Read::read implementation for sercom::Uart

# Summary

According to the `embedded_io::Read::read` contract, the method should block until at least one byte becomes available, then read as many bytes as possible into the buffer *without blocking*, then return. 

The current implementation behaves as `read_exact` should, ie, read as many bytes as specified into the buffer, potentially blocking any number of times as needed; this PR fixes that discrepancy. 

Potentially closes #982.

### DIFF
--- a/hal/src/dmac/channel/mod.rs
+++ b/hal/src/dmac/channel/mod.rs
@@ -543,6 +543,8 @@ impl<Id: ChId> Channel<Id, Ready> {
     ///   compatible lengths. You must guarantee that:
     ///   - Either `source` or `dest` has a buffer length of 1, or
     ///   - Both buffers have the same length.
+    /// * The source and destination buffers must have a length smaller or equal
+    ///   to `u16::MAX`.
     /// * You must ensure that the transfer is completed or stopped before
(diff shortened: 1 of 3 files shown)


jbeaurivage:
@itsLydt, would you mind testing this out to see if it behaves as you would expect?

jbeaurivage:
After testing these changes myself, they appear to behave as expected (ie, conform to the `embedded_io::Read` contracts).

jbeaurivage:
Self-merging based on @sajattack's approval.

PR #984, opened 2026-01-25
fix(dmac): Check+document that the DMAC transfer lengths fit in a u16

### LINKED ISSUE #981: DMA silently fails for transfers longer than u16::MAX beats
Ran into this while messing around with a DVP camera, if I make my frame too large (>181x181, 2 bytes per pixel), the buffer extends past 65535 bytes. Looking at `write_descriptor`, we truncate the len to a u16 so that it fits into the BTCNT. Obviously, the hardware is the limitation here, and short of supporting linked transfers, there's no way around this. However, silently truncating seems like less than ideal behavior to me, and certainly caused me a bit of confusion when stuff just stopped working!

Would a `debug_assert!()` be appropriate? Or a note in the dmac/sercom docs that there's a maximum length? Or a change to `check_buffer_pair` + `dmac::Error::TooManyBeats`?  

```rust
#[inline]
pub(crate) unsafe fn write_descriptor<Src: Buffer, Dst: Buffer<Beat = Src::Beat>>(
    descriptor: &mut DmacDescriptor,
    source: &mut Src,
    destination: &mut Dst,
    next: *mut DmacDescriptor,
) {
    let src_ptr = source.dma_ptr();
    let src_inc = source.incrementing();
    let src_len = source.buffer_len();

    let dst_ptr = destination.dma_ptr();
    let dst_inc = destination.incrementing();
    let dst_len = destination.buffer_len();

    let length = core::cmp::max(src_len, dst_len);

    // Suggest something equivalent to this:
    debug_assert!(
        length <= u16::MAX as usize,
        "DMAC transfer length {} exceeds maximum of 65535 beats",
         length
    );

    // Channel::xfer_complete() tests the channel enable bit, which indicates
    // that a transfer has completed iff the blockact field in btctrl is not
    // set to SUSPEND.  We implicitly leave blockact set to NOACT here; if
    // that changes Channel::xfer_complete() may need to be modified.
    let btctrl = sram::BlockTransferControl::new()
        .with_srcinc(src_inc)
        .with_dstinc(dst_inc)
        .with_beatsize(Src::Beat::BEATSIZE)
        .with_valid(true);

    *descriptor = DmacDescriptor {
        // Next descriptor address:  0x0 terminates the transaction (no linked list),
        // any other address points to the next block descriptor
        descaddr: next,
        // Source address: address of the last beat transfer source in block
        srcaddr: src_ptr as *mut _,
        // Destination address: address of the last beat transfer destination in block
        dstaddr: dst_ptr as *mut _,
        // Block transfer count: number of beats in block transfer
        btcnt: length as u16,
        // Block transfer control: Datasheet  section 19.8.2.1 p.329
        btctrl,
    };
}
```

jbeaurivage:
# Summary
The BTCNT field of a DMAC transfer descriptor is a u16. We currently truncate the transfer lengths to u16, leading to potentially surprising results.

Here this PR adds a length check to the `check_buffer_pair` methods, and documents the buffer length requirement in all unsafe functions where the buffer pairs are explicitly unchecked.

Potentially fixes #981

### DIFF
--- a/hal/src/dmac/channel/mod.rs
+++ b/hal/src/dmac/channel/mod.rs
@@ -543,6 +543,8 @@ impl<Id: ChId> Channel<Id, Ready> {
     ///   compatible lengths. You must guarantee that:
     ///   - Either `source` or `dest` has a buffer length of 1, or
     ///   - Both buffers have the same length.
+    /// * The source and destination buffers must have a length smaller or equal
+    ///   to `u16::MAX`.
     /// * You must ensure that the transfer is completed or stopped before
     ///   returning the [`Channel`]. Doing otherwise breaks type safety, because
     ///   a [`Ready`] channel would still be in the middle of a transfer.
@@ -676,10 +678,6 @@ impl<Id: ChId> Channel<Id, ReadyFuture> {
     ///
     /// # Safety
     ///
-    /// * This method does not check that the two provided buffers have
-    ///   compatible lengths. You must guarantee that:
-    ///   - Either `source` or `dest` has a buffer length of 1, or
-    ///   - Both buffers have the same length.
     /// * You must ensure that the transfer is completed or stopped before
     ///   returning the [`Channel`]. Doing otherwise breaks type safety, because
     ///   a [`ReadyFuture`] channel would still be in the middle of a transfer.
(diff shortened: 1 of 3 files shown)


jbeaurivage:
Self-merging based on @sajattack's approval.

PR #985, opened 2026-01-25
fix(dmac): Use inline asm for compiler fences in DMAC

jbeaurivage:
# Summary

According to [this discussion](https://users.rust-lang.org/t/compiler-fence-dma/132027), it seems `core::sync::atomic::{fence, compiler_fence}` aren't a strong guarantee that the compiler won't reorder memory accesses beyond the fence. The suggested fix is to add an inline assembly call, which apparently guarantees there won't be any memory accesses beyond the asm call.

### DIFF
--- a/hal/src/dmac/channel/mod.rs
+++ b/hal/src/dmac/channel/mod.rs
@@ -33,7 +33,7 @@
 
 #![allow(unused_braces)]
 
-use core::marker::PhantomData;
+use core::{marker::PhantomData, ptr::NonNull};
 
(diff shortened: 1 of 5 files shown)


rnd-ash:
Looks good @jbeaurivage , but maybe, its best to put the fence functions near the crate root for future use by other peripherals should we need to in future? - Or is this specific to DMAC?

jbeaurivage:
I think this should remain a DMAC-specific thing, as peripherals are expected to control DMA through the DMAC API.

PR #988, opened 2026-02-23
feat(dmac)!: add Interrupts type parameter to DMA channels

QuartzShard:
# Summary
Follow on from #977: Adds a new Typelevel enum `Interrupts` with states `Available` and `Blocked`. This gates `enable_interrupts` and associated methods so that a `Blocked` channel cannot change interrupt configuration.

This allows the implementation of `From<>` to convert Async-mode channels (`ReadyFuture`), into Sync-mode channels (`Ready`) (and vice versa) to permit the use of `Transfer` when the controller is configured in Future mode - without risking Incorrect behavior why conflicting with the ISR that the async DMAC uses. My motivation for this was requiring both async usage for UART and a Circular transfer for reading from a DVP camera - but a ReadyFuture channel cannot function with the Transfer api.

# Checklist

### DIFF
--- a/boards/atsame54_xpro/CHANGELOG.md
+++ b/boards/atsame54_xpro/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.13.3](https://github.com/atsamd-rs/atsamd/compare/atsame54_xpro-0.13.2...atsame54_xpro-0.13.3) - 2026-03-02
+
+### Other
+
+- updated the following local packages: atsamd-hal
+
 ## [0.13.2](https://github.com/atsamd-rs/atsamd/compare/atsame54_xpro-0.13.1...atsame54_xpro-0.13.2) - 2026-02-20
 
 ### Other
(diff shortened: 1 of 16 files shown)


QuartzShard:
Failing on `clippy::type_complexity` lint, would like advice on how to address. 
The pattern `Transfer<Channel<Ch::ID, Busy, Ch::Interrupts>, BufferPair<B, Self>>` shows up in a few spots, and with the new param, upsets clippy. Is this a valid ignore, or should an alias be introduced?

rnd-ash:
I guess introducing a type alias for a Busy channel with Interrupts should work, since a similar concept exists for other peripherals, just not sure what to include exactly in the type Alias...

QuartzShard:
I'm happy with the state of this now - finished with janitorial docs stuff

QuartzShard:
I've been [writing a Ring Buffer](https://github.com/QuartzShard/atsamd/tree/ringbuffer-uart) on top of this, both to actually use and as a stress test of the API, and on reflection, I am not convinced this is the correct approach.

The main point of contention is that the current ISR for the Async DMA unconditionally disables channels when TCMPL fires, which is correct for one-shot channels (which is the entire current story for async DMA). This is wrong for circular/linked transfers, which was what I was attempting to solve by disallowing the use of interrupts when the ISR would cause issues; But really, shouldn't the ISR be made less opinionated? The channel can be disabled in `poll()` for TransferFutures before returning Ready. 

I'm thinking to either move channel management into poll calls, or make the disable conditional on a `static CHANNEL_MODE: [AtomicBool; NUM_CHANNELS]`, which then makes all of this unecessary.

PR #989, opened 2026-02-23
fix(adc):  clearing flags when reading mutible channels & make it somewhat consistent

nicolube:
# Summary
Fixes ADC reads with on multible channels\
Make the way reads are discarded consistent

# Checklist


rnd-ash:
Thanks for this. Looks like there were a couple artifacts left over from when we were creating the new ADC implementation that slipped through review. I'd be happy for this to be merged

nicolube:
@jbeaurivage Hello,

How long does it ushaly takes until we get a new version?

Kind of need this fix in the create I want to publish.

PR #998, opened 2026-03-10
Fully implement DSU features (Including for D21/D11)

rnd-ash:
# Summary

This PR adds the DSU Peripheral implementation to the D21 and D11 series of chips, and adds the following methods for all chip families:

1. Free method (To release the DSU peripheral)
2. `is_debugger_present` function to check the presence of a debugger
3. `memory_test` function to run the MBIST test (Marked as unsafe due to it overwriting RAM)
4. Background memory test functionality (See below)

### Background memory testing

Using the `PollingMemoryTest` structure, and `Dsu::polling_memory_test`, its now possible to run a memory test in the background when the CPU is asleep, and abort the memory test, if it isn't completed yet, when the CPU wakes up.

I've added an example to the `PollingMemoryTest` documentation on how this can be used to check RAM in the background on an RTIC application, in the same way this is done on ECUs.

## Additional notes

The CRC32 algorithm has also been updated for the D21 specifically for an errata note when calculating in RAM space on silicon revisions A-D

### DIFF
--- a/hal/src/peripherals/dsu.rs
+++ b/hal/src/peripherals/dsu.rs
@@ -6,7 +6,7 @@
 //! - Run a memory test on RAM
 //! - Check if a debugger is connected
 #![warn(missing_docs)]
-
+#![allow(clippy::doc_lazy_continuation)]
 use atsamd_hal_macros::{hal_cfg, hal_macro_helper};
 
 #[hal_cfg("dsu-d5x")]
@@ -237,18 +237,16 @@ impl Dsu {
     /// ## Algorithm:
     /// 1. Write entire memory to '0', in any order.
     /// 2. Bit by bit read '0', write '1', in descending order.
-    /// 3. Bit by bit read '1', write '0', read '0', write '1',
-    ///   in ascending order.
+    /// 3. Bit by bit read '1', write '0', read '0', write '1', in ascending order.
     /// 4. Bit by bit read '1', write '0', in ascending order.
-    /// 5. Bit by bit read '0', write '1', read '1', write '0',
-    ///   in ascending order.
+    /// 5. Bit by bit read '0', write '1', read '1', write '0', in ascending order.
     /// 6. Read '0' from entire memory, in ascending order.
     ///
     ///
     /// - `address` is an address within the CPUs RAM space; must be
-    ///  word-aligned
+    ///   word-aligned
     /// - `length` is a length of memory region that is being tested.
-    ///  Must be word-aligned
+    ///   Must be word-aligned
     pub unsafe fn memory_test(&mut self, address: u32, length: u32) -> Result<()> {
         if address % 4 != 0 {
             return Err(Error::AlignmentError);
--- a/hal/src/peripherals/dsu.rs
+++ b/hal/src/peripherals/dsu.rs
@@ -12,13 +12,25 @@ use atsamd_hal_macros::{hal_cfg, hal_macro_helper};
 #[hal_cfg("dsu-d5x")]
 use crate::pac::{self, Pac};
 
+#[hal_cfg("dsu-d5x")]
+type DsuAhbClk = crate::clock::v2::ahb::AhbClk<crate::clock::v2::types::Dsu>;
+#[hal_cfg("dsu-d5x")]
+type DsuApbClk = crate::clock::v2::apb::ApbClk<crate::clock::v2::types::Dsu>;
+
 #[hal_cfg(any("dsu-d21", "dsu-d11"))]
 use crate::pac::{self, Pac1};
 
 /// Device Service Unit
+#[hal_macro_helper]
 pub struct Dsu {
     /// PAC peripheral
     dsu: pac::Dsu,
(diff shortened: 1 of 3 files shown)


sajattack:
I suppose this supercedes #925 ?

rnd-ash:
I guess so.....but, I completely forgot about @kyp44 PR for the DSU. I can update mine so we have clock V2 for D5x as well

rnd-ash:
@sajattack I updated my code based on the PR @kyp44 did, so we now have Clock V2 for the d5x implementation of the DSU

rnd-ash:
Alright, got everything to build OK - Had to migrate 1 Example for featherm4 to clock V2 API

hal/src/peripherals/dsu.rs
```diff
@@ -115,43 +190,266 @@ impl Dsu {
...
+    /// **CAUTION** This can overwrite critical data in RAM, use
+    /// with caution
+    ///
+    /// This is useful for background memory check tasks when the
+    /// CPU is at idle, as the test can be aborted immediately
+    /// using the handle once the CPU receives an incomming
+    /// interrupt.
+    ///
+    /// This should always be ran inside a critical-section
+    /// to avoid the CPU context-switching mid memory test.
```
jbeaurivage: The `**CAUTION**` comment and critical section mention should go in a `# SAFETY` section instead

hal/src/peripherals/dsu.rs
```diff
@@ -115,43 +190,266 @@ impl Dsu {
...
             Ok(!self.dsu.data().read().data().bits())
         }
     }
+
+    /// Performs a memory test on a section of RAM using "March LR"
+    /// algorithm, blocking until the test is completed.
+    ///
+    /// **CAUTION** This can overwrite critical data in RAM, use
+    /// with caution
+    ///
```
jbeaurivage, line 266: Should go in a `# SAFETY` section, along with a warning about running in a critical section

hal/src/peripherals/dsu.rs
```diff
@@ -115,43 +190,266 @@ impl Dsu {
...
-        while !self.is_done() && !self.has_failed() {}
-        if self.has_failed() {
-            return Err(Error::CrcFailed);
+        while !self.is_done() {
+            core::hint::spin_loop();
+        }
+
+        #[hal_cfg("dsu-d21")]
+        {
+            // Errita for D21 silicon versions A-D
```
jbeaurivage: Errita -> Errata

hal/src/peripherals/dsu.rs
```diff
@@ -115,43 +190,266 @@ impl Dsu {
...
+
+    /// Polls to see if the memory test is completed, without
+    /// aborting it.
+    ///
+    /// Use [Self::finish_now] to obtain the result of the
+    /// memory test
+    #[inline]
+    pub fn is_completed(&self) -> bool {
+        self.dsu.is_done()
+    }
```
jbeaurivage: Suggestion: use `nb`'s primitives instead, so that users may use `nb::block!`. This way the blocking `memory_test` method can also use it and avoid code duplication

hal/src/peripherals/dsu.rs
```diff
@@ -41,72 +72,116 @@ pub type Result<T> = core::result::Result<T, Error>;
...
+    #[hal_cfg(any("dsu-d21", "dsu-d11"))]
+    pub fn free(self) -> pac::Dsu {
+        self.dsu
+    }
+
+    /// Releases the DSU peripheral
+    #[hal_cfg(any("dsu-d5x"))]
+    pub fn free(self) -> (pac::Dsu, DsuAhbClk, DsuApbClk) {
+        (self.dsu, self.ahb, self.apb)
+    }
```
jbeaurivage, line 114: Based on the conclusions we come to in #1000, do we need to swreset and/or stop mclk? Same for `new` is a swreset necessary?

jbeaurivage (commented):
This isn't a super in-depth review yet, but I had some observations that I wanted to write down as I glanced over the PR

jbeaurivage:
If this is merged, will it also close #925?

sajattack:
> If this is merged, will it also close #925?

See here https://github.com/atsamd-rs/atsamd/pull/998#issuecomment-4099891734
and here https://github.com/atsamd-rs/atsamd/pull/998#issuecomment-4101118113

hal/src/peripherals/dsu.rs
rnd-ash: Not 100% familiar with nb's API, but I think this is now implemented correctly using it

hal/src/peripherals/dsu.rs
rnd-ash: Now this is a good question. I think we should leave the peripheral however it was by default. On our devices, the DSU is enabled by default (MCLK), so I think we should leave it untouched.

Also, since the DSU touches the debugger, I think its best to NOT modify it when we free the DSU Peripheral, in case this harms debug sessions.

PR #1000, opened 2026-03-13
Introduce more peripheral free methods for consistency

rnd-ash:
# Summary

This PR adds free methods for TC,TCC,NVM and TRNG peripherals. The DSU peripheral free implementation is covered in #998 , so I decided to not add it here as well.

I do have a couple observations @jbeaurivage before a potential merge regarding the other peripherals (Maybe worthy of a new issue created).

Essentially, looking at all the peripherals, there is inconsistency with how we handle free/new methods.

1. I noticed that some new/free methods are marked as `#[inline]` and some are not, even when there are bigger constructors (Look at the ADC), we mark it with `#[inline]` whilst some small constructors are not marked as `#[inline]`

2. Some free methods trigger a swrst on the peripheral being freed, and others do not. Which approach is better? - And, if swrst is the better option, why also not disable mclk on the chip too when the peripheral is freed?

### DIFF
--- a/hal/src/peripherals/nvm/mod.rs
+++ b/hal/src/peripherals/nvm/mod.rs
@@ -176,6 +176,12 @@ impl Nvm {
         Self { nvm }
     }
 
+    /// Releases the NvmCtrl resource
+    #[inline]
+    pub fn free(self) -> Nvmctrl {
+        self.nvm
+    }
+
     /// Raw access to the registers.
     ///
     /// # Safety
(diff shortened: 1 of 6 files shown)


jbeaurivage:
Regarding `#[inline]`: It's really just a compiler hint. Chances are the compiler will still inline the function given a sufficiently high optimization level. We could try to strive for consistency and enable the hint though.

Regarding swrst: I think the ideal approach would be to software reset both when taking ownership of the pac struct, and when releasing it (unless some peripherals would break somehow if reset?). That way the user isn't surprised by the behaviour. In any case it's important to document what the method is doing of course.

Regarding disabling mclk: Wouldn't that be something explicitly handled by the user under clock v2?

PR #1001, opened 2026-03-20
Timer - Fix misuse of ctrlbset and ctrlbclr

rnd-ash:
# Summary

The Timer TC Ctrlbset and Ctrlbclr can only have a one written to them to modify the peripherals non exposed CTRLB register.

This PR fixes the user of CTRLBSET and CTRLBCLR, such that there is no writing 0 to them (Which results in nothing happening) - Instead, the opposite register is used to have the intended effect

### DIFF
--- a/hal/src/peripherals/timer/d11.rs
+++ b/hal/src/peripherals/timer/d11.rs
@@ -57,11 +57,11 @@ where
         // need to manually read the bit here
         while count.ctrla().read().bits() & 1 != 0 {}
 
-        count.ctrlbset().write(|w| {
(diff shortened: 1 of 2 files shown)


jbeaurivage:
Can't believe this has been sitting around for so long-evidently the peripheral still worked despite this glaring bug?

rnd-ash:
I think it has always worked since the setting of these bits to 0 is the peripherals default values for these registers, and swrst was called prior to these mistaken ctrlb calls, thus, they effectively always did nothing. So we don't actually need the ctrlb register calls in the first place.

Issue #1007, opened 2026-05-17
SPI pipelined writes are disabled


jdeblese:
Got a weird one - using atsamd-hal 0.23.1 on a Feather M0 and compiling with release optimizations, I see extra transactions on the line between M0 and an SD card, as if the MISO bytes are being delayed by one word. I've verified the issue is in the write_word_by_word call by modifying it to use transfer_word_in_place instead of disabling the receiver, writing data, and then re-enabling the receiver.

Found this issue in the process of tracking down why embedded-sdmmc card intialization worked when compiled in debug mode but failed in release mode. I captured the SPI comms (had to write a new ExclusiveDevice that let me bring the CS line out to an extra pin, since the Feather SD CS line is not accessible) and tried to verify that embedded-sdmmc was deterministic by replaying the conversation, but was finding extra single-word read transactions in my capture. This led me to try modifying write_word_by_word.

Replaced this linked code with what's below, and the SD card finally initialized properly with a release build.
https://github.com/atsamd-rs/atsamd/blob/atsamd-hal-0.23.1/hal/src/sercom/spi/impl_ehal/mod.rs#L155-L171

```rust
    fn write_word_by_word(&mut self, words: &[Word<C>]) -> Result<(), Error> {
        for word in words {
            self.transfer_word_in_place(*word)?;
        }
        Ok(())
    }
```

Anyone else seen anything like this before? I checked the D21 errata but found no mention of any issue with SPI disable.

Edit: another theory, as I hit submit - is it possible that the original write_word_by_word was re-enabling the receiver before the write actually went out, and so was actually receiving the byte it was trying to ignore? That'd suggest that another fix would be to wait for the write to complete before re-enabling.

jdeblese:
Also having problems with DMA-enabled SPI, and the fix was to add a `flush_tx()` call to `write_dma()` before re-enabling the receiver. Likely the same issue, so waiting for the transmission to complete is likely the best fix in both locations.

pellico:
I had the same problem when upgrading `feather_m0` create.
This behavior is due this [commit](https://github.com/atsamd-rs/atsamd/commit/8243da625aa51c7e43fae641bdbd9260ee2cd228#diff-6ae91602bb93e908a6b25e7da6bdb12f44889558eb8f20324bf88bfe8448b01d)

Original code was:
```rust

 #[inline]
    fn write_word_by_word(&mut self, words: &[Word<C>]) -> Result<(), Error> {
[...]

PR #1010, opened 2026-06-12
fix(spi): flush transmitter after write to prevent spurious reads (#1007)

jdeblese:
# Summary

Wait for SPI write transmissions to complete before re-enabling the receiver. If you don't, then code compiled for the samd21 with optimizations is fast enough to re-enable the receiver before the write occurs, resulting in occasional spurious bytes in the receive buffer. This does disable pipelined writes (partially counteracts 8243da6).

An alternative fix suggested in rust-embedded/embedded-hal#739 would be to have the flush occur at the start of a read instead of at the end of a write. That'd allow writes to be pipelined and mixed write/read transactions to work correctly, but requires more changes. I prefer to first fix the faulty behaviour with a minimal change.

Only fixes the sync implementation (regular and DMA), as I'm not familiar enough with the async to know if this is an issue there.

# Checklist

### DIFF
--- a/hal/src/sercom/spi/impl_ehal/dma.rs
+++ b/hal/src/sercom/spi/impl_ehal/dma.rs
@@ -70,6 +70,8 @@ where
 
         // Reenable receiver only if necessary
         if D::RX_ENABLE {
+            // must wait for transmit to complete before re-enable to avoid spurious receives
+            self.flush_tx();
             self.config.as_mut().regs.rx_enable();
         }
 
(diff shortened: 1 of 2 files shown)


jbeaurivage:
Hmm. I'm willing to merge this as an immediate fix, but we should probably think about a better way to handle pipelining. The alternative solution does seem sound to me. I'll keep #1007 open for the time being to track a future pipelined implementation.

PR #1013, opened 2026-08-10
changed(pacs)!: Build PACs on demand and update PAC APIs

jbeaurivage:
# Summary
The pre-generated PACs sitting in the repo have always been a pain point. This PR enables generating the PAC code at build-time using a build.rs scripts, which enormously reduces the number of files in the PAC crates. It also means we no longer need to run an `update-pacs` script whenever we update SVDs or `svd2rust` version, which is now a build-time dependency.

Currently the `svd2rust` dependency is at 0.37.1, which is the latest version.

I created a new helper library, `atsamd-hal-pacs-builder`, which contains the PAC generation code. It also uses `svdtools`, and much nicer YAML-based SVD patch templates rather than XSLT, which don't have good, pure-rust library offerings. The individual `build.rs` scripts just contain glue to call the codegen function. The PACs' `lib.rs` is now just this:

```rs
#![no_std]
#![allow(clippy::all)]
#![allow(non_camel_case_types)]
#![allow(non_snake_case)]
#![allow(unsafe_op_in_unsafe_fn)]

include!(concat!(env!("OUT_DIR"), "/pac.rs"));

pub use __pac_impl::generic::*;
pub use __pac_impl::*;
```

I also updated the SVDs to the latest version provided by Microchip. As of right now, SAMD21 and SAMD51 devices are using official SVD releases from Microchip. The others use unofficial releases provided to us by Microchip while waiting for them to be officially released.

The new SVD+svd2rust combo brings in relatively significant API changes, so this will be a breaking change for PACs and the HAL.

### Notes
1. Deleting the generated PAC code from the repo obviously creates an absolutely monstrous git diff. I'll try to contain the file deletions and SVD updates to their own commits, so we can make a more palatable diff from the other commits in the PR for review.
1. ~~We are also waiting for `svd-parser` to make a new release. We need [these changes](https://github.com/rust-embedded/svd/pull/301) to be released to crates.io for the new PACs to be accepted by `svd2rust`. Currently I'm patching the dependency to the git version in the workspace `Cargo.toml`, but I don't think that's allowed if we want to release `atsamd-hal-pacs-builder` to crates.io.~~ svd-parser released!
1. Removing the generated PAC code and updating SVDs obviously results in a HUGE diff. I tried to break the commits down in more digestible chunks. The interesting stuff (changes to code) can be found [here](https://github.com/atsamd-rs/atsamd/compare/aac74e6cc8e8bd3dab05c8feab067a09bd76c605...8dd009e076803ca281ceac261f590195678881ce).

### Checklist
- [x] SAMD11 SVD v2.8.208, unofficial from Microchip
- [x] SAMD21 SVD v3.8.270, 2026-02-11
- [x] SAME51 SVD v3.9.276, 2026-06-11
- [x] SAME51 SVD v3.10.279, unofficial from Microchip
- [x] SAME53 SVD v3.11.259, unofficial from Microchip
- [x] SAME54 SVD v3.12.276, unofficial from Microchip
- [x] `svd-parser` release > 0.14.10


atsamd-hal-pac-builder/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+    }
+
+    Err(anyhow::anyhow!("Chip not found"))
+}
+
+/// Get the chip name from the PAC package name
+fn get_chip_name(pkg_name: impl AsRef<str>) -> Option<String> {
+    let re = regex::Regex::new(r"atsam[a-z]\d{2}[a-z]+").unwrap();
```
rnd-ash: I think to allow for easier addition of future chips (PIC series), we should have another way of expressing this?

atsamd-hal-pac-builder/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "atsamd-hal-pac-builder"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+anyhow = "1.0"
+svd2rust = "0.37.1"
```
rnd-ash: Maybe its best to pin this package? - Since if anything on SVD2Rust's end changes the PAC generation (Even in patch changes), it could potentially break the generated PAC with the HAL

rnd-ash (approved):
Overall really happy with this, just a couple minor thoughts I've left, but nothing major.

Its also great to see all peripheral register writes are far more clearer, and are now correct with how the datasheets say to modify certain registers (For example interrupt registers)

The only outstanding question I have is what is with the feature flags being changed on all the PACs to include `<PAC>/critical-section` feature flag?

atsamd-hal-pac-builder/src/lib.rs
jbeaurivage: Fixed, the chip names are now specified in the Cargo.toml package metadata instead of trying to parse them with regex.

atsamd-hal-pac-builder/Cargo.toml
jbeaurivage: Done.

jbeaurivage:
Great! To answer your question, the newer versions of svd2rust rely on `critical-section` for the `Peripherals::take` API. If the HAL doesn't use that feature, `Peripherals::take` is not available, and we would be forced to only rely on `Peripherals::steal`. `atsamd-hal` already uses `critical-section` as a non-optional dependency, so this change is transparent to the user.

hal/src/dmac/async_api.rs
```diff
@@ -37,11 +37,11 @@ impl Handler<DMAC> for InterruptHandler {
                 let wake = if dmac.chintflag().read().tcmpl().bit_is_set() {
                     // Transfer complete. Don't clear the flag, but
                     // disable the interrupt. Flag will be cleared when polled
-                    dmac.chintenclr().modify(|_, w| w.tcmpl().set_bit());
+                    dmac.chintenclr().write(|w| w.tcmpl().clear_bit_by_one());
```
TDHolmes, line 40: `<rusty at rust comment>`
I might be misunderstanding the APIs / mechanics here, but shouldn't you write a bit to a clear register/API? `clear_bit_by_one()` sounds like it clears a bit rather than setting it.
`</rusty at rust comment>`

jbeaurivage: Thanks! I understand where you're coming from. Basically microchip updated their SVDs, and now mark registers/bits that are cleared by writing a one. svd2rust understands this, and provides the `clear_bit_by_one` method instead of `set_bit`. It compiles down to the same instructions (ie, writing a one), but the intent is clearer.

TDHolmes: makes sense! Yeah I think it would be clearer just reading it in a vacuum, the diff with what was there before was tripping me up I think.

